PR #11, opened 2015-07-12
Add support for WriteBatch

dgrnbrg:
I've created a new trait, Writable, which is used to allow the same code to interact directly with the database, or with a WriteBatch. I'm not sure what happens if you insert multiple types of records for the same key into the DB in a single WriteBatch; perhaps that should be forbidden.

You can create a writebatch with `WriteBatch::new()`, and you can write it to the DB using `db::write(batch)`.

This is my first Rust code, so please look at it very carefully--I'm not sure whether I'm correctly freeing the C-allocated writebatch object. If that's being allocated and freed correctly, however, I'm reasonably confident in this implementation.

### DIFF
--- a/src/ffi.rs
+++ b/src/ffi.rs
@@ -272,6 +272,10 @@ extern {
     pub fn rocksdb_iter_get_error(iter: RocksDBIterator,
                                   err: *const *const u8);
     // Write batch
+    pub fn rocksdb_write(db: RocksDBInstance,
+                         writeopts: RocksDBWriteOptions,
+                         batch : RocksDBWriteBatch,
+                         err: *mut i8);
     pub fn rocksdb_writebatch_create() -> RocksDBWriteBatch;
     pub fn rocksdb_writebatch_create_from(rep: *const u8,
                                           size: size_t) -> RocksDBWriteBatch;
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -34,6 +34,7 @@ pub use rocksdb::{
     RocksDB,
     RocksDBResult,
     RocksDBVector,
+    Writable,
 };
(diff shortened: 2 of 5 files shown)


spacejam:
Hey Dan!  Welcome to the Rust community :) This is a great addition to the codebase, and I'm happy that you mentioned the concern about memory being freed because when I fired up valgrind a few leaks popped up in the old code (pretty much any of the *Options are not being freed by RocksDB as I incorrectly assumed.)  I'm going to merge your code, push some fixes for the memory leaks, and cut a new crate version (0.0.6) that includes these changes.  In case you're curious, the way I found the memory leaks is fairly straightforward (I feel silly for not doing this earlier):

1. Run the code under test in a loop { ... } construct (I just lazily modified the main function in main.rs)
2. `cargo build`
3. `valgrind [--show-reachable=yes] --leak-check=full ./target/debug/rocksdb`
4. `^c` after some time passes.  The output traces will be rust-friendly (Rust also has nice DWARF symbols and plays well with lldb if you ask cargo/rustc to compile your code with debugging info).

Thanks for contributing this!

spacejam:
re: multiple records (I'm assuming you mean mixing adds and merges?) this is completely fine.  While a merge may take the last add into consideration (it will appear to the merge operator as the existing_val argument), if an add was the last operation then rocksdb will not perform any of the preceding merges.  Please let me know if I misunderstood what you meant by records, however.

dgrnbrg:
I meant making 2 adds to the writebatch. Actually, there's apparently a WriteBatchWithIterator in C++ that has specified semantics, but it's not in the C API yet...oh well, when it lands it'll be amazing!

PR #14, opened 2015-07-12
Improve safety of borrow checker with ffi objects

dgrnbrg:
So, I fixed up the use of borrow and &mut references with the configuration options, with the goal of laying the foundation for exposing the fully configurable read & write APIs, where leaks & API simplicity will be most important.

I started to remove some prefixes from the variables, so that just `use rocksdb` will allow us to write `rocksdb::Options`, `rocksdb::DB`, etc. I hope this is alright; I'll stop doing that if you prefer `use`ing the members of the namespace.

### DIFF
--- a/src/rocksdb.rs
+++ b/src/rocksdb.rs
@@ -25,45 +25,50 @@ use std::slice;
 use std::str::from_utf8;
 
 use rocksdb_ffi;
-use rocksdb_ffi::RocksDBSnapshot;
-use rocksdb_options::RocksDBOptions;
+use rocksdb_options::Options;
 
 pub struct RocksDB {
     inner: rocksdb_ffi::RocksDBInstance,
 }
 
 pub struct WriteBatch {
-    inner: rocksdb_ffi::RocksDBWriteBatch
+    inner: rocksdb_ffi::RocksDBWriteBatch,
+}
+
+pub struct ReadOptions {
+    inner: rocksdb_ffi::RocksDBReadOptions,
 }
 
 // This is for the RocksDB and write batches to share the same API
 pub trait Writable {
-    fn put(&self, key: &[u8], value: &[u8]) -> Result<(), &str>;
-    fn merge(&self, key: &[u8], value: &[u8]) -> Result<(), &str>;
-    fn delete(&self, key: &[u8]) -> Result<(),&str>;
+    fn put(&self, key: &[u8], value: &[u8]) -> Result<(), String>;
+    fn merge(&self, key: &[u8], value: &[u8]) -> Result<(), String>;
+    fn delete(&self, key: &[u8]) -> Result<(), String>;
 }
 
-fn error_message<'a>(ptr: *const i8) -> &'a str {
-    unsafe {
-        return from_utf8(CStr::from_ptr(ptr).to_bytes()).unwrap();
-    }
+fn error_message(ptr: *const i8) -> String {
+    let c_str = unsafe { CStr::from_ptr(ptr) };
+    from_utf8(c_str.to_bytes()).unwrap().to_owned()
 }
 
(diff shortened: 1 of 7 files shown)


src/rocksdb.rs
```diff
@@ -25,45 +25,50 @@ use std::slice;
...
+    fn delete(&self, key: &[u8]) -> Result<(), String>;
 }
 
-fn error_message<'a>(ptr: *const i8) -> &'a str {
-    unsafe {
-        return from_utf8(CStr::from_ptr(ptr).to_bytes()).unwrap();
-    }
+fn error_message(ptr: *const i8) -> String {
+    let c_str = unsafe { CStr::from_ptr(ptr) };
+    from_utf8(c_str.to_bytes()).unwrap().to_owned()
```
spacejam: Didn't know about to_owned, nice!

src/rocksdb_options.rs
```diff
@@ -23,37 +23,91 @@ use merge_operator::{self, MergeOperatorCallback, MergeOperands, full_merge_call
               partial_merge_callback};
 use comparator::{self, ComparatorCallback, compare_callback};
 
-#[derive(Copy, Clone)]
-pub struct RocksDBOptions {
+pub struct BlockBasedOptions {
+    inner: rocksdb_ffi::RocksDBBlockBasedTableOptions,
+}
+
+pub struct Options {
```
spacejam: I'm totally on-board for simplifying these variable names, thanks for beginning this process.

src/rocksdb_options.rs
```diff
@@ -23,37 +23,91 @@ use merge_operator::{self, MergeOperatorCallback, MergeOperands, full_merge_call
               partial_merge_callback};
 use comparator::{self, ComparatorCallback, compare_callback};
 
-#[derive(Copy, Clone)]
```
spacejam: Pretty much any case where this line appears (esp ffi.rs) it's evidence of a reckless hack that was used to get something to compile, so as we move forward we should not hesitate to rip similar lines out wherever possible.

src/rocksdb_options.rs
```diff
@@ -23,37 +23,91 @@ use merge_operator::{self, MergeOperatorCallback, MergeOperands, full_merge_call
...
     pub inner: rocksdb_ffi::RocksDBOptions,
-    block_options: rocksdb_ffi::RocksDBBlockBasedTableOptions,
 }
 
-impl RocksDBOptions {
-    pub fn new() -> RocksDBOptions {
+impl Drop for Options {
+    fn drop(&mut self) {
+        unsafe {
+            rocksdb_ffi::rocksdb_options_destroy(self.inner);
```
spacejam: :+1:

src/rocksdb_options.rs
```diff
@@ -23,37 +23,91 @@ use merge_operator::{self, MergeOperatorCallback, MergeOperands, full_merge_call
...
+    ////            self.inner, cache);
+    ////    }
+    ////}
+
+    ////pub fn set_cache_compressed(&mut self, cache: rocksdb_ffi::RocksDBCache) {
+    ////    unsafe {
+    ////        rocksdb_ffi::rocksdb_block_based_options_set_block_cache_compressed(
+    ////            self.inner, cache);
+    ////    }
+    ////}
```
spacejam: I agree that these hack methods should be removed from the API until a nicer implementation arises.

spacejam:
This is great stuff, I really appreciate you taking the time to clean up some of the horrible things I wrote to get certain functionality hobbling along.  This kind of stuff is really important for people who are hoping to build reliable systems using this! (I've got a few dozen ideas I've been dying to get more time to start implementing on top of it)

PR #15, opened 2015-07-15
Add Iterators and Snapshots

dgrnbrg:
Also cleans up the public API.

Still to come is documentation and tests.

### DIFF
--- a/src/ffi.rs
+++ b/src/ffi.rs
@@ -181,7 +181,7 @@ extern {
         bits_per_key: c_int) -> RocksDBFilterPolicy;
     pub fn rocksdb_open(options: RocksDBOptions,
                         path: *const i8,
-                        err: *mut i8
+                        err: *mut *const i8
                         ) -> RocksDBInstance;
     pub fn rocksdb_writeoptions_create() -> RocksDBWriteOptions;
     pub fn rocksdb_writeoptions_destroy(writeopts: RocksDBWriteOptions);
@@ -189,7 +189,7 @@ extern {
                        writeopts: RocksDBWriteOptions,
                        k: *const u8, kLen: size_t,
                        v: *const u8, vLen: size_t,
-                       err: *mut i8);
+                       err: *mut *const i8);
     pub fn rocksdb_readoptions_create() -> RocksDBReadOptions;
     pub fn rocksdb_readoptions_destroy(readopts: RocksDBReadOptions);
     pub fn rocksdb_readoptions_set_verify_checksums(
@@ -216,14 +216,14 @@ extern {
                        readopts: RocksDBReadOptions,
                        k: *const u8, kLen: size_t,
                        valLen: *const size_t,
-                       err: *mut i8
+                       err: *mut *const i8
                        ) -> *mut c_void;
     pub fn rocksdb_get_cf(db: RocksDBInstance,
                        readopts: RocksDBReadOptions,
                        cf_handle: RocksDBCFHandle,
                        k: *const u8, kLen: size_t,
                        valLen: *const size_t,
-                       err: *mut i8
+                       err: *mut *const i8
                        ) -> *mut c_void;
     pub fn rocksdb_create_iterator(db: RocksDBInstance,
                                    readopts: RocksDBReadOptions
@@ -239,19 +239,19 @@ extern {
     pub fn rocksdb_delete(db: RocksDBInstance,
                           writeopts: RocksDBWriteOptions,
                           k: *const u8, kLen: size_t,
-                          err: *mut i8
+                          err: *mut *const i8
                           ) -> *mut c_void;
     pub fn rocksdb_close(db: RocksDBInstance);
     pub fn rocksdb_destroy_db(options: RocksDBOptions,
-                              path: *const i8, err: *mut i8);
+                              path: *const i8, err: *mut *const i8);
     pub fn rocksdb_repair_db(options: RocksDBOptions,
-                             path: *const i8, err: *mut i8);
+                             path: *const i8, err: *mut *const i8);
     // Merge
     pub fn rocksdb_merge(db: RocksDBInstance,
                          writeopts: RocksDBWriteOptions,
                          k: *const u8, kLen: size_t,
                          v: *const u8, vLen: size_t,
-                         err: *mut i8);
+                         err: *mut *const i8);
     pub fn rocksdb_mergeoperator_create(
         state: *mut c_void,
         destroy: extern fn(*mut c_void) -> (),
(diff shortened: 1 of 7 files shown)


src/rocksdb.rs
```diff
@@ -39,15 +40,130 @@ pub struct ReadOptions {
...
-    fn merge(&self, key: &[u8], value: &[u8]) -> Result<(), String>;
-    fn delete(&self, key: &[u8]) -> Result<(), String>;
+    fn put(&mut self, key: &[u8], value: &[u8]) -> Result<(), String>;
+    fn merge(&mut self, key: &[u8], value: &[u8]) -> Result<(), String>;
+    fn delete(&mut self, key: &[u8]) -> Result<(), String>;
 }
 
 fn error_message(ptr: *const i8) -> String {
     let c_str = unsafe { CStr::from_ptr(ptr) };
+//TODO I think we're leaking the c string here; should be a call to free once realloced into rust String
```
spacejam: Correct, CString does not implement Drop so we're leaking here.

src/rocksdb.rs
```diff
@@ -39,16 +40,134 @@ pub struct ReadOptions {
...
+    fn merge(&mut self, key: &[u8], value: &[u8]) -> Result<(), String>;
+    fn delete(&mut self, key: &[u8]) -> Result<(), String>;
 }
 
 fn error_message(ptr: *const i8) -> String {
     let c_str = unsafe { CStr::from_ptr(ptr) };
-    from_utf8(c_str.to_bytes()).unwrap().to_owned()
+    let s = from_utf8(c_str.to_bytes()).unwrap().to_owned();
+    unsafe{
+        libc::free(ptr as *mut libc::c_void);
```
dgrnbrg: Can you confirm that this is doing the right thing?

We take the pointer to the C string, copy its bytes into a &str, then `to_owned()` should allocate a new buffer according to my research.

At that point, we can free the original string.

spacejam: I believe that this is correct, but I haven't tested it.  Calling to_owned on &str creates a vector, pushes all of the bytes in, and wraps it in a String.

src/ffi.rs
```diff
@@ -294,12 +294,12 @@ extern {
...
                               vlen: *mut size_t) -> *mut u8;
     pub fn rocksdb_iter_get_error(iter: RocksDBIterator,
-                                  err: *const *const u8);
+                                  err: *mut *const u8);
     // Write batch
     pub fn rocksdb_write(db: RocksDBInstance,
                          writeopts: RocksDBWriteOptions,
                          batch : RocksDBWriteBatch,
-                         err: *mut i8);
+                         err: *mut *const i8);
```
spacejam: :+1: Thanks for fixing all of this pointer mis-specification!

src/main.rs
```diff
@@ -17,27 +17,103 @@
 
 extern crate rocksdb;
 extern crate test;
-use rocksdb::{Options, RocksDB, MergeOperands, new_bloom_filter, Writable};
+use rocksdb::{Options, RocksDB, MergeOperands, new_bloom_filter, Writable, };
```
spacejam: not blocker, but if rustfmt is usable by now then we should probably run it on code before merging

src/main.rs
```diff
@@ -17,27 +17,103 @@
...
+        assert!(p.is_ok());
+        let p = db.put(b"k2", b"v2222");
+        assert!(p.is_ok());
+        let p = db.put(b"k3", b"v3333");
+        assert!(p.is_ok());
+        let mut snap = db.snapshot();
+        let mut view1 = snap.iterator();
+        println!("See the output of the first iter");
+        for (k,v) in view1.from_start() {
+            println!("Hello {}: {}", std::str::from_utf8(k).unwrap(), std::str::from_utf8(v).unwrap());
```
spacejam: Rust squelches output during tests, but this can be overridden with something like `cargo test -- --nocapture` and may be useful occasionally to see how far a failing test gets.

src/rocksdb.rs
```diff
@@ -39,16 +40,134 @@ pub struct ReadOptions {
...
+pub struct DBIterator {
+//TODO: should have a reference to DB to enforce scope, but it's trickier than I thought to add
+    inner: rocksdb_ffi::RocksDBIterator,
+    direction: Direction,
+    just_seeked: bool
+}
+
+pub enum Direction {
+  forward, reverse
+}
```
spacejam: Is there a reason why you implemented SubDBIterator instead of [DoubleEndedIterator](https://doc.rust-lang.org/std/iter/trait.DoubleEndedIterator.html)?

src/rocksdb.rs
dgrnbrg: Can you explain to me more about using valgrind? I've never used it before, but I'd like to learn.

src/main.rs
dgrnbrg: I just checked, it seems to require editing the code before you should run it...I think it will be great soon, but not today :)

src/rocksdb.rs
dgrnbrg: A DoubleEndedIterator supports deque-like semantics; however, the rocksdb cursors are designed to scan most efficiently in forward order, and they scan slower in reverse. To support DEI semantics, we'd seek between every key, yielding poor performance.

The solution to that is to create a Snapshot, then create 2 iterators on it, one to go forward; the other, reverse.

src/rocksdb.rs
spacejam: Certainly - basically it accepts an arbitrary binary, instruments the code, and executes it while tracing memory usage and a number of other configurable things.  You want to create a binary that will possibly lead to leaks, which is the motivation behind the valgrind feature that was added.  Basically it just runs in a tight loop, and when you send sigint to it, it will print out a summary of its findings and exit.  The way I invoke it with this project is:
```shell
cargo build --features=valgrind
valgrind --leak-check=full ./target/debug/rocksdb
# wait some time,
^C
# look for lost byte reports
```
It can also give you information about cache usage (cachegrind) or deadlocks/data races (helgrind) or heap data (massif).

spacejam: I'm not sure if you're using OSX Yosemite or not, but in case you are you can install valgrind with `brew install --HEAD valgrind` (without --HEAD it does not work for this OS).

src/rocksdb.rs
spacejam: Makes sense.  So, the downside to implementing DEI is that you need to have 2 internal RocksDB iterators, and because they exhaust the same range each next / next_back call needs to ensure that the two ends don't move past each other.  I think the upside is that by implementing DEI it will present a more idiomatic interface, and you get the rev() method for free, but I'm not sure if this is worth it or not.  I don't have a super strong opinion on it, but we may want to reconsider it in the future if it's confusing to people.

spacejam:
LGTM, apologies for the slow review - crunch time at Mesosphere :) I'll see if I can get rustfmt going a bit later tonight

PR #18, opened 2015-07-17
Fix boxing problems and add more tests

dgrnbrg:
There's a bug in the lifetime tracking of the iterator before--it turns out that I couldn't figure out how to kill the borrows at the correct time, and you could sometimes read wrong data from the iterator. This patch fixes that by boxing the [u8].

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -3,9 +3,20 @@
 name = "rocksdb"
 description = "A Rust wrapper for Facebook's RocksDB embeddable database."
 version = "0.0.6"
-authors = ["Tyler Neely <t@jujit.su>"]
+authors = ["Tyler Neely <t@jujit.su>", "David Greenberg <dsg123456789@gmail.com>"]
 license = "Apache-2.0"
+exclude = [
+  ".gitignore",
+  ".travis.yml",
+  "deploy.sh",
+  "test/**/*",
+]
 
 [features]
 default=[]
 valgrind=[]
+
+[[test]]
+
+name = "test"
+path = "test/test.rs"
--- a/src/rocksdb.rs
+++ b/src/rocksdb.rs
@@ -62,9 +62,9 @@ pub struct SubDBIterator<'a> {
 }
 
 impl <'a> Iterator for SubDBIterator<'a> {
-    type Item = (&'a [u8], &'a [u8]);
+    type Item = (Box<[u8]>, Box<[u8]>);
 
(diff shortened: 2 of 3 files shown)


src/rocksdb.rs
```diff
@@ -83,7 +83,8 @@ impl <'a> Iterator for SubDBIterator<'a> {
             let key = unsafe { slice::from_raw_parts(key_ptr, key_len as usize) };
             let val_ptr = unsafe { rocksdb_ffi::rocksdb_iter_value(native_iter, val_len_ptr) };
             let val = unsafe { slice::from_raw_parts(val_ptr, val_len as usize) };
-            Some((key,val))
+
+            Some((key.to_vec().into_boxed_slice(),val.to_vec().into_boxed_slice()))
```
dgrnbrg: This is the critical line to remove the bug! I wonder if a simple `Box::new(key.clone())` would've sufficed...

spacejam: One thing I want to be careful about is doing too much copying.  What do you think of switching from Box<[u8]>'s to the RocksDBVector type so that we can avoid copying everything using to_vec?

dgrnbrg: This requires a copy--once you call iter_next, the previous pointer is corrupt. That's the problem I ran into while writing integration tests.

The next thing I'd like to do is add a binary codec trait, and then make the RocksDB be parameterized on the actual type. That way, register-sized types could be iterated/retrieved as values, and reference types could do their own custom codec, so that the API is in terms of the actual application logic. I'll probably initially do support for [u8], String, and u64.

spacejam: Ahh I see, makes sense.  One thing I may like eventually is to have a codec that is tied to a particular column family, rather than DB-wide, so feel free to implement it DB-wide in the mean time but with the knowledge that we may want to make it CF-specific somehow down the road.

dgrnbrg: Totally understand; I think of the current incarnation of the code as hardcoded to specialize on [u8], but that will change.

Cargo.toml
```diff
@@ -3,9 +3,20 @@
...
+]
 
 [features]
 default=[]
 valgrind=[]
+
+[[test]]
+
+name = "test"
+path = "test/test.rs"
```
spacejam: Does this need to be `git add`'d?

dgrnbrg: Oops!

Issue #22, opened 2015-07-18
Add Multi-thread support


mneumann:
Atm, it seems to be not possible to access a RocksDB instance from multiple threads.

spacejam:
Thanks for fixing this!

spacejam:
This has landed in the 0.0.8 release that I just cut.

mneumann:
Cool! :+1:

PR #23, opened 2015-07-18
Allow RocksDB to be shared across threads

mneumann:
Issue #22


spacejam:
LGTM, thank you for the contribution!

PR #32, opened 2015-11-08
nuke db result

spacejam:


spacejam:
@steveklabnik @petehunt
I'm merging this now, but please let me know if any of this is bizarre and I'll be happy to keep iterating on it!

steveklabnik:
Ah nice!

This looks great. You may or may not want to create your own alias for more specific errors, but there's no need to do that right away.

PR #33, opened 2015-11-08
Use more regular semver suggestion

steveklabnik:
This is the same as using ^, which is a bit nicer than ~ imho, and is the default for a reason :)


spacejam:
Thanks!

PR #37, opened 2015-12-08
Change iterator to take handle

benjumanji:
This changes the DBIterator interface to match the convention of taking
a handle over a string as per the other column-family-aware apis.


benjumanji:
It leaves the same number of tests busted as before.

spacejam:
Awesome, thank you for this!  I'll cut a new 0.2.2 release on crates.io with this in it in a few minutes

spacejam:
0.2.2 is cut with these changes!

PR #39, opened 2015-12-27
More idiomatic iterators?

petehunt:
I had an issue where I wanted to store a reference to a `SubDBIterator` on a struct. Because it's created from a `DBIterator`, I ran into an issue where I couldn't store a reference to both the iterator and sub-iterator on the same struct, and couldn't get the `DBIterator` instance to live long enough (see: https://internals.rust-lang.org/t/self-referencing-structs/418)

This PR merges `SubDBIterator` into `DBIterator`. While I was working on it, I fixed a few todos in the code (the DB lifetime one was also confusing).

I can't think of a better name for the `IteratorMode` enum. Also, I am a rust newb :)

### DIFF
--- a/README.md
+++ b/README.md
@@ -49,7 +49,7 @@ fn main() {
     match db.get(b"my key") {
         Ok(Some(value)) => println!("retrieved value {}", value.to_utf8().unwrap()),
         Ok(None) => println!("value not found"),
-        Err(e) => println!("operational problem encountered: {}", e), 
+        Err(e) => println!("operational problem encountered: {}", e),
     }
 
     db.delete(b"my key");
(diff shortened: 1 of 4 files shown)


rscnt:
:+1:

spacejam:
Sorry for the slow feedback, I've been avoiding github for the last few days!  The code looks good, can you update the readme to make use of the new syntax?  Then I'll merge and cut a new version with these improvements!  Thank you for the pull request!

petehunt:
Updated the README. Please let me know if you see anything weird!

spacejam:
Thanks!  I'll cut `0.3.0` now with your improvements!

Issue #44, opened 2016-02-09
32-bit build fails


petevine:
On `i686` there's a `mismatched types` failure:

```rust
  Compiling rocksdb v0.3.0
/tmp/.cargo/registry/src/github.com-48ad6e4054423464/rocksdb-0.3.0/src/rocksdb_options.rs:65:69: 65:73 error: mismatched types:
 expected `u32`,
    found `u64` [E0308]
/tmp/.cargo/registry/src/github.com-48ad6e4054423464/rocksdb-0.3.0/src/rocksdb_options.rs:65                                                                     size);
                                                                                                                                                                 ^~~~
/tmp/.cargo/registry/src/github.com-48ad6e4054423464/rocksdb-0.3.0/src/rocksdb_options.rs:65:69: 65:73 help: run `rustc --explain E0308` to see a detailed explanation
/tmp/.cargo/registry/src/github.com-48ad6e4054423464/rocksdb-0.3.0/src/rocksdb_options.rs:234:64: 234:68 error: mismatched types:
 expected `u64`,
    found `u32` [E0308]
/tmp/.cargo/registry/src/github.com-48ad6e4054423464/rocksdb-0.3.0/src/rocksdb_options.rs:234                                                                size);
                                                                                                                                                             ^~~~
/tmp/.cargo/registry/src/github.com-48ad6e4054423464/rocksdb-0.3.0/src/rocksdb_options.rs:234:64: 234:68 help: run `rustc --explain E0308` to see a detailed explanation
error: aborting due to 2 previous errors
Build failed, waiting for other jobs to finish...
Could not compile `rocksdb`.
````
Naturally, ARM is affected too.

petevine:
Fixed in https://github.com/ethcore/parity/pull/464

PR #45, opened 2016-02-12
Fix compilation on ARM for Rust >= 1.6

petevine:
Due to this [breakage](https://github.com/rust-lang/rust/issues/29867) ARM doesn't work out of the box.

### DIFF
--- a/src/ffi.rs
+++ b/src/ffi.rs
@@ -90,7 +90,7 @@ pub enum DBUniversalCompactionStyle {
 }
 
 pub fn error_message(ptr: *const i8) -> String {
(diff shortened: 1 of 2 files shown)


spacejam:
Awesome, thanks for this fix, @petevine!

petevine:
You're welcome!

There's still issue #44 (that I solved locally by casting to correct types) but it requires a more cross-platform solution I believe.

-original message-
Subject: Re: [rust-rocksdb] Fix compilation on ARM for Rust >= 1.6 (#45)
From: Tyler Neely <notifications@github.com>
Date: 12.02.2016 22:07

Awesome, thanks for this fix, @petevine!

---

PR #49, opened 2016-03-01
*: Fix clippy warnings.

ngaut:


spacejam:
@ngaut thank you for putting in the time to clean things up!

PR #53, opened 2016-04-09
Disable clippy

ngaut:
Clippy can't build with nightly rust.


spacejam:
This is great stuff, and I would be super happy for a PR with as much of it as you would care to upstream when you have time to submit one!

PR #54, opened 2016-04-19
export `iterator.valid()`

disksing:
The `valid()` method could be helpful when we want to know if an iterator is positioned at a kv pair after calling `seek()`.


spacejam:
LGTM, thanks for the contribution!  I'll cut a new version 0.3.2 that contains these changes now!

Issue #64, opened 2016-08-03
Error Compiling the documentation


timlyo:
I'm trying to build the documentation with `cargo doc` and I get the following:

`error: cannot document a package where a library and a binary have the same name. Consider renaming one or marking the target as 'doc = false'`

Adding 

```
[lib]
name = "rocksdb"

[[bin]]
name = "rocksdb_bin"
```

or 

```
[[bin]]
name = "rocksdb"
doc = false
```

to `cargo.toml` fixes this issue and allows me to compile. Would it be worth adding some form of one of the two?

spacejam:
Feel free to nuke the binary completely! It has no purpose

timlyo:
Ok, submitted a pull request.

spacejam:
Thanks, @timlyo! merged

PR #65, opened 2016-08-03
main.rs deleted

timlyo:
As discussed in #64


spacejam:
Thanks for fixing this!

PR #66, opened 2016-08-05
Added some documentation

timlyo:
I've been adding some documentation as I'm learning about the library. Not much, but it might be useful to someone else.

### DIFF
--- a/src/rocksdb.rs
+++ b/src/rocksdb.rs
@@ -251,16 +251,25 @@ pub trait Writable {
 }
 
 impl DB {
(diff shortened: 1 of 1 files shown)


spacejam:
Thanks for helping with this! I've been meaning to go through this and add proper docs for a while now, as the lack of them is a big wart at this point.

timlyo:
No problem, I would have done a bit more, but I've moved away from rocksdb as it doesn't really fit the needs of my project. Thanks for everything you've done with this crate though.

PR #68, opened 2016-09-18
Add Multiget Operations

jezell:


src/rocksdb.rs
```diff
@@ -508,6 +518,117 @@ impl DB {
...
+                    let em = error_message(err);
+                    values.push(Err(em));
+                }
+
+                let val = value_ptrs[i];
+                if val.is_null() {
+                    values.push(Ok(None));
+                } else {
+                    values.push(Ok(Some(DBVector::from_c(val, length_ptrs[i]))));
+                }
```
spacejam: I think it's important to maintain the 1:1 mapping between requested keys, and a single `Err`/`Ok` in `values`. Can we make the `if val.is_null() {` an `else if` after the check for errors, to guarantee a 1:1 mapping in returned values? (and the same for the multi_get_cf code below?)

src/rocksdb.rs
```diff
@@ -975,6 +1096,36 @@ fn writebatch_works() {
...
+            _ =>  assert!(false)
+        }
+        match all[2] {
+            Ok(Some(ref u)) => assert_eq!(u.to_utf8(),Some("v3333")),
+            _ =>  assert!(false)
+        }
+    }
+    let opts = Options::default();
+    assert!(DB::destroy(&opts, path).is_ok());
+}
```
spacejam: Can we add a test here that requests a key that is not present, and makes sure the 1:1 mapping between requested keys and the resulting `Result`?

src/rocksdb.rs
spacejam: could also continue the for loop in the error check if statement, or whatever you prefer to maintain 1:1 return values to requested keys.

src/rocksdb.rs
jezell: sure

src/rocksdb.rs
jezell: yeah, that's a good catch.

PR #70, opened 2016-10-12
Implement keys_iterator methods

kaedroho:
This pull request adds a `keys_iterator` method to `DB` and `Snapshot`.

This allows iterating keys quickly without loading all the values into memory.

### DIFF
--- a/src/rocksdb.rs
+++ b/src/rocksdb.rs
@@ -55,6 +55,12 @@ pub struct DBIterator {
     just_seeked: bool,
 }
 
+pub struct DBKeysIterator {
+    inner: rocksdb_ffi::DBIterator,
+    direction: Direction,
+    just_seeked: bool,
+}
+
 pub enum Direction {
     Forward,
     Reverse,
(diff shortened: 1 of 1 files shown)


kaedroho:
Just been looking into implementing prefix iterators, adding two extra methods for that is starting to feel a bit bloated.

Developing the API following this PR, will give us four iterator methods:

 - ``.iterator()`` - Range iterator (keys and values)
 - ``.keys_iterator()`` - Range iterator (keys only)
 - ``.prefix_iterator()`` - Prefix iterator (keys and values)
 - ``.prefix_keys_iterator()`` - Prefix iterator (keys only)


But I think there's at least two better alternatives to this:

**Make iterators iterate keys by default; provide a method on the key object to get the value**

The iterator returns ``Key`` objects which provide methods to get the key/value. This is quite similar to the API RocksDB has in C++:

```rust
for k in db.iterator(...) {
    let key = k.key();
    let value = k.value();
}

for k in db.prefix_iterator(...) {
    let key = k.key();
    let value = k.value();
}
```

**Override the keys() method on DBIterator to make it return a DBKeysIterator instead**

This would make ``.iterator().keys()`` equivilant to ``.keys_iterator()`` implemented in this PR. It feels more "rustic" to me, and is backwards-compatible but I think the benefits will go away if the user uses another iterator method before ``.keys()``.

```rust
for key in db.iterator(...).keys() {

}

for key in db.prefix_iterator(...).keys() {

}
```


Let me know what you think

spacejam:
@kaedroho in light of the plan to merge changes from forks, would you like to push this through as-is, discard this, or hold off for the time being?

kaedroho:
I'd like to discard the changes I've made in the PR and instead implement one of the approaches explained in https://github.com/spacejam/rust-rocksdb/pull/70#issuecomment-255583821. But I think it's best to wait until the forks have merged before proceeding with that though.

kaedroho:
Replaced by #112

PR #71, opened 2016-10-23
Run tests on Travis CI

kaedroho:
This commit adds configuration for building/installing RocksDB on Travis so the unit tests can be run.

Builds now take around 10 minutes. Also, I'm not sure how to configure it for OSX but I'm not sure if this is really important?

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -5,9 +5,29 @@ rust:
   - 1.1.0
   - 1.2.0
 
-os:
-  - linux
-  - osx
+dist: trusty
+sudo: true
+
+addons:
+  apt:
+    sources:
+    - ubuntu-toolchain-r-test
+    packages:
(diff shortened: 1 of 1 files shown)


kaedroho:
I've just added a test matrix that allows building against different combinations of Rust/RocksDB versions.

It runs the following Rust versions:
 - 1.2
 - stable
 - beta

And the following RocksDB versions
 - 4.13 (latest version at time of this message)
 - 4.5 (the version that is shipped with Ubuntu Xenial)
 - 4.1 (an arbitary old version to see if it works)

spacejam:
Awesome! Thanks for adding this! I need to mark one of the tests skip which I think is also buggy when using the straight up C API (merge operator + column family) but I need to revisit this assumption.

PR #74, opened 2016-10-27
Added an Error type

kaedroho:
This makes it easier for people who are using the try!() macro to convert errors into their own Error type. It isn't possible to diffrentiate between errors raised from RocksDB and other string errors at the moment.

This adds a simple ``Error`` type that wraps ``String``. People using RocksDB can now implement ``impl From<rocksdb::Error> for MyError`` and add custom behaviour for handling RocksDB errors.

I've implemented ``From<rocksdb:Error> for String`` and added a ``to_string`` method to the ``Error`` type itself to make it easier for those relying on strings to upgrade.


spacejam:
:+1:  Thanks!

Issue #69, opened 2016-09-29
[Question] Any chance bringing the forks together?


xenji:
Hey there!

First, thanks for making this all happen, it helps me a lot in my daily work!

Secondly: Is there a chance, that you/we can get all forks on one table and see how this can be beneficial for the project? For example: You and also https://github.com/ngaut/rust-rocksdb seem to have made pretty good progress, but the branches are too diverged and I lack of knowledge to merge them my self. 

So I try and shoutout to @ngaut, @BusyJay and @zhangjinpeng1987 to join the discussion if those two forks should be one again?

Hope that helps somehow ...

Cheers,
Mario

spacejam:
Yeah! ngaut's (being used for the tikv project) is the main one I would
like to pull in! Right now I'm backpacking in the Alps though and will be
away from a computer until the middle of October, but I can start to work
on this more when I get back! If you would like to work on this at all
before then just say so and I'll be able to review/merge from my phone at
some point every other day or so!

ngaut:
Hi, @xenji We would like to merge all of the changes to the repo. It's just there is a problem:
The new interfaces are not compatible with the previous version of rust-rocksdb.
And that could bother users who are already  used rust-rocksdb in their project.

spacejam:
I'm happy cutting a new major number for an api breaking new version

xenji:
Seems like a plan. Good to have you all in the conversation!

Should we revive this discussion and dive deeper when @spacejam is back from the Alps? I think this needs more than just a bit of talking, doesn't it?

vmx:
@spacejam If you stick to Semantic Versioning you don't even need to cut a new major version [as long as you don't have a 1.0.0 release yet][1].

[1]: http://semver.org/#doesnt-this-discourage-rapid-development-and-fast-iteration

alexreg:
Any updates on all this?

kaedroho:
Hey Everyone,

I'd just like to say, I'm keen to help out with improving the general quality of this wrapper (such as writing API docs and improving the tests). I also need some more features of RocksDB for my project, which I'm happy to put some of my own time in to implementing.

It looks like some of this work may have already been done in the other forks though. I'm not sure what I should do as my availability to help out will deminish after this week but any work I do now will probably conflict or duplicate with what has been done in the other forks.

I'm happy to help out in any way I can to get the changes in the forks pushed upstream so let me know if you need an extra pair of hands.

ngaut:
[...]

PR #75, opened 2016-10-27
Port path() method from ngauts fork and implement Debug for DB

kaedroho:
Part of #69

This pull request pulls across https://github.com/ngaut/rust-rocksdb/pull/15 which implements a ``path()`` method on ``DB`` that returns the path the ``DB`` was opened with. I've also implemented the ``Debug`` trait for ``DB`` which uses this new method.


spacejam:
:+1:

PR #77, opened 2016-10-27
Add set_iterate_upper_bound method on ReadOptions

kaedroho:
Part of #69

This pull request ports https://github.com/ngaut/rust-rocksdb/pull/31. I've removed the dead code and there is no longer a need to change the ownership of ``ReadOptions``


spacejam:
:+1:

PR #78, opened 2016-10-27
Implement len() and is_empty() methods on WriteBatch

kaedroho:
Part of #69

These changes we're ported from https://github.com/ngaut/rust-rocksdb/pull/22.

The original implementation named the ``len()`` method ``count()``. I renamed it to ``len()`` to be more conventional with Rust.

``count()`` in Rust usually means that you have to actually "count" something that you do not know the length of, such as an iterator. ``len()`` in Rust usually means that you already know the length and can return it quickly. Looking at the [RocksDB source code](https://github.com/facebook/rocksdb/blob/2c1f95291d0142957d74042aea14682539085501/db/write_batch.cc#L12), ``count`` is just a value stored in memory so I think it makes sense for it to be named ``len()`` in the Rust API.

### DIFF
--- a/src/rocksdb.rs
+++ b/src/rocksdb.rs
@@ -763,6 +763,10 @@ impl WriteBatch {
     pub fn count(&self) -> usize {
         unsafe { rocksdb_ffi::rocksdb_writebatch_count(self.inner) as usize }
     }
+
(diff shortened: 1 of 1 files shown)


spacejam:
:+1:

PR #81, opened 2016-10-28
Doctests for Options methods

kaedroho:
Part of #69, but also applies to existing options methods

### DIFF
--- a/src/rocksdb_options.rs
+++ b/src/rocksdb_options.rs
@@ -114,6 +114,15 @@ impl Options {
     /// `total_threads` is used. Good value for `total_threads` is the number of
     /// cores. You almost definitely want to call this function if your system is
     /// bottlenecked by RocksDB.
+    ///
+    /// # Example
+    ///
+    /// ```
+    /// use rocksdb::Options;
+    ///
+    /// let mut opts = Options::default();
+    /// opts.increase_parallelism(3);
+    /// ```
     pub fn increase_parallelism(&mut self, parallelism: i32) {
         unsafe {
             rocksdb_ffi::rocksdb_options_increase_parallelism(self.inner,
@@ -132,6 +141,15 @@ impl Options {
     /// If true, the database will be created if it is missing.
     ///
     /// Default: false
+    ///
+    /// # Example
+    ///
+    /// ```
+    /// use rocksdb::Options;
+    ///
+    /// let mut opts = Options::default();
+    /// opts.create_if_missing(true);
+    /// ```
     pub fn create_if_missing(&mut self, create_if_missing: bool) {
         unsafe {
             rocksdb_ffi::rocksdb_options_set_create_if_missing(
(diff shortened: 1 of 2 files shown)


kaedroho:
Looks like ``set_wal_recovery_mode`` doesn't work.

src/rocksdb_options.rs
```diff
@@ -578,6 +818,15 @@ impl Options {
...
     /// Default: DBRecoveryMode::PointInTime
+    ///
+    /// # Example
+    ///
+    /// ```
+    /// use rocksdb::{Options. DBRecoveryMode};
+    ///
+    /// let mut opts = Options::default();
+    /// opts.set_wal_recovery_mode(DBRecoveryMode::AbsoluteConsistency);
+    /// ```
```
spacejam: any ideas why this test is failing?

src/rocksdb_options.rs
```diff
@@ -565,6 +796,15 @@ impl Options {
...
     /// Default: false
+    ///
+    /// # Example
+    ///
+    /// ```
+    /// use rocksdb::Options;
+    ///
+    /// let mut opts = Options::default();
+    /// opts.set_report_bg_io_stats(true);
+    /// ```
```
spacejam: any idea why this fails on the 4.5 & 4.1 tests?

kaedroho: Looks like this function was implemented recently. I can't find it in the 4.5 version of the CAPI: https://github.com/facebook/rocksdb/blob/4.5.fb/db/c.cc

Waiting for static linking to be merged would be the easiest solution to this.

src/rocksdb_options.rs
kaedroho: Just turned out to be a typo in the use statement. Fixed now

kaedroho:
Rebased

kaedroho:
``set_wal_recovery_mode`` seems to be broken in some other way on older versions of RocksDB. But everything works on 4.13 so we should be able to merge this after static linking is in.

spacejam:
tests failing on 
```
test rocksdb_options::Options::compression_per_level_0 ... FAILED
test rocksdb_options::Options::set_compression_type_0 ... FAILED
test rocksdb_options::Options::set_wal_recovery_mode_0 ... FAILED
```

kaedroho:
They should be fixed when I rebase this (will quickly do that now)

spacejam:
the same 3 appear to be failing after rebase

kaedroho:
Looks like the errors have become import errors now. I accidentally missed out a couple of ``pub use``'s in https://github.com/spacejam/rust-rocksdb/commit/e34692948eb786f9f4d93424e3f7ff482b82edf3 (they were merged into master after I did the initial work on static linking).

I'll reinstate them in this PR

spacejam:
Cool! I'll merge as soon as tests pass!

spacejam:
Thanks!

PR #82, opened 2016-10-28
Improve naming of some of the new Options methods

kaedroho:
Part of #69 

None of these methods have been in a past released version so I don't think we need to worry about backwards compatibility.

Changes as follows:
 - Renamed ``compression`` to ``set_compression_type``. I think this is a much clearer name
 - Renamed ``add_merge_operator`` to ``set_merge_operator`` and ``add_comparator`` to ``set_comparator``. Both of these methods replace the entire value of their respective option so "add" is the wrong term to use here.
 - Renamed ``set_block_cache_size_mb`` to ``rocksdb_options_optimize_for_point_lookup``. To match RocksDB's API

### DIFF
--- a/src/merge_operator.rs
+++ b/src/merge_operator.rs
@@ -185,7 +185,7 @@ fn mergetest() {
     let path = "_rust_rocksdb_mergetest";
     let mut opts = Options::default();
     opts.create_if_missing(true);
-    opts.add_merge_operator("test operator", test_provided_merge);
+    opts.set_merge_operator("test operator", test_provided_merge);
     {
         let db = DB::open(&opts, path).unwrap();
         let p = db.put(b"k1", b"a");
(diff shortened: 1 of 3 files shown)


kaedroho:
Ahh, totally forgot that ``add_merge_operator`` and ``add_comparator`` have been around for a while.

Should we provide a deprecation path? or scrap the renaming?

kaedroho:
@space If you're happy with the renaming, we could keep the old versions around and decorate them with this: https://github.com/rust-lang/rfcs/blob/master/text/1270-deprecation.md#intended-use

spacejam:
Yeah! I'm in favor of having the old ones pass-through to the new ones, and have deprecation markings on them!

kaedroho:
Cheers @spacejam! Changes made in last commit

spacejam:
Thanks for this!

PR #83, opened 2016-10-28
Remove Options.set_filter_deletes

### LINKED ISSUE #79: Options.set_filter_deletes calls non-existant RocksDB function
``Options.set_filter_deletes`` calls a function in the ffi bindings called ``rocksdb_options_set_filter_deletes``. However, no function of that name is exported by RocksDB itself, leading to a link error (when using static linking).

https://github.com/spacejam/rust-rocksdb/blob/c0e0331d0cb1f652c541d35cebd66de00df26992/src/rocksdb_options.rs#L539
https://github.com/facebook/rocksdb/blob/a10e8a056d569acf6a52045124e6414ad33bdfcd/db/c.cc

Should we remove this method? is there another way to implement it?

kaedroho:
Fixes #79

Deprecated (and removed) as of RocksDB 4.10

https://github.com/facebook/rocksdb/commit/7b79238b655ef6c7ef05c8052c3ded9981c4b54e

### DIFF
--- a/src/ffi.rs
+++ b/src/ffi.rs
@@ -201,7 +201,6 @@ extern "C" {
         options: DBOptions, max_bg_compactions: c_int);
     pub fn rocksdb_options_set_max_background_flushes(options: DBOptions,
                                                       max_bg_flushes: c_int);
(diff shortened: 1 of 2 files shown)


spacejam:
:+1:

PR #84, opened 2016-10-28
Make iterator not copy

kaedroho:
Part of #69 

This pulls in https://github.com/ngaut/rust-rocksdb/pull/9. I've made changes to how the lifetime is defined on `DBIterator` and tweaked the tests to make them work without copying.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -25,7 +25,7 @@ mod rocksdb;
 mod rocksdb_options;
 
 pub use rocksdb::{DB, DBCompactionStyle, DBCompressionType, DBIterator, DBRecoveryMode, DBVector,
-                  Direction, Error, IteratorMode, Snapshot, Writable, WriteBatch, new_bloom_filter};
+                  Direction, Error, IteratorMode, Kv, Snapshot, Writable, WriteBatch, new_bloom_filter};
 pub use merge_operator::MergeOperands;
 
 pub struct BlockBasedOptions {
--- a/src/rocksdb.rs
+++ b/src/rocksdb.rs
@@ -83,7 +83,10 @@ pub struct Snapshot<'a> {
     inner: *const ffi::rocksdb_snapshot_t,
 }
 
-pub struct DBIterator {
+// We need to find a better way to add a lifetime in here.
+#[allow(dead_code)]
+pub struct DBIterator<'a> {
+    db: &'a DB,
     inner: *mut ffi::rocksdb_iterator_t,
     direction: Direction,
     just_seeked: bool,
@@ -94,7 +97,7 @@ pub enum Direction {
     Reverse,
 }
 
-pub type KVBytes = (Box<[u8]>, Box<[u8]>);
+pub type Kv<'a> = (&'a [u8], &'a [u8]);
 
 #[derive(Debug, PartialEq)]
 pub struct Error {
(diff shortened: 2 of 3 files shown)


spacejam:
iterator test is failing, do you know what's up with this?

kaedroho:
Looks like the key isn't being updated between iterations, I haven't been able to reproduce this locally though. Just going to try a bit of debugging through Travis.

kaedroho:
Looks like rocksdb reuses the same buffer to store the key between iterations. So all the slices we are collecting in this test actually point at the same bit of memory. When the assertion is made, the slices all have the same value (the last key).

It might be possible to do some trickery with lifetimes to prevent the keys from being moved outside of the for-loop body (and force the user to copy the keys before collecting them into a vector).

kaedroho:
The value slices may also be a problem as well. There is nowhere where we explicitly "free" them and the iterator could be too large to fit in the cache so I assume RocksDB frees past value slices as it iterates.

If this is the case, this could be a safety issue as the slices we have will be able to live longer than the data they point to.

alexreg:
Maybe worth revising this so that it can work on top of the latest changes to the codebase?

kaedroho:
> Maybe worth revising this so that it can work on top of the latest changes to the codebase?

I'll rebase it tonight. But I'll probably end up abandoning this in favour of proposing an overhaul of the iterator API as I haven't yet had any luck in resolving the issue described above.

kaedroho:
Sorry it took longer than expected. This PR has been rebased on master now

spacejam:
I'm pretty hesitant to introduce this kind of change for the safety of results returned by an iterator. I've [reached out for some guidance about this](https://users.rust-lang.org/t/constrain-iterator-next-return-value-validity-until-next-call-to-iterator-next/7985) and it seems like when we have higher-kinded lifetimes we may be able to constrain the validity of an `Item` until the next call to `next()` but until then I'm not sure we can provide an Iterator implementation that users will not cut themselves painfully with. Maybe we can create an unsafe method on the DBIterator impl itself that can be used by consenting performance seekers in a while let loop?
```while let Some((k, v)) = unsafe { db_iterator.__whatever_you_want_to_name_it__() } {
...
}```

kaedroho:
I've created an alternative iterator API: https://github.com/spacejam/rust-rocksdb/compare/master...kaedroho:new-iterator-api. This API is both safe and copy-free, but backwards incompatibile and uses a while loop instead of a foor loop.

Example usage: https://github.com/kaedroho/rusticsearch/blob/3d080049855045679240ded7dcb90d218789ccc9/kite_rocksdb/src/lib.rs#L550-L553

I think it would be nice to get something like this upstream. Would you be interested in a pull request for this? If so, do you think it would be better to introduce it as a separate API in order to maintain backwards compatibility?

spacejam:
I'm curious how the performance compares for a spectrum of key and value sizes. Someone who spent a lot of their career building various JVM's loved to respond to some of my gut reactions about needless copying "COPYING IS CHEAP!!!" So their voice is in my mind again right now, reminding me of my cognitive bias against copying. I'm in favor of using a no-copy public API underneath a cushier `Iterator` implementation that performs copying to retain expected `Iterator` semantics, but I want to double check our assumptions about the benefit before enriching or modifying the API. I'm happy to do some benchmarks in the next week, but feel free to post some here if you are motivated to do so before I get to it!

matrix: scan 1M {16,4096} byte {key,value} with {1,4,16} cpu cores in-use using {Iterator, no-copy next()} on the same machine.

kaedroho:
@spacejam 
I'm in favour of that approach.

> matrix: scan 1M {16,4096} byte {key,value} with {1,4,16} cpu cores in-use using {Iterator, no-copy next()} on the same machine.

I reckon the most significant factor in performance here is whether or not the keys are already in the cache as we probably won't see much of an improvement if RocksDB has to load the keys off the disk.

derekdreery:
The streaming iterator problem is well known, and can be solved with *associated type constructors*, where the associated type in a Trait is generic over a lifetime. Waiting on this to be implemented in rust-lang/rust https://github.com/rust-lang/rfcs/pull/1598

kaedroho:
Replaced by #112

PR #86, opened 2016-11-02
Use PathBuf/Path type for file paths

kaedroho:
This shouldn't break backwards compatibility for the open/destroy/repair methods. The return type of the ``path()`` method has been changed, however


spacejam:
Looks good! Thank you!

PR #88, opened 2016-11-07
Mainly the addition of the `ffi_try!` macro, plus a few minor things @kaedroho mentioned in the comments for the PR you already accepted.

alexreg:

PR #89, opened 2016-11-09
Make Snapshot public

vmx:
If you want to pass snapshots around, the Snapshot type needs to be
public.


vmx:
This change got in by #88. Hence closing this one.

PR #90, opened 2016-11-09
Pass in WriteBatch by reference and add `clear` method.

vmx:
If you use WriteBatch'es within your project as kind of a buffer you
occasionally flush into RocksDB, it might be part of a struct. It's
difficult to move a single field out of a struct. Hence it is easier
if the methods on WriteBatch would take a WriteBatch by reference
and you call `clear()` on it manually once done.

I can see the point of the original API, though in my use case it's
hard (almost impossible) to use it that way. Though perhaps I'm
missing something and there is a nice way to move a single field
out of a borrowed struct.

### DIFF
--- a/src/rocksdb.rs
+++ b/src/rocksdb.rs
@@ -400,7 +400,7 @@ impl DB {
         &self.path.as_path()
     }
 
-    pub fn write_opt(&self, batch: WriteBatch, writeopts: &WriteOptions) -> Result<(), Error> {
+    pub fn write_opt(&self, batch: &WriteBatch, writeopts: &WriteOptions) -> Result<(), Error> {
         let mut err: *mut c_char = ptr::null_mut();
         unsafe {
             ffi::rocksdb_write(self.inner, writeopts.inner, batch.inner, &mut err);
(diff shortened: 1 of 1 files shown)


spacejam:
I feel like this would result in more bugs.

Does this pattern fit your bill?
```
struct WB;

struct A {
    wb: Option<WB>,
}

impl A {
    fn take_map<F,B>(&mut self, mut f: F) -> Option<B>
    where F: FnMut(WB) -> B {
        self.wb.take().map(|wb| f(wb))
    }
    
    fn with_wb<F, B>(&self, mut f: F) -> Option<B>
    where F: FnMut(&WB) -> B {
        if let Some(ref wb) = self.wb {
            Some(f(wb))
        } else {
            None
        }
    }
}

fn op(_: &WB) {
    println!("operating on wb");
}

fn consume(_: WB) {
    println!("consuming wb");
}

fn main() {
    let mut a = A { wb: None };
    let wb = WB;
    a.wb = Some(wb);
    a.with_wb(|wb| op(wb));
    a.take_map(|wb| consume(wb));
}
```

vmx:
Thanks @spacejam. I didn't know that using Option<> is a way to get things moved out of a borrowed context. Thanks for the hints, it works for me. Thus I'm closing this pull request.

PR #92, opened 2016-11-13
Implemented `std::error::Error` trait for error type.

alexreg:


src/rocksdb.rs
```diff
@@ -117,6 +118,16 @@ impl From<Error> for String {
...
 }
 
+impl error::Error for Error {
+    fn description(&self) -> &str {
+        &self.message
+    }
+
+    fn cause(&self) -> Option<&error::Error> {
+        None
+    }
```
spacejam: https://doc.rust-lang.org/src/std/up/src/libstd/error.rs.html#141

spacejam:
Awesome, thanks for adding this! I'll merge it as soon as you remove the implementation of `cause`!

alexreg:
Done!

PR #95, opened 2016-11-13
Renamed 'librocksdb-sys' to 'rocksdb-sys'.

alexreg:
Also added back `error::Error` impl.


spacejam:
I renamed the FFI to be librocksdb-sys so that we could upload it to crates.io, as rocksdb-sys was already taken.

spacejam:
It's ugly for sure, but it works.

alexreg:
Oh, sorry! I forgot about this.

(I presume the previous owner wasn’t willing to give up the name…)

I’ll let you put in the error::Error impl yourself then.

alexreg:
(And actually, it might be worth accepting this PR, and keeping the dir name as rocksdb-sys, and only change the *crate* name. But up to you of course.)

spacejam:
I'd like to keep them in-sync for now to minimize breakage due to forgotten naming. It's nice to keep all usage of a thing aligned to as few names as possible, to reduce effort to understand usage while doing future work on it. I appreciate this effort, but I'm going to close this PR now. Master now has the error::Error impl in it!

alexreg:
The problem wasn’t master having the Error impl in it, but rather the 0.5.0 branch.

alexreg:
Oh, ignore me. You merged 0.5.0 into master. Probably worth remotely deleting the 0.5.0 branch then. :-)

PR #99, opened 2016-11-20
Added explicit `CompareFn` type for comparator function.

alexreg:
Changed return type of comparator function from `i32` to `std::cmp::Ordering`.

### DIFF
--- a/librocksdb-sys/src/lib.rs
+++ b/librocksdb-sys/src/lib.rs
@@ -958,6 +958,11 @@ pub const rocksdb_fifo_compaction: c_int = 2;
 pub const rocksdb_similar_size_compaction_stop_style: c_int = 0;
 pub const rocksdb_total_size_compaction_stop_style: c_int = 1;
 
+pub const rocksdb_recovery_mode_tolerate_corrupted_tail_records: c_int = 0;
+pub const rocksdb_recovery_mode_absolute_consistency: c_int = 1;
+pub const rocksdb_recovery_mode_point_in_time: c_int = 2;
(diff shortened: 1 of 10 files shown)


alexreg:
By the way, what do you think about changing all the `Default` impls to inherent `new` functions? This would be more in line with the stdlib, and you could also then implement `Default` for `DB`, but have it replace the `open_default` method. Slightly more ergonomic, if you ask me.

alexreg:
@spacejam Mind taking a look at this? :)

spacejam:
Sorry for the delay! This is great! Thank you for adding this functionality and cleaning things up :) 

RE: `open_default()`, `default()`, `new()`, I'm in favor of following the stdlib convention of a `Default` impl that just calls `new()` under the hood for most types. I'm not sure about `Default` for `DB` though, because RocksDB requires a path for persistence, and it feels a little weird to me to assume a path here.  Do you know if other wrappers provide default paths?

alexreg:
Ah, silly me. I meant a `Default` for `DBOptions`.

By the way, minor aesthetic point: should we change `DB`, `DBOptions`, etc. to `Db`, `DbOptions`, etc.? This seems to be the Rust casing convention, even for abbreviations.

spacejam:
Yeah, I'm in favor of adding a `new()` function for `Options` and changing `DB` to `Db` etc... in the future to be more in-line with the stdlib conventions!

alexreg:
Great. I will try to submit a PR for this in the near future…

alexreg:
Been going over some of the code lately, and remembered you return boxed slices rather than vectors from iterators. Is there any specific reason for this?

PR #102, opened 2016-12-16
Compaction filter

tmccombs:
Add set_compaction_filter to Options and compact_range to Db.

### DIFF
--- a/src/db.rs
+++ b/src/db.rs
@@ -16,6 +16,7 @@
 
 use {DB, Error, Options, WriteOptions};
 use ffi;
+use ffi_util::opt_bytes_to_ptr;
 
 use libc::{self, c_char, c_int, c_uchar, c_void, size_t};
 use std::collections::BTreeMap;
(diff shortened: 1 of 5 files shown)


src/db.rs
```diff
@@ -674,6 +675,32 @@ impl DB {
...
+    pub fn compact_range_cf(&self,
+                            cf: *mut ffi::rocksdb_column_family_handle_t,
+                            start: Option<&[u8]>,
+                            end: Option<&[u8]>) {
+        unsafe {
+            ffi::rocksdb_compact_range_cf(self.inner,
+                                          cf,
+                                          opt_bytes_to_ptr(start),
+                                          start.map_or(0, |s| s.len()) as size_t,
+                                          opt_bytes_to_ptr(end),
```
spacejam: I feel a little weird about not taking advantage of the [nullable pointer optimization](https://doc.rust-lang.org/book/ffi.html#the-nullable-pointer-optimization) here, but we're not using it elsewhere right now, and I think it would be sort of a departure from how pointers have been handled so far. Maybe in the future it would be better to try to use it, but for now this is fine IMO.

spacejam:
LGTM, thanks for adding this, @tmccombs! I'll cut a new release in a few minutes with these changes.

spacejam:
version 0.6.0 cut with these changes!

PR #105, opened 2017-01-07
Add some documentation and wrap column_family_t

derekdreery:
The main change is to hide column_family_t behind a containing struct, so there are no raw pointers in the api. Strictly speaking this isn't necesary as the objects are used in an opaque way, but I think it makes the docs nicer.

I've also documented a few things, and marked the DBVector from_c function as unsafe, as the wrong length can cause undefined behaviour

### DIFF
--- a/src/db.rs
+++ b/src/db.rs
@@ -14,7 +14,7 @@
 //
 
 
-use {DB, Error, Options, WriteOptions};
+use {DB, Error, Options, WriteOptions, ColumnFamily};
 use ffi;
 use ffi_util::opt_bytes_to_ptr;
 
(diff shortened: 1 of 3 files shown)


src/db.rs
```diff
@@ -513,6 +513,7 @@ impl DB {
         let cf_handler = unsafe {
             let cf_handler =
                 ffi_try!(ffi::rocksdb_create_column_family(self.inner, opts.inner, cname.as_ptr()));
+            let cf_handler = ColumnFamily { inner: cf_handler };
```
spacejam: This is a bit pedantic, but would you mind naming `cf_handler` something else?

src/db.rs
```diff
@@ -525,14 +526,14 @@ impl DB {
...
+            ffi_try!(ffi::rocksdb_drop_column_family(self.inner, cf.unwrap().inner));
         }
         Ok(())
     }
 
     /// Return the underlying column family handle.
-    pub fn cf_handle(&self, name: &str) -> Option<&*mut ffi::rocksdb_column_family_handle_t> {
-        self.cfs.get(name)
+    pub fn cf_handle(&self, name: &str) -> Option<ColumnFamily> {
+        self.cfs.get(name).map(|cf| *cf)
```
spacejam: `.map(|cf| *cf)` -> `.cloned()` is more idiomatic

derekdreery:
I've made those changes. Thanks for the lib btw!

spacejam:
LGTM, thank you for this contribution :)

PR #109, opened 2017-02-09
A bunch of enhancements I made locally over recent months...

alexreg:


alexreg:
Also, really minor/pedantic point: the project description has no capitalisation, and I think it might be a bit too easy to pass over it on a quick glance. Maybe correct the capitalisation to match the documentation, i.e. "Rust wrapper for RocksDB"? Or maybe this is just me being OCD heh...

alexreg:
@spacejam Ping.

spacejam:
Sorry, I've been having trouble finding time to look at longer PR's. Could we avoid changing the existing API? I want to avoid people getting frustrated because their existing code no longer works.

alexreg:
Okay. The changes are pretty minimal. I hope this could go towards 0.6? You're allowed (nay, encouraged) to make breaking change between minor revisions pre-1.0. :)

alexreg:
Sure I can't persuade you, @spacejam? :)

spacejam:
To be honest I feel kind of like a grumpy old dude. I'd happily accept all
the non-api-changing changes, but I have a bunch of code in my projects
that I'd have to go change as well, and that's less appealing to me unless
it's a significant issue for other people.

alexreg:
Hmm, okay. The API changes are pretty minor though really, aren't they? And I think it's good to change API while it's still 1.0, if they're improvements... because after 1.0 it gets a lot harder. We could always backport them too. :)

daboross:
@alexreg Even with these API changes, it'd probably be best to have those as a separate PR, so each change can be discussed? Generally with contributions of this size, there'd be motivation / reasoning presented for the breaking changes, and the benefits/drawbacks can be weighed.

It'd be great to have the non-breaking-changes in as well so that they can be used before the project decides to jump to the next major version. Even for a pre-1.0 crate, it would be best to consolidate all of the breaking changes into a single major version jump, and having them mixed up with other additions does prevent this somewhat.

alexreg:
@daboross I think that's far too formal, myself. Oh well, I may just have to fork the project....

spacejam:
It's pretty standard etiquette to keep PR's focused on a single change. Single PR's with many unrelated changes are a huge pain to review.

alexreg:
Sorry. I hoped the individual commits would be good enough. Don't even know how to use Git to factor out the different PRs!

daboross:
@alexreg I'm sure it's fine, it's just hard to comment on the single PR.

I can help you out with splitting it up if you'd want - shouldn't be too hard with `git`

alexreg:
@daboross Thanks, that would be appreciated. That is, if @spacejam is still interested. :)

daboross:
@alexreg no problem! I'd probably be able to help more in a live-chat like on irc, but I'll give some directions for what I can here! Hope I don't go into too much detail.

To start, it is nice that you have them separated into commits - and this will make the next part significantly easier. In your clone, it seems the latest commit before your changes was https://github.com/alexreg/rust-rocksdb/commit/5e3f4351a9381f5743f0455acf7b5f7587268f68. To split things out, we'll use this hash (`5e3f4351a9381f5743f0455acf7b5f7587268f68`) to specify when to start.

There's probably a more simple way to do this using 'reflog', but as I don't have real knowledge of that sub-command, here's what I would do:

First step: create one "patch file" for each commit, in the directory above your project clone:

```
git format-patch 5e3f4351a9381f5743f0455acf7b5f7587268f68..HEAD -o ../
```

This will create one .patch file for each commit - we can then use these patch files individually to make different pull requests.

I see you still have your `master` branch, so I'd then checkout on that, and then start a new branch for the first change you want to do:

```
# reset branch to master
git checkout master
# create new branch with master as starter
git checkout -b cleanup-01
```

then you can apply one or more of the patch files to this:

```
git am -3 ../0001-*
```

I use `0001-*` here because I find that easier to type, that will just match any file starting with `0001` (which likely will just be a single file, the first patch).

After you've applied all the patches you want to this branch, you can push it to github, and then checkout back to 'master' in order to do the next branch:

```
git push --set-upstream origin cleanup-01
git checkout master
git checkout -b cleanup-02
# ...
```

After you push the branch, github will give you an option to make a new PR for it when you go to either your project or this project's homepage.

I'd recommend making probably 1 PR per change, with the exception of making the "Updated *rocksdb*..." commits one PR, and possibly making the breaking changes (name styles, in-range remove, removing _opts, fix visibility of comparator) one PR.

Keep in mind that not all changes may be merged, and some (like https://github.com/spacejam/rust-rocksdb/pull/109/commits/bba796fff569ceb6e7ec4a576ac2460c00979fef) might be considered not enough benefit to justify the change. Even so, having the breaking changes as open PRs might be good for the project so that they can be merged if/when RocksDB chooses break to a new major version. Breaking changes do hurt an ecosystem, so it is valid for the author to delay renames/etc. which break without a feature benefit until there's also a breaking change which would allow for more functionality.

In any case, thank you for making these changes! I'm definitely looking forward to them being considered and the ones that will be, merged.

alexreg:
@daboross Many thanks for that! I'm on IRC all the time (Freenode, nick *Noldorin*), so maybe we can chat there at some point, but this should be good to start. I'll probably just wait to confirm @spacejam is interested in this first though.

spacejam:
I'm interested in getting all non-api-breaking improvements in asap, and for things that do break the API, I'm happy to see what other people think about them in small, specific PR's. I'm not super in-favor of changing api's, but if there is support from others then we can include them in a future major version.

PR #110, opened 2017-02-16
Fail if RocksDB or Snappy source is missing

vmx:
When I tried to build via `cargo build` I saw a lot of error messages and had no clue what was wrong. I hope with this patch things get a little bit clearer if people run into similar issues.


spacejam:
Thanks!

PR #111, opened 2017-02-16
Makefiles use tabs

vmx:
The indentation in Makefiles is tabs. Without this change it failed for me with:

    Makefile:4: *** missing separator.  Stop


spacejam:
Thanks for fixing this!

PR #112, opened 2017-02-18
Raw iterators

kaedroho:
This pull request implements an alternative iterator API that closely matches the underlying the iterators API in RocksDB. It gives access to all the flexibility and performance of RocksDB but the API it implements is quite different to what Rust developers are used to.

This can replace both of by earlier iterator pull requests #70 and #84

Example usage:

```rust
let mut iter = db.raw_iterator();
iter.seek_to_first();
while iter.valid() {
    let key = iter.key().unwrap();
    let value = iter.value().unwrap();

    // ...

    iter.next();
 }

// Reverse iteration
let mut reverse_iter = db.raw_iterator();
reverse_iter.seek_to_last();
while reverse_iter.valid() {
    let key = reverse_iter.key().unwrap();
    let value = reverse_iter.value().unwrap();

    // ...

    reverse_iter.prev();
 }

// Fowards iteration from a specified key
let mut seek_iter = db.raw_iterator();
seek_iter.seek(b"foo");
while seek_iter.valid() {
    let key = seek_iter.key().unwrap();
    let value = seek_iter.value().unwrap();

    // ...

    seek_iter.next();
 }
```

Here's an example of how this API can be misused:

```rust
let mut iter = db.raw_iterator();

let key1 = unsafe { iter.key_inner().unwrap() };

iter.next();
let key2 = unsafe { iter.key_inner().unwrap() };

println!("{} {}", key1, key2);  // Prints the value of key2 twice
```
(note: this only happens with the ``*_inner()`` methods as ``.key()`` and ``.value()`` always copy)

Also, forgetting to call ``.next()`` or ``.prev()`` will cause the application to hang.

TODO:

 - [x] Safe versions of ``.key()`` and ``.value()`` that copy (I'll probably make the safe versions the default and rename the unsafe ones to something like ``.key_inner()`` and ``.value_inner()``)
 - [x] Implement DBIterator on top of DBRawIterator
 - [x] Tests
 - [x] Documentation

### DIFF
--- a/librocksdb-sys/src/lib.rs
+++ b/librocksdb-sys/src/lib.rs
@@ -294,8 +294,6 @@ extern "C" {
 
     pub fn rocksdb_iter_seek(iterator: *mut rocksdb_iterator_t, k: *const c_char, klen: size_t);
 
-    pub fn rocksdb_iter_seek_for_prev(iterator: *mut rocksdb_iterator_t, k: *const c_char, klen: size_t);
-
     pub fn rocksdb_iter_next(iterator: *mut rocksdb_iterator_t);
 
     pub fn rocksdb_iter_prev(iterator: *mut rocksdb_iterator_t);
--- a/librocksdb-sys/src/lib.rs
+++ b/librocksdb-sys/src/lib.rs
@@ -294,6 +294,8 @@ extern "C" {
 
     pub fn rocksdb_iter_seek(iterator: *mut rocksdb_iterator_t, k: *const c_char, klen: size_t);
 
+    pub fn rocksdb_iter_seek_for_prev(iterator: *mut rocksdb_iterator_t, k: *const c_char, klen: size_t);
+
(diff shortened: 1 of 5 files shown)


spacejam:
I am a fan of exposing this functionality, and am happy that the functions that return exploding results are marked as unsafe. :+1: I'm wondering about rewriting the normal DBIterator on top of the DBRawIterator, to avoid duplicated code. That can be a separate piece of work, unless you are motivated to go down that path. Thanks for putting this together, it will be nice to use for consenting performance seekers :)

kaedroho:
Thanks @spacejam! Sorry, I haven't had much spare time recently to work on this. I'll close my previous PRs now marking this as the agreed way forward.

> I'm wondering about rewriting the normal DBIterator on top of the DBRawIterator, to avoid duplicated code.

I think this is worth doing. I'll hopefully get time to look at this soon

kaedroho:
This is now ready for review!

spacejam:
Awesome work :) I'll cut a new version today with these changes!

PR #117, opened 2017-03-13
Split 0.6.1 release note into its own section

kaedroho:
Makes it clear to 0.6.0 users that they don't have this feature.


spacejam:
Ahh yeah I should have done this, thanks for the fix!

PR #120, opened 2017-04-14
Fixed race condition in tests

debris:
on my machine (osx 10.12.3) running tests very often fails with

```
thread 'main' panicked at 'called `Result::unwrap()` on an `Err` value: Error { message: "IO error: lock path/for/rocksdb/storage6/LOCK: Resource temporarily unavailable" }', /Users/rustbuild/src/rust-buildbot/slave/stable-dist-rustc-mac/build/src/libcore/result.rs:868
stack backtrace:
```

This error is caused by race condition, cause tests in both `lib.rs` and `db.rs` are reusing database at the same location. This pr fixes that


spacejam:
Thanks for fixing this!

PR #121, opened 2017-06-09
Implement Clone and AsRef<str> for Error

daboross:
These seem like useful functions we could add without much downside.


spacejam:
LGTM, thanks!

PR #125, opened 2017-06-19
Move tests to the default `tests` directory

vmx:
Instead of specifying a custom test directory, use the default
Rust way of doing things (which is having a `tests` directory.


spacejam:
Thanks!

PR #123, opened 2017-06-13
Make rocksdb work with gcc 7 (v5.4.6)

derekdreery:
Note that this upgrade causes some tests to fail in ffi, but I lack the skills to fix them.

PR #129, opened 2017-07-25
Update rocksdb

kaedroho:
Based on #123 with a compatibility fix and implementation of ``seek_for_prev`` for ``RawIterator``


kaedroho:
Now merged into #123

PR #133, opened 2017-08-19
use bindgen to generate ffi

benesch:
Closes #128.

PR #139, opened 2017-08-29
Remove `set_disable_data_sync`

glittershark:
This is removed from rocksdb core as of facebook/rocksdb@eb912a9, and having the reference in the ffi causes loading the shared library to break.

The removal of this function is covered already by #133, but I figured I'd submit this separately in order to get things working for us until the build for that is fixed.


spacejam:
Thanks for this! I'll push out a new release that includes these changes.

spacejam:
publishing crate `0.8.0` now

PR #141, opened 2017-09-05
list column families

jeizsm:
Add list_cf method

### DIFF
--- a/src/db.rs
+++ b/src/db.rs
@@ -671,18 +671,17 @@ impl DB {
         };
 
         let mut length = 0;
-        let vec;
 
         unsafe {
             let ptr = ffi_try!(ffi::rocksdb_list_column_families(opts.inner,
                                                                  cpath.as_ptr() as *const _,
                                                                  &mut length));
 
-            vec = Vec::from_raw_parts(ptr, length, length).iter().map(|&ptr| {
+            let vec = Vec::from_raw_parts(ptr, length, length).iter().map(|&ptr| {
                 CString::from_raw(ptr).into_string().unwrap()
             }).collect();
+            Ok(vec)
         }
-        Ok(vec)
     }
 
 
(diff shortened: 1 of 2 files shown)


src/db.rs
```diff
@@ -659,6 +659,33 @@ impl DB {
...
+        unsafe {
+            let ptr = ffi_try!(ffi::rocksdb_list_column_families(opts.inner,
+                                                                 cpath.as_ptr() as *const _,
+                                                                 &mut length));
+
+            vec = Vec::from_raw_parts(ptr, length, length).iter().map(|&ptr| {
+                CString::from_raw(ptr).into_string().unwrap()
+            }).collect();
+        }
+        Ok(vec)
```
spacejam: You can have `Ok(vec)` be the last line in the unsafe block, and the function will return it there. No need to declare `vec` on line 674 before assigning it.

tests/test_column_family.rs
```diff
@@ -62,6 +62,17 @@ pub fn test_column_family() {
             Err(e) => panic!("failed to open db with column family: {}", e),
         }
     }
+
+    // should be able to list a cf
+    {
+        let opts = Options::default();
+        let vec = DB::list_cf(&opts, path);
+        match vec {
+            Ok(vec) => println!("list cf: {:?}", vec),
```
spacejam: Could you change the `println!` to an `assert_eq!` that ensures the vec matches what you expect it to be?

spacejam:
Awesome work! Just a couple small changes and this is good for merging :)

spacejam:
Thanks for adding this! I'll cut a new version with these changes now!

spacejam:
`0.8.1` is up on crates.io with these changes :)

jeizsm:
Thank you :)

PR #143, opened 2017-10-19
bindgen - tidying up benesch's bindgen PR

rrichardson:
* merged up to master
* Fixed up travis for bindgen (clang dependencies) 
* hid max_align_t from bindgen due to (this bug) [https://github.com/rust-lang-nursery/rust-bindgen/issues/550]
* tweaked tests to pass
* Oh almost forgot.  I got tired of rustfmt adding a trailing comma and breaking try_ffi, so I tweaked try_ffi to require a trailing command and adjusted the rustfmt.toml accordingly (there are more vertical function calls than not, so it's better this way ;) )

PR #146, opened 2017-10-22
Non-Associative merge ops

rrichardson:
This set of changes depends on the bindgen PR #143

The existing MergeOperator scheme assumes that values are associative and that an existing value is always present in the full_merge operation.  

This PR adds the ability to specify a partial_merge that can be None.  It also does a value check on the merge callback for existing value, and supplies none to the Rust callback if it is null or 0 length. 

This is a minor change to the API, so it might be worth looking into backwards compatible changes.

### DIFF
--- a/librocksdb-sys/Cargo.toml
+++ b/librocksdb-sys/Cargo.toml
@@ -22,6 +22,6 @@ libc = "0.2"
 const-cstr = "0.2"
 
 [build-dependencies]
-cc = { version = "^1.0", features = ["parallel"] }
+cc = { git = "https://github.com/alexcrichton/cc-rs", features = ["parallel"] }
 make-cmd = "0.1"
 bindgen = "0.29"
--- a/src/db_options.rs
+++ b/src/db_options.rs
@@ -200,10 +200,16 @@ impl Options {
         }
     }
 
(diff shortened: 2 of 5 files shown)


rrichardson:
@spacejam -  Care to have a look at this?  It is a slight API break, but I'm not sure if there is a better way around it.   It has some decent tests, and I have been using this code in production for a while, so am fairly sure it is stable :) 

This depends on bindgen and cc-rs updates, which are committed.

spacejam:
@rrichardson sorry for the huge delay on this, lots of chaos happening as I move continents. Is there a reason to have non-partial merge functions ever return `None`?

rrichardson:
@spacejam -  The case I had in mind was bad data.  My particular use case is building data structures via merge operations, in a CRDT style fashion.  However, if the user messes up and inserts data that doesn't fit the schema, then it should fail. (maybe it should be louder about that failure) 

Also with regards to actual CRDTs, there could be conflicts, at which point an operation cannot be applied.. maybe it should fail at that point as well.  

Unfortunately the only way to get the necessary information back to the calling code would be to leverage the error log, which might be a future exercise. 

see :  https://github.com/QuiverMedia/rocks-cache/blob/master/src/collections.rs#L380

spacejam:

Issue #137, opened 2017-08-29
investigate speeding up builds by short-circuiting librocksdb-sys c++ compilation when unchanged


spacejam:

rrichardson:
This is a good idea,  until then, people might experience a much nicer build experience after https://github.com/alexcrichton/cc-rs/pull/261

iSynaptic:
Is this still an open issue, or was it solved by alexcrichton/cc-rs#261?

iSynaptic:
@rrichardson I don't experience a rebuild of `librocksdb-sys` every time - only the initial build for a given `Cargo.toml`.  Can you share an example of where/how this is happening?

iSynaptic:
Not sure how to move this forward and this issue has been fairly inactive for some time. I'm going to close. We can always revisit if needed...

PR #147, opened 2017-10-24
Update dependency on cc (previously gcc) crate

glittershark:
Rename the gcc crate to cc per alexcrichton/cc-rs@410992f, and update the one breaking change from the `Config` struct to `Build`, per alexcrichton/cc-rs@dc4bfb4.

The primary impetus for this change is to pull over alexcrichton/cc-rs@2b1ce98, which *drastically* speeds up build times of rocksdb-sys (20m to 1m, on my machine) - see the discussion on #137.


rrichardson:
My motivation for tracking down that parallelism bug in cc-rs was, in fact, rust-rocksdb :)

spacejam:
Thanks!

PR #151, opened 2017-11-19
Preliminary: Allow creating iterators over prefixes

glittershark:
This is a proposal for an API that exposes the ability to limit (db or cf) iterators to key prefixes by using the `prefix_same_as_start` read option. It won't work at all until (or unless!) https://github.com/facebook/rocksdb/pull/3185 lands, but I figured I'd open it here to start discussion on the proposed API.

### DIFF
--- a/src/db.rs
+++ b/src/db.rs
@@ -872,6 +872,12 @@ impl DB {
         DBIterator::new(self, &opts, mode)
     }
 
+    pub fn prefix_iterator<'a>(&self, prefix: &'a [u8]) -> DBIterator {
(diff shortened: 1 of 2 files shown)


spacejam:
Please excuse the terrible delay on this! Thank you for adding this functionality :)

spacejam:

spacejam:
fyi @glittershark this was required to get the tests to pass https://github.com/spacejam/rust-rocksdb/pull/163/files#diff-ab2b8433bbcfad4f50dd22e663bf4512L166

PR #153, opened 2017-12-08
fix windows build (bump rocksdb to 5.7.1)

### LINKED ISSUE #132: Failed build on Windows platform
I am trying to build my project that uses rocksdb againt Windows platform, but getting build errors during compilation of rocksdb.
Here is [build log](https://github.com/spacejam/rust-rocksdb/files/1206432/log.txt)

jquesnelle:
Fixes #132.

#132 was caused by facebook/rocksdb#2709 and fixed in facebook/rocksdb@6a9de434775abb2868132e2d438cfe932e225032. I've updated the submodule to rocksdb 5.7.1 (facebook/rocksdb@acf935e40f9d6f4c3d13c7d310def7064c1f1c95), which includes the fix. This is also the very next public release after the version that is currently being used (5.6.2).

However, just using rocksdb 5.7.1 doesn't completely solve the problem. `env_posix.cc` and `io_posix.cc` were moved from `env/` to `util/`, so I changed this in `build.rs`. Also, 68f6a123df8c3e688642eb34919007a56d478773 added `-Wno-unused-parameter` as a global command line flag, which causes an error when building with MSVC. I've moved it up into the non-MSVC conditional.


brson:
Ping @spacejam. Needing this patch.

spacejam:
Thank you for this! I'll cut a release now that includes this. Sorry for the huge delay on merging, I've been underwater with the logistics of moving away from berlin and then back to berlin short-notice. Would not recommend double 0-net-distance moving.

spacejam:
0.8.2 has been published with these changes

brson:
Thanks a bunch @spacejam!

PR #160, opened 2018-02-02
Expose advise_random_on_open option

ongardie-ebay:
I found this knob to be quite helpful for improving sequential iterator performance.


ongardie-ebay:
(fc40d10 updated the comment)

spacejam:
awesome, thanks!

Issue #164, opened 2018-02-12
Link error on Windows


PiedmontStream:
rocksdb link error.
add `port/win/win_thread.cc` into the file `rust-rocksdb/librocksdb-sys/rocksdb_lib_sources.txt` can fix it.
the methods of `rocksdb::port::WindowsThread` were lost on windows.

gagarin55:
@spacejam Could you please apply this fix and release new version of librocksdb-sys crate

iSynaptic:
This should be closed.

Issue #155, opened 2018-01-15
Can't use on Windows


gagarin55:
I'm trying to compile empty bin project with this crate as dependency on Windows. I'm getting such error:
```
error: failed to run custom build command for `librocksdb-sys v5.7.1`
process didn't exit successfully: `C:\Users\вася\Projects\test1\target\debug\build\librocksdb-sys-d4cbcc9a730ac9e2\build-script-build` (exit code: 101)

--- stdout
cargo:rerun-if-changed=build.rs
cargo:rerun-if-changed=rocksdb/

--- stderr
thread 'main' panicked at 'Unable to find libclang: "couldn\'t find any of [\'clang.dll\', \'libclang.dll\'], set the LIBCLANG_PATH environment variabl
e to a path where one of these files can be found (skipped: [])"', src\libcore\result.rs:906:4
note: Run with `RUST_BACKTRACE=1` for a backtrace.
```

spacejam:
Hey, вася, thanks for opening this issue. Were you able to make any progress by setting that `LIBCLANG_PATH` environment variable? My access to a windows machine is a bit limited at the moment, but I would like to address this issue for windows users.

gagarin55:
Do I need install this tool http://releases.llvm.org/download.html to build rust-rocksdb ?

gagarin55:
After installing CLang I've got another error:
```
error: linking with `link.exe` failed: exit code: 1120
  |
[...]

PR #165, opened 2018-02-13
fix link error on windows

PiedmontStream:
add  `port/win/win_thread.cc` to `rust-rocksdb/librocksdb-sys/rocksdb_lib_sources.txt`
for #155 #164


gagarin55:
May be it is better to fix this place? https://github.com/spacejam/rust-rocksdb/blob/701669d0763a0fb30708d36b1af2e152bda1e63d/librocksdb-sys/build.rs#L102

PR #166, opened 2018-02-19
Link with system rocksdb and snappy libs through envvars

ozkriff:
This PR adds 4 env variables: `ROCKSDB_LIB_DIR`, `SNAPPY_LIB_DIR`, `ROCKSDB_STATIC` and `SNAPPY_STATIC`.

What do you think of this? :)

(related to https://github.com/spacejam/rust-rocksdb/issues/138)

### DIFF
--- a/librocksdb-sys/build.rs
+++ b/librocksdb-sys/build.rs
@@ -29,10 +29,7 @@ fn fail_on_empty_directory(name: &str) {
     }
 }
 
-fn build_rocksdb() {
-    println!("cargo:rerun-if-changed=build.rs");
-    println!("cargo:rerun-if-changed=rocksdb/");
-
+fn bindgen_rocksdb() {
     let bindings = bindgen::Builder::default()
         .header("rocksdb/include/rocksdb/c.h")
         .hide_type("max_align_t") // https://github.com/rust-lang-nursery/rust-bindgen/issues/550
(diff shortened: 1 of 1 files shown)


DarkEld3r:
I want to add more details: we are currently using [fork](https://github.com/exonum/exonum_rocksdb) of `rust-rocksdb` for our [project](https://github.com/exonum/exonum), but we prefer to switch to your crate. Still we need several things to be done and this is one of them. Thus, we are very interested in any feedback.

spacejam:
Sorry for the huge delay on this! Just started a new job and I've been getting used to a new schedule. I like this approach! I'll cut a new version in a few minutes that includes this change :)

spacejam:
I've published rocksdb `0.10.0` which includes these changes!

DarkEld3r:
@spacejam Thank you! Probably, there will be several more things that we need before switching to your crate, but we are looking forward to that.

jonhoo:
It'd be really neat if `rocksdb-sys` checked if `rockdb` and `snappy` were available in the standard OS locations too. It's a little unfortunate to have to pass custom environment variables every time we build even though those libraries are installed system-wide.

jonhoo:
Tangentially related to my comment above — this should probably be mentioned in the docs.

PR #169, opened 2018-02-20
Fix reverse iteration from a given key

ongardie-ebay:
Now when you reverse iterate past the last key, it goes backwards from the end.
When you reverse iterate before the first key, you get nothing out.

Note: This is a breaking change if users have come to depend on the old
behavior.

### DIFF
--- a/src/db.rs
+++ b/src/db.rs
@@ -466,10 +466,13 @@ impl DBIterator {
                 self.raw.seek_to_last();
                 self.direction = Direction::Reverse;
             }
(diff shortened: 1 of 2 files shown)


spacejam:
Sorry for the huge delay on this! Just started a new job and have been getting used to a new schedule. Thanks for fixing this! I'll cut a new version in a few minutes that includes these changes :)

spacejam:
I've published rocksdb `0.10.0` which includes these changes!

PR #170, opened 2018-02-25
improve multithreaded db.rs

Elzor:
When the first thread create iterator, and than the second thread is using it I got issue:
```
 `*mut librocksdb_sys::rocksdb_iterator_t` cannot be sent between threads safely. 
```
I have fixed this issue with DBIterator. Now we can use iterator that was created in some thread in any other threads.


spacejam:
Hey @Elzor ! 

Looking at [the rocksdb iterator source code](https://github.com/facebook/rocksdb/blob/8ada876dfe2356a9f078ce17aa51057eeb8f17b9/include/rocksdb/iterator.h#L14) they say:

```
// Multiple threads can invoke const methods on an Iterator without
// external synchronization, but if any of the threads may call a
// non-const method, all threads accessing the same Iterator must use
// external synchronization.
```

So I don't believe this is safe.

sfackler:
What non-const methods can be called on a `&DBIterator`?

Elzor:
Actually, I thought that iterators are safe and run in snapshots because in general iterator is consistent view. Are snapshots safe?

iSynaptic:
@sfackler Non-const methods would be ones that mutate the underlying state of the iterator.

@Elzor I'm comfortable implementing `Send` here, but not `Sync`. Once `Send` is implemented, you can wrap it with a smart pointer that adds `Sync`, for example  `Arc<Mutex<DBIterator>>`.

Elzor:
@iSynaptic I agree with you. 'Send' method is enough for implementation.

iSynaptic:
@Elzor Great! Go ahead an update the PR to be `Send` only and I'll approve it.

Elzor:
@iSynaptic I've updated my PR. Please merge it

PR #176, opened 2018-03-30
Change DB::open_cf signature to the more generic one

### LINKED ISSUE #175: DB::open_cf signature
It is inconvenient to use `open_cf` function with the result of `list_cf`. It requires transforming `Vec<String>` to `&[&str]`. The signature can be made more generic the following way:
```rust
pub fn open_cf<P: AsRef<Path>, N: AsRef<str>>(opts: &Options, path: P, cfs: &[N]) -> Result<DB, Error>
```

DarkEld3r:
Closes #175.

### DIFF
--- a/src/db.rs
+++ b/src/db.rs
@@ -587,6 +587,15 @@ impl ColumnFamilyDescriptor {
 
 impl DB {
     /// Open a database with default options.
+    ///
+    /// # Examples
+    /// 
+    /// ```no_run
+    /// use rocksdb::DB;
+    ///
+    /// let path = "path/to/the/database";
+    /// let db = DB::open_default(path).unwrap();
+    /// ```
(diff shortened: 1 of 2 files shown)


DarkEld3r:
This is a little bit controversial, but I suppose that `open_cf` function is mostly used in combination with `list_cf` and it will be a little more optimal to pass `IntoIterator` in that case.

iSynaptic:
> 
> 
> This is a little bit controversial, but I suppose that `open_cf` function is mostly used in combination with `list_cf` and it will be a little more optimal to pass `IntoIterator` in that case.

@DarkEld3r Curious: What makes this change controversial?

iSynaptic:
Closing due to branch conflicts and inactivity.

DarkEld3r:
@iSynaptic Sorry, I somehow missed your question. I suppose it isn't actual anymore and I don't remember all details already, but I wasn't sure if such signature change would be OK.

PR #179, opened 2018-04-03
Add get_cf_by_name method to `Snapshot`

andreyluchin:
Right now to get data from a column family in a `Snapshot` you need the handle (`ColumnFamily`). To get a handle, you must consult the `DB` object with column family name.
This PR simplifies the process by adding `get_cf_by_name` method which allows you to directly access data in a column family with a name instead of handle.

### DIFF
--- a/src/db.rs
+++ b/src/db.rs
@@ -838,6 +838,10 @@ impl DB {
         self.get_cf_opt(cf, key, &ReadOptions::default())
     }
 
+    pub fn get_cf_by_name<T: AsRef<str>>(&self, name: T, key: &[u8]) -> Option<Result<Option<DBVector>, Error>> {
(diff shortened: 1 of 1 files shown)


src/db.rs
```diff
@@ -838,6 +838,10 @@ impl DB {
         self.get_cf_opt(cf, key, &ReadOptions::default())
     }
 
+    pub fn get_cf_by_name<T: AsRef<str>>(&self, name: T, key: &[u8]) -> Option<Result<Option<DBVector>, Error>> {
```
spacejam: Can you remove the outer wrapping `Option`?

iSynaptic:
@andreyluchin Their is a build error with the PR. Is it something you can take a look at?

iSynaptic:
Closing due to inactivity and branch conflicts.

PR #180, opened 2018-04-05
Add support for customizing the memtable factory

ekmartin:
See https://github.com/facebook/rocksdb/wiki/MemTable for more information.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -136,6 +136,14 @@ pub struct BlockBasedOptions {
     inner: *mut ffi::rocksdb_block_based_table_options_t,
 }
 
+/// Defines the underlying memtable implementation.
+/// See https://github.com/facebook/rocksdb/wiki/MemTable for more information.
+pub enum MemtableFactory {
+    Vector,
+    HashSkipList { bucket_count: usize, height: i32, branching_factor: i32 },
+    HashLinkList { bucket_count: usize }
+}
+
 /// Database-wide options around performance and behavior.
 ///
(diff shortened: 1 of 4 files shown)


src/lib.rs
```diff
@@ -136,6 +136,14 @@ pub struct BlockBasedOptions {
     inner: *mut ffi::rocksdb_block_based_table_options_t,
 }
 
+/// Defines the underlying memtable implementation.
+/// See https://github.com/facebook/rocksdb/wiki/MemTable for more information.
+pub enum MemtableFactory {
+    Vector,
+    HashSkipList { bucket_count: usize, height: i32, branching_factor: i32 },
```
ekmartin: wasn't sure if these should be defined with `libc` types or not - let me know what you prefer.

ekmartin:
Also added `db.full_iterator()` which sets `total_order_seek` to `true` before iterating. This is needed for hash based memtable implementations where you want to iterate across different prefixes.

This is what the [RocksDB docs](https://github.com/facebook/rocksdb/wiki/Prefix-Seek-API-Changes) say about it:
> Performance might be worse in this mode. Please, be aware that not all implementations of prefix seek support this option. For example, some implementations of PlainTable doesn't support it and you'll see an error in status code when you try to use it. Hash-based mem tables might do a very expensive online sorting if you use it. This mode is supported in prefix bloom and hash index of block based tables.

src/lib.rs
spacejam: I prefer the way you chose, by using rust types where possible for the public API.

spacejam:
Thanks!

PR #182, opened 2018-04-05
Update rustfmt settings

ekmartin:
Since `always` isn't wrapped in quotes the current rustfmt settings crashes with a `Could not parse TOML: invalid number at line 4`. It doesn't look like the rest of the codebase follows `trailing_comma = "Always"` anyway, so I took the liberty of removing it (together with the width related settings, which follow the default value anyway).

Also ran `cargo fmt` with the latest `rustfmt-preview`. Feel free to close this if you'd rather run `rustfmt` yourself and push to master though - completely understandable!


iSynaptic:
#229 incorporates these changes, closing.

PR #183, opened 2018-04-06
Make sure DB is dropped after all tests

ekmartin:
Adds a `DBName` struct that calls `DB::Destroy` in `drop`. Also suffixes the given `prefix` with a timestamp to ensure that subsequent test runs don't reuse old databases in case of panics prior to `drop` being called.


iSynaptic:
I like this.  Would you consider implementing `Deref` and `AsRef<str>` for the DBName struct?  That way it isn't needed to `.name` into the struct to pass the `&str` to the `open*` and `list_cf` functions.

@aleksuss @vitvakatu Thoughts?

vitvakatu:
Yep, `AsRef`  would definitely be nice

iSynaptic:
I've not heard anything from Martin in a while, both this PR or on discussions about #209. I'm going to close this PR and open another one to get this across the finish line.

PR #187, opened 2018-04-09
Expose compaction_readahead_size option

romanz:
From https://github.com/facebook/rocksdb/blob/bde1c1a72afe365240004d5b75d61004618110bb/include/rocksdb/options.h#L687-L695.


spacejam:
Thanks! Today I'll cut a point release on the 0.10 branch that includes these changes!

PR #193, opened 2018-07-14
Update RocksDB to 5.14.2

ekmartin:


spacejam:
Thanks!

spacejam:
@ekmartin currently blocked on cutting a release that includes this change:

```
   Compiling librocksdb-sys v5.14.2                                                                                                                                                                                                                           
error: expected one of `crate`, `fn`, `pub`, `static`, `type`, or `}`, found `;`
    --> /home/t/src/rust-rocksdb/target/package/rocksdb-0.10.1/target/debug/build/librocksdb-sys-882744c53a3eb87c/out/bindings.rs:1166:5                                                                                                                      
     |                                               
1165 | extern "C" {                               
     |             - expected one of `crate`, `fn`, `pub`, `static`, `type`, or `}` here
1166 |     ;                           
     |     ^ unexpected token
          
error: aborting due to previous error
```

ekmartin:
@spacejam I might be doing something wrong, but I can't seem to reproduce this: `cargo build` completes successfully using the nightly version from May 19th (also tried with a newer one). Which version are you using?

spacejam:
I'm not sure exactly what was happening, but I was able to resolve this by upgrading from rust stable 1.26 to 1.27.1. On 1.26, librocksdbsys was building fine on its own but I couldn't build it as a dependency for the top level rocksdb crate. Marking this in my mind as a weird thing, but not spending more effort on it for now. Thanks for trying to reproduce!

PR #196, opened 2018-07-27
link error fix

tempbottle:
fix link error on windows when run `cargo test`.
 [https://docs.microsoft.com/en-us/windows/desktop/api/shlwapi/nf-shlwapi-pathisrelativea](url)


iSynaptic:
Fixed by #219.

PR #201, opened 2018-08-07
Allow retrieving WriteBatch serialized size

romanz:


src/db.rs
```diff
@@ -1106,6 +1106,15 @@ impl WriteBatch {
         unsafe { ffi::rocksdb_writebatch_count(self.inner) as usize }
     }
 
+    /// Return WriteBatch serialized size (in bytes).
+    pub fn size(&self) -> usize {
+        unsafe {
+            let mut batch_size: size_t = 0;
+            ffi::rocksdb_writebatch_data(self.inner, &mut batch_size);
+            batch_size as usize
+        }
```
aleksuss: @romanz `size` is ambiguous name of the method as for me.

romanz: @aleksuss You're right.
Should I change it to `serialized_size`, `size_in_bytes` or anything else?

aleksuss: @romanz I guess `size_in_bytes` looks good.

PR #202, opened 2018-08-07
Expose 'set_skip_stats_update_on_db_open' option

romanz:
Following the suggestion from https://github.com/facebook/rocksdb/wiki/RocksDB-Tuning-Guide#spinning-disks-usually-provide-much-lower-random-read-throughput-than-flash

### DIFF
--- a/src/db_options.rs
+++ b/src/db_options.rs
@@ -1098,6 +1098,17 @@ impl Options {
             ffi::rocksdb_options_set_wal_dir(self.inner, p.as_ptr());
         }
     }
+
+    /// If true, then DB::Open() will not update the statistics used to optimize
+    /// compaction decision by loading table properties from many files.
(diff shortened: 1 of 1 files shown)


src/db_options.rs
```diff
@@ -1098,6 +1098,17 @@ impl Options {
             ffi::rocksdb_options_set_wal_dir(self.inner, p.as_ptr());
         }
     }
+
+    // If true, then DB::Open() will not update the statistics used to optimize
```
aleksuss: @romanz isn't it better to use doc comments here, is it ?

romanz: @aleksuss Oops, you're right. Will fix and re-push.

vitvakatu:
great, looks like we're ready to merge!

PR #205, opened 2018-08-27
Rename crate to parity-rocksdb

andresilva:
This is necessary to allow publishing to crates.io. Merge only after https://github.com/paritytech/rust-snappy/pull/6 since we'll want to change our snappy dependency to the one we'll publish to crates.io.


andresilva:
Sorry I opened this PR against your repo by mistake.

PR #209, opened 2018-10-06
Adding LZ4, ZSTD, ZLIB, and BZIP2 compression support...

iSynaptic:
Cursory testing performed on both Linux and Windows...


ekmartin:
Am I misunderstanding something here, or would this cause _all_ compression methods to be compiled by default? That would be a pretty big increase to compile time.

Perhaps it would be useful to add compile flags similar to in #185.
I'm not sure which of the two approaches are preferable, but if building from source is, perhaps it would be better to depend on e.g. `lz4-sys` instead of a submodule?

lz4-rs: https://github.com/bozaro/lz4-rs
bzip2-rs: https://github.com/alexcrichton/bzip2-rs.git
zstd-rs: https://github.com/gyscos/zstd-rs

iSynaptic:
@ekmartin Thanks for taking a look at this PR.

Yes, this would by default cause all compression algorithms supported by the native RocksDB library to be compiled and included within `librocksdb-sys` (with the exception of the "Xpress" algorithm, which I can't find much info on).  I agree, it does increase compile time, but I think that is acceptable for a few reasons: 

1) It's not something that would be compiled frequently since it affects the `librocksdb-sys` crate. Once compiled, it should be a rare thing that triggers it to be compiled again. 

2) The Snappy compression algorithm, although the default, is not the recommended compression algorithm.  It is the default for backwards compatibility with databases created for older RocksDB versions.  Facebook [recommends](https://github.com/facebook/rocksdb/wiki/Compression) using LZ4 for standard level compression and ZSTD for bottommost.

3) Including these compression algorithms brings us to parity with other RocksDB language bindings (eg. RocksJava, gorocksdb, etc).  Snappy is a pretty terrible choice when LZ4 and ZSTD exists. Not including them by default when they are readily available in other language bindings may hinder adoption of this library (I know it did for me).

I'm not opposed to adding compiler/feature flags for each compression algorithm, but I would want them to be on by default and then allow them to be omitted by specifying an alternative list of desired features on a project-by-project basis.

As for using existing compression algorithm crates like `lz4-rs`, `bzip2-rs`, and `zstd-rs`, I don't think that is possible in this case.  The core RocksDB library from Facebook in the `librocksdb-sys/rocksdb` folder is written in C/C++ and it is what has support for these specific compression algorithms (see `librocksdb-sys/rocksdb/util/compression.h`). It uses them when it is reading/writing blocks in on-disk storage.  Finding a way for the core RocksDB library to make use of Rust crates would require a lot of rework of the internals of the library.  I don't think we gain anything by this (feel free to correct me) and it would make upgrading to future versions of the core RocksDB library very laborious.

I hope this addresses your concerns! Thanks again for taking a look at the PR.

iSynaptic:
> Thank you for the thorough reply!

Likewise!

> This is still a problem for initial compilations though, such as when you're building an image from scratch for a CI server or for deployments. Building RocksDB itself already takes a significant amount of time, so it would be nice to try to keep it from growing further.

I always like data when evaluating these kinds of tradeoffs.  I built the code both before (`0475bbe`) and after (`b56881e`) my changes and timed how much the new compression code added to the build time.  Before each timed build I ran `cargo clean`, `cargo build`, and `cargo clean -p librocksdb-sys` to ensure the next `cargo build` would only time compiling the core library.  The absolute timing isn't important - just the relative change.  On my workstation, the time it took to compile `librocksdb-sys` before the change was `2m 43s`, and after the change it took `2m 50s` for a delta of `7` seconds.  This isn't surprising since there isn't a lot of compression code relative to the rest of the RocksDB codebase.  Adding `4.2%` or `7` seconds to the build doesn't seem like a big deal to me.

As far as CI builds go, when you start facing long compile times, I've seen that addressed with tiered or layered builds - first a `cargo check`, followed by an `cargo build` with caches in tact, and then a `cargo clean && cargo build --release` on an hourly or nightly basis.  That way CI builds give you fast feedback and you have the benefit of a regular clean build.  Even with builds inside of Docker containers, this can be addressed many ways, including using mounted volumes to persist cache across builds and base Docker images.  Some CI servers that build within Docker containers even support this out of the box (eg. [Concourse](https://concourse-ci.org/tasks.html#task-caches) and [Drone](http://plugins.drone.io/drillster/drone-volume-cache)).

> I'd actually be in favor of releasing a new rust-rocksdb major version where the default is changed to LZ4, with snappy moved to a compilation flag. I don't think it's unreasonable to ask people that have existing snappy databases to enable a compilation flag when upgrading to a new major version.

This I disagree with.  Snappy is the default in the official RocksDB library and all the other language bindings I'm aware of.  The fact that Facebook didn't change the default from Snappy to LZ4 for backwards compatibility reasons, even though they prefer LZ4 over Snappy, tells me it probably isn't a safe move.  I can imagine opening an existing RocksDB, writing to it with the Rust bindings, having new blocks/SST files compressed with LZ4, and then rendering the DB inaccessible (fully or partially) when it's open with a different language binding because everything is based on the default options.  I'm not sure that is what would happen, but I think the safer choice is to leave the native defaults in tact.  Especially, when it is trivial to switch to LZ4 explicitly.

> I might be wrong here, but isn't it usually considered best practice to have conditional compilation turn on additional features — not the other way around? I'm not sure if I see the benefits of having them all on by default either, it's not like you'd want to use more than one compression algorithm at a time anyway, right?

I don't believe the best practice per is to disable all features by default, although I've never seen one explicitly stated.  Looking at the [Cargo.toml documentation](https://doc.rust-lang.org/cargo/reference/manifest.html#the-features-section) (and example TOML file), it would seem the convention is to provide sensible defaults that include the features that most projects would use, allow experimental or rarely used features to be turned on, and for scenarios where the common features are not needed for those to be turned off if desired.  For example, the reference TOML file linked to above includes this commentary:

```
[features]
# The default set of optional packages. Most people will want to use these
# packages, but they are strictly optional. Note that `session` is not a package
# but rather another feature listed in this manifest.
default = ["jquery", "uglifier", "session"]
```
Given that, and the fact that I believe these compression algorithms are frequently included in other bindings, I would suggest we default them as being turned on.  I'm curious to hear if others feel they need to be behind feature flags at all.

> I didn't mean using the bindings — I meant depending on the underlying -sys dependencies to simply build the original source and then e.g. statically linking to them in build.rs. 

I guess I still don't know the benefit of this, but I'm willing to be convinced otherwise if there is a strong beneficial reason.  I like using submodules because it's been working fine with Snappy so far and, more importantly, it doesn't require coordinating efforts with the other compression `-sys` crates when upgrading to newer versions of the compression code, or finding/creating another compression `-sys` crate if the core library add support for another one.

iSynaptic:
Unless someone objects, I will update this PR to included feature flags for each compression algorithm that are **enabled by default**.

iSynaptic:
@ekmartin I updated the PR to put compression algorithms behind crate features that are enabled by default. The README has been updated with basic instructions on how enable only specific algorithms, if desired.

iSynaptic:
@vitvakatu Can we get a review from you on this one?

iSynaptic:
Looks like this PR is approved to be merged, but the CI builds don't appear to be running. Suggest we merge once #214 is fixed, we get a good CI build, and an approval from @vitvakatu. Agreed?

iSynaptic:
@ozkriff You created #166 to link to RocksDB and Snappy via environment variables.  This PR adds support for more compression algorithms. Just wanted to give you a heads up. I think I followed the pattern you established, but if you want to review my work, that would be appreciated.

vitvakatu:
@iSynaptic sure, will review today

README.md
```diff
@@ -6,6 +6,15 @@ rust-rocksdb
 
 Feedback and pull requests welcome!  If a particular feature of RocksDB is important to you, please let me know by opening an issue, and I'll prioritize it.
 
-This binding is statically linked with a specific version of RocksDB. If you want to build it yourself, make sure you've also cloned the RocksDB and Snappy submodules:
+This binding is statically linked with a specific version of RocksDB. If you want to build it yourself, make sure you've also cloned the RocksDB and compression submodules:
 
     git submodule update --init --recursive
+
+## Compression Support
+By default, support for the [Snappy](https://github.com/google/snappy), [LZ4](https://github.com/lz4/lz4), [Zstd](https://github.com/facebook/zstd), [Zlib](https://zlib.net), and [Bzip2](http://www.bzip.org) compression is enabled through crate features.  If support for all of these compression algorithms is not needed, default features can be disabled and specific compression algorithms can be enabled. For example, to enable only LZ4 compression support, make these changes to your Cargo.toml:
```
vitvakatu: Well, this is not true, as far as I can see - default features includes `static` feature only.

vitvakatu: I also recommend you to use 80-chars long lines with explicit newlines

iSynaptic: > 
> 
> Well, this is not true, as far as I can see - default features includes `static` feature only.

The [default is set](https://github.com/rust-rocksdb/rust-rocksdb/pull/209/files#diff-80398c5faae3c069e4e6aa2ed11b28c0R17) at the main crate, not the `-sys` crate.

iSynaptic: > 
> 
> I also recommend you to use 80-chars long lines with explicit newlines

In the README, or somewhere else?

vitvakatu: Oh, my bad, I see now. 
> In the README, or somewhere else?

I prefer to see them everywhere, actually. The main reason is that it's much easier to review and modifying some words in the middle of the sentence usually doesn't produce huge diffs.

iSynaptic: Modified the README to wrap the lines.

PR #212, opened 2018-10-16
Implement Checkpoints

### LINKED ISSUE #35: Add support for checkpoints
It would be a very nice feature to have!

eupn:
# Description

This pull-request adds FFI for [Checkpoints](https://github.com/facebook/rocksdb/wiki/Checkpoints) feature that allows to create snapshots of RocksDB database at specified directory.

Closes #35.

## Changes made

Added `rocksdb::checkpoint` module that provides `Checkpoint` structure with implementation of FFI to RocksDB's [Checkpoints](https://github.com/facebook/rocksdb/wiki/Checkpoints) creation feature.

## Usage example

```rust
// Open existing RocksDB database
let db = DB::open_default("db".to_string()).unwrap();

// Create checkpoint object pointing to this DB
let checkpoint = Checkpoint::new(&db).unwrap();

// Create checkpoint no. 1 of DB
checkpoint.create_checkpoint("cp1".to_string()).unwrap();

// ... do some changes to DB ...

// Create checkpoint no. 2 of DB
checkpoint.create_checkpoint("cp2".to_string()).unwrap();

// ... and so forth ...
```

# How Has This Been Tested?

By executing unit tests that are located in `tests/test_checkpoint.rs`

### DIFF
--- a/src/checkpoint.rs
+++ b/src/checkpoint.rs
@@ -1,0 +1,80 @@
+// Copyright 2018 Eugene P.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+// http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+//
+
+///! Implementation of bindings to RocksDB Checkpoint[1] API
+///
+/// [1]: https://github.com/facebook/rocksdb/wiki/Checkpoints
+
+use {DB, Error};
+use ffi;
+use std::ffi::CString;
+use std::path::Path;
+
+/// Undocumented parameter for `ffi::rocksdb_checkpoint_create` function. Zero by default.
+const LOG_SIZE_FOR_FLUSH: u64 = 0_u64;
+
+/// Database's checkpoint object.
+/// Used to create checkpoints of the specified DB from time to time.
+pub struct Checkpoint {
+    inner: *mut ffi::rocksdb_checkpoint_t,
+}
+
+impl Checkpoint {
+    /// Creates new checkpoint object for specific DB.
+    ///
+    /// Does not actually produce checkpoints, call `.create_checkpoint()` method to produce
+    /// a DB checkpoint.
+    pub fn new(db: &DB) -> Result<Checkpoint, Error> {
+        let checkpoint: *mut ffi::rocksdb_checkpoint_t;
+
+        unsafe { checkpoint = ffi_try!(ffi::rocksdb_checkpoint_object_create(db.inner,)) };
+
+        if checkpoint.is_null() {
+            return Err(Error::new("Could not create checkpoint object.".to_owned()));
+        }
+
+        Ok(Checkpoint {
+            inner: checkpoint,
+        })
+    }
+
+    /// Creates new physical DB checkpoint in directory specified by `path`.
+    pub fn create_checkpoint<P: AsRef<Path>>(&self, path: P) -> Result<(), Error> {
+        let path = path.as_ref();
+        let cpath = match CString::new(path.to_string_lossy().as_bytes()) {
+            Ok(c) => c,
+            Err(_) => {
+                return Err(Error::new(
+                    "Failed to convert path to CString when creating DB checkpoint"
+                        .to_owned(),
+                ));
+            }
+        };
+
+        unsafe {
+            ffi_try!(ffi::rocksdb_checkpoint_create(self.inner, cpath.as_ptr(), LOG_SIZE_FOR_FLUSH,));
+
+            Ok(())
+        }
+    }
+}
+
+impl Drop for Checkpoint {
+    fn drop(&mut self) {
+        unsafe {
+            ffi::rocksdb_checkpoint_object_destroy(self.inner);
+        }
+    }
+}
--- a/tests/test_checkpoint.rs
+++ b/tests/test_checkpoint.rs
@@ -1,0 +1,113 @@
+// Copyright 2018 Eugene P.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+// http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+//
+extern crate rocksdb;
+
+use rocksdb::{checkpoint::Checkpoint, DB, Options};
+use std::fs::remove_dir_all;
+
+#[test]
+pub fn test_single_checkpoint() {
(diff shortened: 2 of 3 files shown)


src/checkpoint.rs
```diff
@@ -0,0 +1,77 @@
...
+            Ok(c) => c,
+            Err(_) => {
+                return Err(Error::new(
+                    "Failed to convert path to CString when creating DB checkpoint"
+                        .to_owned(),
+                ));
+            }
+        };
+
+        let log_size_for_flush = 0_u64;
```
aleksuss: @eupn I propose to create a constant for this value.

tests/test_checkpoint.rs
```diff
@@ -0,0 +1,141 @@
...
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+//
+extern crate rocksdb;
+
+use rocksdb::{checkpoint::Checkpoint, DB, Options};
+
+fn remove_dir_silently(path: &String) {
```
aleksuss: `&String` -> `&str` ?

tests/test_checkpoint.rs
```diff
@@ -0,0 +1,141 @@
...
+}
+
+#[test]
+pub fn test_multi_checkpoints() {
+    const PATH_PREFIX: &str = "_rust_rocksdb_cp_multi_";
+
+    // Create DB with some data and make a checkpoint
+    let cp1;
+    let cp1_path;
+    let db_path;
```
aleksuss: Why're these vars out of scope ?

tests/test_checkpoint.rs
```diff
@@ -0,0 +1,141 @@
...
+    {
+        db_path = PATH_PREFIX.to_string() + "db1";
```
aleksuss: I suppose `format!` marco is more suitable in this case.
`let db_path = format!("{}db1", PATH_PREFIX);`

tests/test_checkpoint.rs
```diff
@@ -0,0 +1,121 @@
...
+// limitations under the License.
+//
+extern crate rocksdb;
+
+use rocksdb::{checkpoint::Checkpoint, DB, Options};
+
+fn remove_dir_silently(path: &str) {
+    use std::fs::remove_dir_all;
+
+    match remove_dir_all(&path) {
```
vitvakatu: If you want to ignore result, I think `let _ = remove_dir_all(&path)` is more readable.

src/checkpoint.rs
```diff
@@ -0,0 +1,76 @@
...
+use {DB, Error};
+use ffi;
+use std::ffi::CString;
+use std::path::Path;
+
+const LOG_SIZE_FOR_FLUSH: u64 = 0_u64;
+
+/// Database's checkpoint object.
+/// Used to create checkpoints of the specified DB from time to time.
+pub struct Checkpoint {
```
vitvakatu: Don't you think that `CheckpointBuilder` would be a better name?

iSynaptic: I guess I like `Checkpoint` better. It matches the naming in the core RocksDB library.

src/checkpoint.rs
```diff
@@ -0,0 +1,77 @@
...
+/// Undocumented parameter for `ffi::rocksdb_checkpoint_create` function. Zero by default.
+const LOG_SIZE_FOR_FLUSH: u64 = 0_u64;
+
+/// Database's checkpoint object.
+/// Used to create checkpoints of the specified DB from time to time.
+pub struct CheckpointBuilder {
+    inner: *mut ffi::rocksdb_checkpoint_t,
+}
+
+impl CheckpointBuilder {
```
iSynaptic: I'd recommend this is changed back to `Checkpoint`. That is what the underlying object is named in the [core RocksDB library](https://github.com/facebook/rocksdb/blob/master/include/rocksdb/utilities/checkpoint.h), and is similarly named in the [Java](https://github.com/facebook/rocksdb/blob/master/java/src/main/java/org/rocksdb/Checkpoint.java) and [Go](https://github.com/tecbot/gorocksdb/blob/master/checkpoint.go) language bindings.

eupn: This was my initial thought on how to name it. 

@vitvakatu should I rename it back?

vitvakatu: I can agree with you, the consistency of code base among different bindings and original project is important, but don't you think that `Checkpoint` naming is misleading? I can't understand why `Checkpoint` does not have `create` method (perhaps static?), but has verbose `CreateCheckpoint`. One of the reason may be consistency with C API, but again, that looks weird to me.

eupn: To be more clear, I'd rename:
* `CheckpointBuilder` to `Checkpoint`
* `create_checkpoint` to `save_to`

So the workflow becomes:

```rust
let checkpoint = Checkpoint::new(); // Internal checkpoint_t object is created

let path = "cp1";
checkpoint.save_to(&cp1)?; // Physical directory produced
```

vitvakatu: @eupn I like your idea, yep. Of course, every review comment is my personal opinion and should be considered :wink:

iSynaptic: Thanks all - good discussion here.

I agree, the underlying API is not ideal.  I can't even discern why there even needs to be a `Checkpoint` class/object in the core RocksDB library - it seems like it would have been simpler to just put a `create_checkpoint` method right on the `DB` object.

That said, I'd like to make an argument to leave the function name as `create_checkpoint` instead of changing it to `save_to`.  Even though we all seem to agree that `save_to` is a much better name, it now creates a mismatch between this crate and underlying core library.  Each time we change the names of things it requires more work for the users of the crate to translate between the two.  The [official documentation](https://github.com/facebook/rocksdb/wiki/Checkpoints) on Checkpoints states that you first create the `Checkpoint` object, and then call the `CreateCheckpoint` method.  If we keep the name `create_checkpoint`, the official documentation works very well for this crate.  If we change the name, the user will likely have to do more work to verify that `save_to` is just `create_checkpoint` with a different name.

I recall reading somewhere - and just spent the last 30 minutes trying to find the reference, unsuccessfully - that when creating a Rust crate that is wrapping an underlying library, the suggested course of action is to keep the wrapper crate thin and as 1-to-1 as possible.  The idea being that developers who have used the underlying library in other languages will not experience an "impedance mismatch", so to speak, when using the wrapper crate.  It also noted that if there is a strong desire to create an "opinionated" Rust create that does drastically change the name and shape of things, it suggested that two crates be created - one that is thin and 1-to-1, and one that is opinionated.  I wish I could find that reference - if I do, I will post it.  To be clear, I'm not suggesting we create another crate.

Since the official documentation, the [C API](https://github.com/facebook/rocksdb/blob/0bd2ede10e97944798449e0f954658fac8bfdbd2/include/rocksdb/c.h#L200-L202), and the [Java](https://github.com/facebook/rocksdb/blob/master/java/src/main/java/org/rocksdb/Checkpoint.java#L49-L52) and [Go](https://github.com/tecbot/gorocksdb/blob/master/checkpoint.go#L36-L50) language bindings all use idiomatic versions of `create_checkpoint`, as much as I think we all agree it's not the greatest API, I suggest we keep the `create_checkpoint` function name.

eupn: @iSynaptic renamed it back

iSynaptic: @eupn @vitvakatu Thanks!

iSynaptic:
Looks like this PR is approved to be merged, but the CI builds don't appear to be running. Suggest we merge once #214 is fixed and we get a good CI build. Agreed?

aleksuss:
I've closed and reopened the PR to trigger CI build

PR #222, opened 2018-11-11
Made the DBIterator new functions public - To support passing ReadOptions to iterator

elichai:
Hi,
I want to use `DBIterator` with `set_iterate_upper_bound`, but no iterator function opens the possibility to give my own ReadOptions.

I think just making `DBIterator::new()` public is the simplest solution, although maybe it's better to add functions to DB that will support transforming it into an iterator and receiving Readoptions.

Thanks.

### DIFF
--- a/src/db.rs
+++ b/src/db.rs
@@ -433,7 +433,7 @@ impl Drop for DBRawIterator {
 }
 
 impl DBIterator {
-    fn new(db: &DB, readopts: &ReadOptions, mode: IteratorMode) -> DBIterator {
+    pub fn new(db: &DB, readopts: &ReadOptions, mode: IteratorMode) -> DBIterator {
         let mut rv = DBIterator {
             raw: DBRawIterator::new(db, readopts),
             direction: Direction::Forward, // blown away by set_mode()
@@ -443,7 +443,7 @@ impl DBIterator {
         rv
     }
 
-    fn new_cf(
+    pub fn new_cf(
         db: &DB,
         cf_handle: ColumnFamily,
         readopts: &ReadOptions,


elichai:
@iSynaptic Can I get some feedback on this? Thanks!

iSynaptic:
> @iSynaptic Can I get some feedback on this? Thanks!

Sure. As you noted, the historical approach has been to add functions to the `DB` impl to expose this kind of functionality. To keep things consistent with the current Rust library, it might be better to continue with that.

However, I wonder if this is the right long term approach.  The core library exposes a [`NewIterator`](https://github.com/facebook/rocksdb/wiki/Basic-Operations#iteration) function on the `DB` that takes in a `ReadOptions`. It might be worth shifting to that approach to keep things consistent with the core library.

elichai:
Whatever you decide :), 
I can implement it myself if you'd like (at least the first one, changing the whole approach might take too much time for me to invest)

Thanks!

iSynaptic:
@vitvakatu @aleksuss What are your thoughts?

vitvakatu:
Well, I think that we decided to follow parent library's design decisions as long as it's possible, didn't we? So I think we might better split responsibilities a bit and use some kind of `NewIterator`.
Although if it's a blocker @elichai, we might consider merging this PR and change API later.

elichai:
Thanks for the consideration.
Right now I'm just pointing to my github for it and it's fine cause my code is still not on crates.io

But if you'll make a design on how  the new iterator should be I can try and contribute however I can.

iSynaptic:
I'm going to make a PR to move iterator construction to the `DB` object and make sure it exposes a `&ReadOptions` argument. You can watch my progress on [this branch](https://github.com/iSynaptic/rust-rocksdb/tree/ergonomics-and-safety), along with some other ergonomic and safety updates.

PR #223, opened 2018-11-14
Makes ColumnFamily Send

Tpt:
ColumnFamily is a marker, I believe it is safe to tag it Send and Sync

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -229,3 +229,5 @@ pub struct WriteOptions {
 pub struct ColumnFamily {
     inner: *mut ffi::rocksdb_column_family_handle_t,
 }
+
+unsafe impl Send for ColumnFamily {}


Tpt:
The implementation of ColumnFamilyHandle inside of RocksDB seems safe (see ColumnFamilyHandleImpl type) but if you prefer I could drop `Sync`.

Tpt:
> Can you point to some documentation that indicates no external synchronization is needed?

`ColumnFamilyHandle` C++ type has some methods that seems to require external synchronization according to the source code comments but they are not accessible from the `ColumnFamily` type in this library. All the external accesses go through the DB object that is thread safe I believe. But I'm new to Rust and Rocksdb so I am maybe missing something

iSynaptic:
> `ColumnFamilyHandle` C++ type has some methods that seems to require external synchronization according to the source code comments but they are not accessible from the `ColumnFamily` type in this library.

Given that it's possible to add access to those methods in the future, it's a bit dangerous to implement `Sync` here.  You should be able to get to what you want wrapping a `Send`-able ColumnFamily with an `Arc<Mutex<T>>`.

Tpt:
Thank you!

PR #225, opened 2018-11-20
Bump bindgen up to 0.43

s-panferov:
I would like to build `rocksdb` with `mozjs`, but cannot
do this bacause of the version conflict.

Everything seems to work fine with the new version.


s-panferov:
@vitvakatu actually I don't, local override works for me for now. So maybe it's better to release this together with something else

PR #228, opened 2018-11-27
Renaming DBName to DBPath

iSynaptic:
This finalizes @ekmartin's changes from #183, but incorporated feedback from maintainers.


tests/util/mod.rs
```diff
@@ -0,0 +1,41 @@
+extern crate rocksdb;
+
+use std::time::{SystemTime, UNIX_EPOCH};
+use std::path::{PathBuf, Path};
+
+use rocksdb::{DB, Options};
+
+// Ensures that DB::Destroy is called for this database when DBPath is dropped.
```
aleksuss: I'd change a comment here to doc-comment and below as well.

PR #229, opened 2018-11-28
Fixing rustfmt.toml and applying formatting...

iSynaptic:


aleksuss:
I guess we need to add `cargo fmt` in .travis.yml as well. And one more thing. Do we really need `rustfmt.toml` file with custom settings? For example, we use `rustfmt` with default settings in our project.

iSynaptic:
> What rustfmt version is used?

@vitvakatu:
```
$ cargo fmt --version
rustfmt 0.99.4-stable (1c40881 2018-08-27)
```

@aleksuss How would that work? The Travis build would validate that `rustfmt` was run on all PRs? I deleted the `rustfmt.toml`, since the only setting remaining uses the default setting.

iSynaptic:
@aleksuss Cool. I suggest we wait until Rust 1.31.0 is released in a few days -- `rustfmt` is 1.0 at that point. Otherwise we will have to remove the `-preview` part of the component name. Added #230 to track.

@vitvakatu @aleksuss Besides the CI build changes, are we okay with this PR?

PR #238, opened 2018-12-16
Allow setting options on the DB

romanz:
This should allow changing options (e.g. `disable_auto_compactions`) without closing and re-opening the database.

### DIFF
--- a/src/db.rs
+++ b/src/db.rs
@@ -1123,6 +1123,35 @@ impl DB {
             );
         }
     }
+
+    pub fn set_options(&self, opts: &[(&str, &str)]) -> Result<(), Error> {
+        let copts = opts
+            .into_iter()
+            .map(|(name, value)| {
+                let cname = match CString::new(name.as_bytes()) {
+                    Ok(cname) => cname,
+                    Err(e) => return Err(Error::new(format!("Invalid option name `{}`", e))),
+                };
+                let cvalue = match CString::new(value.as_bytes()) {
+                    Ok(cvalue) => cvalue,
+                    Err(e) => return Err(Error::new(format!("Invalid option value: `{}`", e))),
+                };
+                Ok((cname, cvalue))
+            })
+            .collect::<Result<Vec<(CString, CString)>, Error>>()?;
+
+        let cnames: Vec<*const c_char> = copts.iter().map(|opt| opt.0.as_ptr()).collect();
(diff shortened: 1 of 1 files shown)


src/db.rs
```diff
@@ -1123,6 +1123,17 @@ impl DB {
             );
         }
     }
+
+    pub fn set_option(&self, name: &str, value: &str) -> Result<(), Error> {
+        let cname = CString::new(name.as_bytes()).unwrap();
```
aleksuss: To use `unwrap()` here it's a bad idea. 
```suggestion
         let cname = CString::new(name.as_bytes())?;
```

src/db.rs
```diff
@@ -1497,3 +1508,15 @@ fn snapshot_test() {
...
 }
+
+#[test]
+fn set_option_test() {
+    let path = "_rust_rocksdb_set_optionstest";
+    {
+        let db = DB::open_default(path).unwrap();
+        assert!(db.set_option("disable_auto_compactions", "true").is_ok());
+        assert!(db.set_option("disable_auto_compactions", "false").is_ok());
+        assert!(db.set_option("disable_auto_compactions", "INVALID_VALUE").is_err());
```
aleksuss: What about other options or this implementation related only for `disable_auto_compactions` option ?

src/db.rs
```diff
@@ -1123,6 +1123,17 @@ impl DB {
             );
         }
     }
+
+    pub fn set_option(&self, name: &str, value: &str) -> Result<(), Error> {
```
aleksuss: Have you thought about to make this method similar to original function? What if I want to change more than one parameter ? The implementation as I can see use synchronisation to apply this changes in runtime. So this is not a cheap operation.

src/db.rs
romanz: This should also work for other options as well (IIUC, the list is [here](https://github.com/facebook/rocksdb/blob/46e3209e0da1d7563e345d0e28a82ec05f1ddf32/options/options_helper.cc#L1663)). Currently in [electrs](https://github.com/romanz/electrs), I am planning using it for setting `disable_auto_compactions` to `false` after the initial bulk load is over.

romanz:
Thanks for the review :)
Answered the comments inline, and force-pushed 73aa6bf.

iSynaptic:
@aleksuss Was all of your feedback addressed?

romanz:
Thanks!

PR #167, opened 2018-02-20
Enable hardware CRC in rocksdb

ongardie-ebay:
This requires passing -msse4.2 and -DHAVE_SSE42 to the compiler.
You can check with:
  cargo test --release && grep CRC _rust*/LOG
Previously, this would output:
  Fast CRC32 supported: Not supported on x86
Now it shows:
  Fast CRC32 supported: Supported on x86

---

More generally, there may be other compiler flag differences between the Rust build and the upstream `make static_lib` build. I'm including below the full set of differences in my builds. I don't know what all of these are for, and I'm a bit worried we're missing out on other important ones. Thoughts?

```
Flags on Mac with Rust but not make:
-Irocksdb/
-Irocksdb/include/
-Irocksdb/third-party/gtest-1.7.0/fused-src/
-Isnappy/
-O3
-fPIC
-fdata-sections
-ffunction-sections
-m64
-msse4.2
c++

Flags on Mac with make but not Rust:
-DBZIP2
-DGFLAGS
-DHAVE_PCLMUL
-DLZ4
-DROCKSDB_BACKTRACE
-DROCKSDB_SUPPORT_THREAD_LOCAL
-DZLIB
-DZSTD
-I./include
-O2
-W
-Werror
-Wno-missing-field-initializers
-Wnon-virtual-dtor
-Woverloaded-virtual
-Wshadow
-Wshorten-64-to-32
-Wsign-compare
-fno-omit-frame-pointer
-fno-rtti
-g
-isystem=./third-party/gtest-1.7.0/fused-src
-march=native
-momit-leaf-frame-pointer
g++

Flags on Linux with Rust but not make:
-Irocksdb/
-Irocksdb/include/
-Irocksdb/third-party/gtest-1.7.0/fused-src/
-Isnappy/
-O3
-fPIC
-fdata-sections
-ffunction-sections
-m64
-msse4.2
c++

Flags on Linux with make but not Rust:
-DBZIP2
-DGFLAGS
-DHAVE_PCLMUL
-DLZ4
-DROCKSDB_BACKTRACE
-DROCKSDB_FALLOCATE_PRESENT
-DROCKSDB_MALLOC_USABLE_SIZE
-DROCKSDB_PTHREAD_ADAPTIVE_MUTEX
-DROCKSDB_RANGESYNC_PRESENT
-DROCKSDB_SCHED_GETCPU_PRESENT
-DROCKSDB_SUPPORT_THREAD_LOCAL
-DZLIB
-DZSTD
-I./include
[...]

PR #245, opened 2019-01-15
Disabled warnings when builds c++ deps

aleksuss:


iSynaptic:
Are these warnings suppressed in the upstream RocksDB build as well? I ask because I wouldn't want to hide something that needs our attention...

iSynaptic:
#167 notes many of the differences in compiler flags between our build scripts and the official RocksDB scripts. When I have the time, I'd like to eliminate the differences...

spacejam:
I'm in favor of muting noise that causes warning fatigue - if we are not able to act on information, it exists only as a cost, and we become numb to it anyway.

PR #246, opened 2019-01-27
Implement PinnableSlice based API

xxuejie:


ekmartin:
Are there any reasons why you wouldn't want to use `PinnableSlice` at all times? Since rust-rocksdb controls when things need to get cleaned up, wouldn't it be viable to simply use pinned slices by default? Something like: https://github.com/ekmartin/rust-rocksdb/commit/d350720106dadbaf50f08d15d955aa5c7115d87c

xxuejie:
@ekmartin Thanks for the tip! In our use case, we would be all good when rust-rocksdb always uses `PinnableSlice`. That said, I wonder if it might affect others when we alter existing APIs. Even though the interface stays the same, the underlying semantic(especially the performance implication) is changed. This is why we choose to add new APIs here in the PR.

That said, if you still feel that it's better we change `get` directly to use PinnableSlice, we are happy to make the change.

iSynaptic:
@xxuejie Thanks for putting this together.  I think `DBPinnableSlice` would also need a lifetime tying it's lifetime to the `DB`, something like this:

```
pub struct DBPinnableSlice<'a> {
    ptr: *mut ffi::rocksdb_pinnableslice_t,
    db: PhantomData<&'a DB>
}
```

Otherwise, the `DBPinnableSlice` can outlive the `DB`, at which point the `ptr` is invalid and using it would likely cause a segfault.  I'm currently working on a [branch ](https://github.com/iSynaptic/rust-rocksdb/tree/ergonomics-and-safety)to add lifetimes to `ColumnFamily`, `DBIterator` and `DBRawIterator` for the same reasons.

xxuejie:
Thanks @iSynaptic! I've taken a look at your branch and made modifications accordingly, let me know if you have any comments.

iSynaptic:
@xxuejie Awesome. Two other things I should have included in my earlier review:

1. Can you have `DBPinnableSlice` implement `AsRef<[u8]>`?
2. For all the `get_` methods, can you change the `key` type to be [passed in by a generic type](https://rust-lang-nursery.github.io/api-guidelines/flexibility.html#functions-minimize-assumptions-about-parameters-by-using-generics-c-generic)? eg:

```
pub fn get_pinned_opt<T: AsRef<[u8]>>(&self, key: T, readopts: &ReadOptions) {
  // some code ommitted
  unsafe {
    let key = key.as_ref();
    let val = ffi_try!(/* code that uses key */);
```
This will allow a more types to be passed in as a key. I'm currently making the same change in my ergonomics branch to update the existing APIs to follow the same pattern.  Nice thing is it shouldn't require any changes to your test.

xxuejie:
@iSynaptic Of course! Thanks for the tip! I've changed the code per your suggestion, please let me know if there's something we can do to make the code even better :)

src/db.rs
```diff
@@ -869,6 +870,76 @@ impl DB {
         self.get_cf_opt(cf, key, &ReadOptions::default())
     }
 
+    pub fn get_pinned_opt<T: AsRef<[u8]>>(&self, key: T, readopts: &ReadOptions) -> Result<Option<DBPinnableSlice>, Error> {
```
aleksuss: @xxuejie I'd prefer to see some description for public API

src/db.rs
```diff
@@ -1423,6 +1494,51 @@ fn to_cpath<P: AsRef<Path>>(path: P) -> Result<CString, Error> {
     }
 }
 
+pub struct DBPinnableSlice<'a> {
```
aleksuss: @xxuejie Add please some description what `DBPinnableSlice` is.

tests/test_pinnable_slice.rs
```diff
@@ -0,0 +1,24 @@
+extern crate rocksdb;
+
+use rocksdb::{Options, DB};
+
+#[test]
+pub fn test_pinnable_slice() {
```
aleksuss: `pub` is redundant I suppose ?

iSynaptic:
@xxuejie This is a minor nit: Would it be possible to change the generic type parameter from `T` to `K`? After making the recommendation, I landed #253 which adds the same `AsRef<[u8]>` changes to the rest of the API, and by convention I used `K` for keys and `V` for values.

I believe we can get this merged after this change and the suggestions from @aleksuss.

Thanks @xxuejie!

iSynaptic:
@xxuejie CI builds now have `rustfmt` checks. After you make your changes, you may need to run `cargo fmt --all` before this can be merged to master.

xxuejie:
@iSynaptic @aleksuss  Thanks for all your suggestions! And sorry for the delay since this is now holiday time for us. Anyway I will work on a patch which fixes the code according to your suggestions in the coming days.

iSynaptic:
@xxuejie No problem. Thanks for helping out!

xxuejie:
@iSynaptic @aleksuss This is now updated, let me know if you have more suggestions.

tests/test_pinnable_slice.rs
```diff
@@ -0,0 +1,24 @@
+extern crate rocksdb;
+
+use rocksdb::{Options, DB};
+
+#[test]
+fn test_pinnable_slice() {
+    let path = "_rust_rocksdb_pinnable_slice_test";
```
aleksuss: @xxuejie use `utils::DBPath` struct here please. It helps to destroy database and delete unnecessary folders with data for this test.

xxuejie: Thanks for the tip! This is resolved now

Issue #252, opened 2019-02-02
ColumnFamily structs can become invalid if column family is dropped.


iSynaptic:
Calling `drop_cf` on `DB` will cause any `ColumnFamily` struct copies pointing to the same column family to become invalid and may cause segfaults.

iSynaptic:
Apparently this is not an issue: https://github.com/facebook/rocksdb/wiki/Column-Families

> Drop the column family specified by ColumnFamilyHandle. Note that the actual data is not deleted until the client calls delete column_family;. You can still continue using the column family if you have outstanding ColumnFamilyHandle pointer.

Issue #235, opened 2018-12-07
Invalid pointers to underlying objects outlive DB


iSynaptic:
This is related to #234.

Pointers to underlying RocksDB objects (eg. iterators) can outlive the underlying `DB` object. If these objects are used after the `DB` is `Drop`-ped and closed, this can cause segfaults. Consider the implications to adding lifetime annotations to structs that manage these underlying objects.

aleksuss:
Very interesting. Could you please express an example of code that illustrates that bug ?

iSynaptic:
@aleksuss Yes, I will (assigned this issue to myself). #234 has one example, but I suspect there are others.  I'll try to find as many as I can.

iSynaptic:
Also, thanks to @rob-solana for finding this issue via #234 .

iSynaptic:
I have a branch that fixes this issue. Should be a PR soon.

iSynaptic:
Fixed by #253

PR #253, opened 2019-02-02
Ergonomics and safety improvements

iSynaptic:
- Adding lifetimes to `ColumnFamily`, `DBIterator` and `DBRawIterator` to prevent underlying pointer to outlive the `DB` struct - fixing #235, ~~mostly (see below)~~.
- Adding methods that construct `DBIterator` and `DBRawIterator` struct while accepting `&ReadOptions` - fixing #222
- Replaced all `key` and `value` parameters from `&[u8]` to generic types that implement `AsRef<[u8]>` - this will enable more argument types to be passed in and follows Rust API [design guidelines](https://rust-lang-nursery.github.io/api-guidelines/flexibility.html#functions-minimize-assumptions-about-parameters-by-using-generics-c-generic).

~~Adding a lifetime to `ColumnFamily` addresses most of the safety issues involving the underlying pointer. However, since `ColumnFamily` is `Copy + Clone`, if the column family is dropped with the `drop_cf` function on `DB`, any copied `ColumnFamily` referring to the same column family now has an invalid underlying pointer. A future PR should address this issue (#252).~~

**Update**: Apparently, [it's okay](https://github.com/facebook/rocksdb/wiki/Column-Families) to have copies of underlying column family pointer after a call to `drop_cf`:

> Drop the column family specified by `ColumnFamilyHandle`. Note that the actual data is not deleted until the client calls `delete column_family;`. You can still continue using the column family if you have outstanding ColumnFamilyHandle pointer.


src/db.rs
```diff
@@ -952,18 +1004,18 @@ impl DB {
         DBIterator::new_cf(self, cf_handle, &opts, mode)
     }
 
-    pub fn prefix_iterator_cf(
+    pub fn prefix_iterator_cf<P: AsRef<[u8]>>(
         &self,
         cf_handle: ColumnFamily,
-        prefix: &[u8]
+        prefix: P
```
```suggestion
        prefix: P
        prefix: impl AsRef<u8>
```

src/db.rs
```diff
@@ -952,18 +1004,18 @@ impl DB {
         DBIterator::new_cf(self, cf_handle, &opts, mode)
     }
 
-    pub fn prefix_iterator_cf(
+    pub fn prefix_iterator_cf<P: AsRef<[u8]>>(
```
```suggestion
    pub fn prefix_iterator_cf(
```

src/db.rs
aleksuss: But it's a matter of taste.

src/db.rs
```diff
@@ -1035,14 +1107,21 @@ impl DB {
...
+        key: K,
+        value: V,
         writeopts: &WriteOptions,
-    ) -> Result<(), Error> {
+    ) -> Result<(), Error>
+        where K: AsRef<[u8]>, 
+              V: AsRef<[u8]> {
+        
+        let key = key.as_ref();
+        let value = value.as_ref();
```
aleksuss: Is it necessary ? What if just write below:
```rust
key.as_ref().as_ptr() as *const c_char,
``` 
???

aleksuss: The point is less lines of code is better 🙂

iSynaptic: Only because I'd have to do it twice on lines 1129 and 1130. Either way works I suppose...

src/db.rs
iSynaptic: I avoid it because it doesn't allow you to [explicitly specify type arguments](https://github.com/rust-lang/rust/issues/50950).

tests/util/mod.rs
```diff
@@ -39,4 +39,4 @@ impl AsRef<Path> for DBPath {
     fn as_ref(&self) -> &Path {
         &self.path
     }
-}
+}
```
aleksuss: New line ? ☺️

tests/test_db.rs
```diff
@@ -0,0 +1,183 @@
...
+        // empty options are not allowed
+        assert!(db.set_options(&[]).is_err());
+        // multiple options can be set in a single API call
+        let multiple_options = [
+            ("paranoid_file_checks", "true"),
+            ("report_bg_io_stats", "true"),
+        ];
+        db.set_options(&multiple_options).unwrap();
+    }
+}
```
aleksuss: New line ?

tests/test_compationfilter.rs
```diff
@@ -0,0 +1,51 @@
...
+        let db = DB::open(&opts, &path).unwrap();
+        let _ = db.put(b"k1", b"a");
+        let _ = db.put(b"_k", b"b");
+        let _ = db.put(b"%k", b"c");
+        db.compact_range(None::<&[u8]>, None::<&[u8]>);
+        assert_eq!(&*db.get(b"k1").unwrap().unwrap(), b"a");
+        assert!(db.get(b"_k").unwrap().is_none());
+        assert_eq!(&*db.get(b"%k").unwrap().unwrap(), b"secret");
+    }
+}
```
aleksuss: `\n` ?

tests/util/mod.rs
iSynaptic: Fixed.

tests/test_db.rs
iSynaptic: Fixed.

tests/test_compationfilter.rs
iSynaptic: Fixed.

tests/test_compationfilter.rs
```diff
@@ -0,0 +1,51 @@
+// Copyright 2014 Tyler Neely
```
aleksuss: Sorry for my pickyness, but now is 2019 year 😊

tests/test_db.rs
```diff
@@ -0,0 +1,183 @@
+// Copyright 2014 Tyler Neely
```
aleksuss: 2019 ?

tests/test_compationfilter.rs
iSynaptic: No need to apologize. Thanks for catching it.  Fixed.

tests/test_db.rs
iSynaptic: Fixed.

aleksuss:
Might be I skipped something but I couldn't find a test for issue #235 ?

tests/test_db.rs
```diff
@@ -0,0 +1,183 @@
...
+        let iter = db.iterator(IteratorMode::Start);
+        for (k, v) in iter {
+            println!(
+                "Hello {}: {}",
+                str::from_utf8(&*k).unwrap(),
+                str::from_utf8(&*v).unwrap()
+            );
+        }
+    }
+}
```
aleksuss: I'd rewrite this test for something like this:
```rust
#[test]
fn iterator_test() {
    let path = DBPath::new("_rust_rocksdb_iteratortest");
    {
        let data = [(b"k1", b"v1111"), (b"k2", b"v2222"), (b"k3", b"v3333")];
        let db = DB::open_default(&path).unwrap();

        for (key, value) in &data {
            assert!(db.put(key, value).is_ok());
        }

        let iter = db.iterator(IteratorMode::Start);

        for (idx, (db_key, db_value)) in iter.enumerate() {
            let (key, value) = data[idx];
            assert_eq!((&key[..], &value[..]), (db_key.as_ref(), db_value.as_ref()));
        }
    }
}
```

tests/test_db.rs
```diff
@@ -0,0 +1,183 @@
...
+fn external() {
+  let path = DBPath::new("_rust_rocksdb_externaltest");
+  let db = DB::open_default(&path).unwrap();
+  let p = db.put(b"k1", b"v1111");
+  assert!(p.is_ok());
+  let r: Result<Option<DBVector>, Error> = db.get(b"k1");
+  assert!(r.unwrap().unwrap().to_utf8().unwrap() == "v1111");
+  assert!(db.delete(b"k1").is_ok());
+  assert!(db.get(b"k1").unwrap().is_none());
+}
```
aleksuss: I suppose there is something wrong with formatting ?

iSynaptic: This was just code that was moved from `db.rs` into this file so `DBPath` could be used.  If the formatting is wrong, I suppose it was prior to me moving it around.  That said, I'll try to fix it up...

tests/test_db.rs
iSynaptic: This test was moved from `db.rs` to make use of `DBPath`, but I'll make the changes just the same...

tests/test_db.rs
```diff
@@ -0,0 +1,183 @@
...
+
+#[test]
+fn snapshot_test() {
+    let path = DBPath::new("_rust_rocksdb_snapshottest");
+    {
+        let db = DB::open_default(&path).unwrap();
+        let p = db.put(b"k1", b"v1111");
+        assert!(p.is_ok());
+
```
aleksuss: Do we really need `let p ...` here ? 
```suggestion
        assert!(db.put(b"k1", b"v1111").is_ok());
```

tests/test_db.rs
```diff
@@ -0,0 +1,183 @@
...
+    }
+}
+
+#[test]
+fn snapshot_test() {
+    let path = DBPath::new("_rust_rocksdb_snapshottest");
+    {
+        let db = DB::open_default(&path).unwrap();
+        let p = db.put(b"k1", b"v1111");
+        assert!(p.is_ok());
```
aleksuss:
```suggestion
       assert!(db.put(b"k1", b"v1111").is_ok());
```

aleksuss: Do we really need let p ... here ?

iSynaptic:
> 
> 
> @iSynaptic Might be I skipped something but I couldn't find a test for issue #235 ?

@aleksuss I don't know that I can write a test for this, since the lifetimes prevent code from compiling that demonstrate this issue. If you have any ideas on how to write a test that proves `ColumnFamily`, `DBIterator` and `DBRawIterator` can't outlive `DB`, please let me know. I'd love to have that test.

tests/test_db.rs
iSynaptic: No. Fixed.

tests/test_db.rs
iSynaptic: Fixed.

tests/test_db.rs
iSynaptic: Changed.

aleksuss:
@iSynaptic Yeah, I guess you're right. I've checked it out and found out a compile time error if an  iterator outlive a db instance. It's ok. 👍🏻

iSynaptic:
> 
> 
> Almost LGTM [cf4679d](https://github.com/rust-rocksdb/rust-rocksdb/commit/cf4679d8e8347ef54fa65a7594ba5194a88d34ff)

Almost?

tests/test_db.rs
```diff
@@ -0,0 +1,185 @@
...
+}
+
+#[test]
+fn external() {
+    let path = DBPath::new("_rust_rocksdb_externaltest");
+
+    {
+        let db = DB::open_default(&path).unwrap();
+        
+        let p = db.put(b"k1", b"v1111");
```
aleksuss: I'd remove this binging as well.

tests/test_db.rs
```diff
@@ -0,0 +1,185 @@
...
+#[test]
+fn snapshot_test() {
+    let path = DBPath::new("_rust_rocksdb_snapshottest");
+    {
+        let db = DB::open_default(&path).unwrap();
+        
+        assert!(db.put(b"k1", b"v1111").is_ok());
+
+        let snap = db.snapshot();
+        let r: Result<Option<DBVector>, Error> = snap.get(b"k1");
```
aleksuss: And here

tests/test_db.rs
```diff
@@ -0,0 +1,185 @@
...
+            let r: Result<Option<DBVector>, Error> = db.get(b"k1");
+            assert!(r.unwrap().unwrap().to_utf8().unwrap() == "v1111");
+        }
+        {
+            // test delete
+            let mut batch = WriteBatch::default();
+            let _ = batch.delete(b"k1");
+            assert_eq!(batch.len(), 1);
+            assert!(!batch.is_empty());
+            let p = db.write(batch);
```
aleksuss: And here

tests/test_db.rs
```diff
@@ -0,0 +1,185 @@
...
+            // test put
+            let mut batch = WriteBatch::default();
+            assert!(db.get(b"k1").unwrap().is_none());
+            assert_eq!(batch.len(), 0);
+            assert!(batch.is_empty());
+            let _ = batch.put(b"k1", b"v1111");
+            assert_eq!(batch.len(), 1);
+            assert!(!batch.is_empty());
+            assert!(db.get(b"k1").unwrap().is_none());
+            let p = db.write(batch);
```
aleksuss: And here

tests/test_db.rs
aleksuss: But I understand that this code isn't yours.

iSynaptic: Still, good idea to make it better.

iSynaptic: Fixed.

tests/test_db.rs
iSynaptic: Fixed.

tests/test_db.rs
iSynaptic: Fixed.

elichai:
Looks great :)
Eager for a release on crates.io so I could stop using my fork

elichai:
@iSynaptic So somehow I missed the fact that `iterator_cf_opt` is only implemented on `Snapshot` and not on `DB` :/

PR #254, opened 2019-02-03
Adding test to validate correct behavior of prefix_iterator...

iSynaptic:


tests/test_iterator.rs
```diff
@@ -207,6 +207,50 @@ pub fn test_prefix_iterator() {
     }
 }
 
+#[test]
+pub fn test_prefix_iterator_uses_full_prefix() {
```
aleksuss: Is `pub` redundant ?

PR #255, opened 2019-02-04
Using generic types to specify column families when opening database...

### LINKED ISSUE #175: DB::open_cf signature

### LINKED ISSUE #177: Database opening ergonomics
Currently, if you use column families you must write something like that:
```rust
fn open_db<P: AsRef<Path>>(path: P) -> Result<DB, Error> {
    let options = Options::default();
    let column_families = DB::list_cf(&options, path)?;
    let cfs = column_families.iter().map(|name| name.as_str()).collect::<Vec<_>>();
    DB::open_cf(&options, cfs.as_ref(), path)
}
```
I suppose this logic (obtaining the list of column families and passing it to the `open_cf`) can be hidden inside of an additional `DB` constructor.

iSynaptic:
This change requires no changes to existing tests. Fixes #175 and fixes #177.

### DIFF
--- a/src/db.rs
+++ b/src/db.rs
@@ -679,7 +679,7 @@ impl DB {
         DB::open_cf_descriptors(opts, path, cfs)
     }
 
-    /// Open a database with the given database options and column family names/options.
+    /// Open a database with the given database options and column family descriptors.
     pub fn open_cf_descriptors<P, I>(
         opts: &Options,
         path: P,
--- a/src/db.rs
+++ b/src/db.rs
@@ -22,7 +22,6 @@ use std::collections::BTreeMap;
 use std::ffi::{CStr, CString};
 use std::fmt;
 use std::fs;
-use std::iter;
 use std::marker::PhantomData;
 use std::ops::Deref;
 use std::path::Path;
@@ -661,7 +660,7 @@ impl DB {
 
     /// Open the database with the specified options.
     pub fn open<P: AsRef<Path>>(opts: &Options, path: P) -> Result<DB, Error> {
-        DB::open_cf(opts, path, iter::empty::<&str>())
+        DB::open_cf(opts, path, None::<&str>)
     }
 
     /// Open a database with the given database options and column family names.
--- a/src/db.rs
+++ b/src/db.rs
@@ -22,6 +22,7 @@ use std::collections::BTreeMap;
 use std::ffi::{CStr, CString};
 use std::fmt;
 use std::fs;
+use std::iter;
 use std::marker::PhantomData;
 use std::ops::Deref;
 use std::path::Path;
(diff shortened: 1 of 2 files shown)


iSynaptic:
@DarkEld3r I hope this change represents a happy medium between the way `open_cf` has worked in the past and your suggestion in #177. Rather than creating a new constructor, the existing ones should be easier to use. For example:

https://github.com/rust-rocksdb/rust-rocksdb/blob/d00e03f72a5f9fe34abefba102e6dd61fb76a69a/tests/test_column_family.rs#L112-L114

src/db.rs
```diff
@@ -660,28 +661,37 @@ impl DB {
...
+
+        let cfs = cfs
+            .into_iter()
+            .map(|name| ColumnFamilyDescriptor::new(name.as_ref(), Options::default()));
 
-        DB::open_cf_descriptors(opts, path, cfs_v)
+        DB::open_cf_descriptors(opts, path, cfs)
     }
 
     /// Open a database with the given database options and column family names/options.
```
```suggestion
    /// Open a database with the given database options and column family descriptors.
```

tests/test_column_family.rs
```diff
@@ -88,6 +88,35 @@ pub fn test_column_family() {
...
+
+    let n = DBPath::new("_rust_rocksdb_cftest_with_list_cf");
+
+    {
+        let mut opts = Options::default();
+        opts.create_if_missing(true);
+        let db = DB::open(&opts, &n).unwrap();
+        let opts = Options::default();
+
+        match db.create_cf("cf1", &opts) {
```
aleksuss: I'd replace this `match` with only one line of code:
```rust
assert!(db.create("cf1", &opts).is_some());
```

src/db.rs
```diff
@@ -660,28 +661,37 @@ impl DB {
...
     /// Open a database with the given database options and column family names.
     ///
     /// Column families opened using this function will be created with default `Options`.
-    pub fn open_cf<P: AsRef<Path>>(opts: &Options, path: P, cfs: &[&str]) -> Result<DB, Error> {
-        let cfs_v = cfs
-            .to_vec()
-            .iter()
-            .map(|name| ColumnFamilyDescriptor::new(*name, Options::default()))
-            .collect();
+    pub fn open_cf<P, I, N>(opts: &Options, path: P, cfs: I) -> Result<DB, Error>
```
aleksuss: By the way, what do you think about such signature: 
```rust
 pub fn open_cf<P, I, N>(opts: &Options, path: P, cfs: Option<I>) -> Result<DB, Error>;
```
???

iSynaptic: Since `Option<T>` [already implements](https://doc.rust-lang.org/std/option/enum.Option.html#impl-IntoIterator-2) `IntoIterator`, you should be able to pass an `Option<I>` to the function as it is. Is there some other reason to change it?

aleksuss: Oh. Than I'd prefer to see on line 664 something like this:
```rust
 DB::open_cf(opts, path, None::<&str>())
```

iSynaptic: Not at my workstation, but it would probably be something like:

`DB::open_cf(opts, path, None::<&[&str]>())`

iSynaptic: Done.

PR #263, opened 2019-02-13
Adding set_readahead_size setter to ReadOptions...

### LINKED ISSUE #203: Missing `ReadOptions.readahead_size`
I like to "tune" my RocksDB instances by changing random settings that end up having no impact on performance whatsoever. `ReadOptions.readahead_size` will be the one though so it would be great to have.

iSynaptic:
Fixes #203.

### DIFF
--- a/src/db.rs
+++ b/src/db.rs
@@ -1721,6 +1721,18 @@ impl ReadOptions {
     pub fn set_total_order_seek(&mut self, v: bool) {
         unsafe { ffi::rocksdb_readoptions_set_total_order_seek(self.inner, v as c_uchar) }
     }
+
+    /// ```
+    /// use rocksdb::{ReadOptions};
+    ///
+    /// let mut opts = ReadOptions::default();
+    /// opts.set_readahead_size(4_194_304); // 4mb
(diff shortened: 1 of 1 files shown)


src/db.rs
```diff
@@ -1721,6 +1721,18 @@ impl ReadOptions {
...
         unsafe { ffi::rocksdb_readoptions_set_total_order_seek(self.inner, v as c_uchar) }
     }
+
+    /// ```
+    /// use rocksdb::{ReadOptions};
+    ///
+    /// let mut opts = ReadOptions::default();
+    /// opts.set_readahead_size(4_194_304); // 4mb
+    /// ```
+    pub fn set_readahead_size(&mut self, v: usize) {
```
aleksuss: I'd add some description for this option. Just copy past from `rocksdb/include/options.h`
```
// If non-zero, NewIterator will create a new table reader which
  // performs reads of the given size. Using a large size (> 2MB) can
  // improve the performance of forward iteration on spinning disks.
  // Default: 0
```

iSynaptic: Done.

PR #266, opened 2019-02-15
Applied clippy lints

aleksuss:


iSynaptic:
@aleksuss Is there an opportunity to make Clippy part of the CI build?

aleksuss:
@iSynaptic I don't think so because format of `bindgen.rs` file is unpredictable.

iSynaptic:
@aleksuss Could `bindgen.rs` be ignored by Clippy?

aleksuss:
@iSynaptic I don't know right now but I will think about this. Actually It's a really good idea to add clippy's check in CI. 👍🏻

Issue #135, opened 2017-08-27
Expose API to restore database from backups


benoitc:
the backup module can backup and has restore options, but how do we backup?

iSynaptic:
Can you provide more detail? What are you trying to do?

rohitjoshi:
How do I restore from an existing backup directory?  I don't see `pub unsafe extern "C" fn rocksdb_backup_engine_restore_db_from_latest_backup()` is made available in the Rust interface.

iSynaptic:
I can see that the library doesn't expose the ability to restore an backup. Looks like the work has been started but not finished.  I've labeled this as a bug. Anyone is free to take on fixing this, otherwise I will plan on it once other in-flight work is completed.

rohitjoshi:
I have implemented this on my local branch. I will submit a PR.

iSynaptic:
Looks like this will be addressed in #249 by @rohitjoshi.

rohitjoshi:
@iSynaptic I am close #249 and splitting into two different PRs.
Submitted #272 for this issue.

iSynaptic:
Fixed by #272.

PR #272, opened 2019-02-25
Support for restore from the backup

rohitjoshi:
This PR implements restoring DB from the latest backup.  
Resolves issue# https://github.com/rust-rocksdb/rust-rocksdb/issues/135

### DIFF
--- a/src/backup.rs
+++ b/src/backup.rs
@@ -78,6 +78,60 @@ impl BackupEngine {
             Ok(())
         }
     }
+    /// Restore from the latest backup
+    /// db_dir:  input db directory
+    /// wal_dir:  input wal directory
+    /// opts: restore options
+    /// example:
+    /// use rocksdb::backup::{BackupEngine, BackupEngineOptions};
+    /// let backup_opts = BackupEngineOptions::default();
+    /// let backup_engine = BackupEngine::open(&backup_opts, &backup_path).unwrap();
+    /// let mut restore_option = rocksdb::backup::RestoreOptions::default();
+    /// restore_option.set_keep_log_files(true); /// true to keep log files
+    /// if let Err(e) = backup_engine.restore_from_latest_backup(&db_path, &wal_dir, &restore_option) {
+    ///     error!("Failed to restore from the backup. Error:{:?}", e);
+    ///     return Err(e.to_string());
+    ///  }
+    pub fn restore_from_latest_backup<P: AsRef<Path>>(
+        &mut self,
+        db_dir: P,
+        wal_dir: P,
+        opts: &RestoreOptions,
+    ) -> Result<(), Error> {
+        let db_dir = db_dir.as_ref();
+        let c_db_dir = match CString::new(db_dir.to_string_lossy().as_bytes()) {
+            Ok(c) => c,
+            Err(_) => {
+                return Err(Error::new(
+                    "Failed to convert db_dir to CString \
+                     when restoring from latest backup"
+                        .to_owned(),
+                ));
+            }
+        };
+
+        let wal_dir = wal_dir.as_ref();
+        let c_wal_dir = match CString::new(wal_dir.to_string_lossy().as_bytes()) {
+            Ok(c) => c,
+            Err(_) => {
+                return Err(Error::new(
+                    "Failed to convert wal_dir to CString \
+                     when restoring from latest backup"
+                        .to_owned(),
+                ));
+            }
+        };
+
+        unsafe {
+            ffi_try!(ffi::rocksdb_backup_engine_restore_db_from_latest_backup(
+                self.inner,
+                c_db_dir.as_ptr(),
+                c_wal_dir.as_ptr(),
+                opts.inner,
+            ));
+        }
+        Ok(())
+    }
 }
 
 impl BackupEngineOptions {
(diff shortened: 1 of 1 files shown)


src/backup.rs
```diff
@@ -78,6 +78,60 @@ impl BackupEngine {
...
+    /// use rocksdb::backup::{BackupEngine, BackupEngineOptions};
+    /// let backup_opts = BackupEngineOptions::default();
+    /// let backup_engine = BackupEngine::open(&backup_opts, &backup_path).unwrap();
+    /// let mut restore_option = rocksdb::backup::RestoreOptions::default();
+    /// restore_option.set_keep_log_files(true); /// true to keep log files
+    /// if let Err(e) = backup_engine.restore_from_latest_backup(&db_path, &wal_dir, &restore_option) {
+    ///     error!("Failed to restore from the backup. Error:{:?}", e);
+    ///     return Err(e.to_string());
+    ///  }
+    pub fn restore_from_latest_backup<P: AsRef<Path>>(
```
iSynaptic: Can you have two generic type arguments here, eg: `D: AsRef<Path>, W: AsRef<Path>`? That way each path can be provided by different types.

src/backup.rs
```diff
@@ -78,6 +78,60 @@ impl BackupEngine {
             Ok(())
         }
     }
+    /// Restore from the latest backup
```
vitvakatu: Please change the formatting of this doc-comment.
Markdown will not follow your line feeds, so all these lines would be probably just one paragraph in the docs.
Moreover, you need to surround your example with `\`\`\`ignore` and `\`\`\`` lines.  With `ignore`, your example will _not_ be considered doc-test, so compiler will ignore it. This is not good in terms of actuality of documentation, so I suggest you to try make your doc-example compilable.
See https://doc.rust-lang.org/rustdoc/documentation-tests.html for more information. Also see https://doc.rust-lang.org/rust-by-example/meta/doc.html as an example of good doc-comment.

Thanks

src/backup.rs
```diff
@@ -78,6 +78,60 @@ impl BackupEngine {
...
+            ffi_try!(ffi::rocksdb_backup_engine_restore_db_from_latest_backup(
+                self.inner,
+                c_db_dir.as_ptr(),
+                c_wal_dir.as_ptr(),
+                opts.inner,
+            ));
+        }
+        Ok(())
+    }
 }
```
aleksuss: @rohitjoshi I'd like to see a test for this functionality. What do you think ?

rohitjoshi: Let me see how can I add the test cases but I have used and validated at https://github.com/rohitjoshi/bhatho/blob/master/src/db/rocks_db.rs

src/backup.rs
rohitjoshi: Thanks.  I need to spend more time understanding rust documentation. Will take care of this.

rohitjoshi: @vitvakatu thanks for `cargo doc --open`. I helped me to see the doc output.  I have resolved the doc issue now.

src/backup.rs
aleksuss: Test could be like this I suppose:
1. Create database. 
2. Put into it one key, value. 
3. Create backup. 
4. Put one more key, value.
5. Close database. 
6. Restore database.
7. Inserted key, value on step 4 shouldn't exist in restored database. 

If I understand correctly purpose of the functionality.

rohitjoshi: I will give a try this weekend.

src/backup.rs
```diff
@@ -139,3 +202,43 @@ impl Drop for RestoreOptions {
         }
     }
 }
+
+#[test]
+fn backup_restore() {
+    use db::DBVector;
+    // create backup
+    let path = "_rust_rocksdb_backup_restore_test";
```
aleksuss: @rohitjoshi you can use `util::DBPath` struct here

rohitjoshi: I followed existing test cases patterns from `db.rs`.  `util::DBPath` is under `tests` directory.

aleksuss: It will be cool if you'd invoke `DB::Destroy` to remove tmp directories `_rust_rocksdb_backup_restore_test`, `_rust_rocksdb_backup_path` and `_rust_rocksdb_restore_from_backup_path` after test complete.

iSynaptic: @rohitjoshi I think this is the last change we need. Invoking `DB::Destroy` will make sure the test database will be cleaned up at the end of the test run.  As you noted, `util::DBPath` is only available in the `tests` crate.  I'm working on a PR that, among other things, moves `DBPath` into the main crate and calls it `TemporaryDBPath`.  Once we merge your PR into `master`, I can incorporate it into my PR and updated it to use `TemporaryDBPath`. But for now, `DB::Destroy` will suffice to keep what's on `master` from leaving temporary databases around.

Thanks for all your work on this!

anlihust:
When this PR  will  be merged？

iSynaptic:
> 
> 
> When this PR will be merged？
@anlihust We are waiting for one small change to be pushed from @rohitjoshi and then I believe we are ready for it to be merged.

rohitjoshi:
@iSynaptic @anlihust Added DB:destroy(). Please review.

PR #277, opened 2019-03-07
Implementing AsRef<[u8]> for DBVector...

iSynaptic:
I'd like to get this in before we do a 0.12.0 release...


iSynaptic:
@aleksuss I added a test.

PR #280, opened 2019-03-17
Remove allocations from SliceTransform

ekmartin:
This adopts [pingcap/rust-rocksdb](https://github.com/pingcap/rust-rocksdb)'s SliceTransform [approach](https://github.com/pingcap/rust-rocksdb/blob/master/src/slice_transform.rs), which changes the signature of `transform_fn` from `&[u8] -> Vec<u8>` to `&'a [u8] -> &'a [u8]` while removing the manual `malloc` call we're currently doing in `transform_callback`.

This significantly reduces memory usage, to the point where I wonder if this might actually be a memory leak if rocksdb isn't actually freeing the prefixes that we allocate in `transform_callback`. I might be wrong about this though—see the attached heap dump below.

### Breaking change
The main use case for slice transforms is to create prefixes from existing keys, which you don't really need allocations for. However, in theory, the existing API could be used to create prefixes from completely different keys than what's passed in:

```rust
fn transform(k: &[u8]) -> Vec<u8> {
    "aa".as_bytes().to_vec()
}
```

This wouldn't be possible anymore with the API proposed in this PR.

### Alternative API
PingCAP implements slice transforms [as a trait](https://github.com/pingcap/rust-rocksdb/blob/master/src/slice_transform.rs#L19-L38) where the transform function has a mutable receiver. This is a more flexible API in my opinion, but it would make the upgrade lift even more significant than what this PR already proposes. What do you think?

### Heap dump
Test program: https://gist.github.com/ekmartin/e96275f09874a13a578faf0fbd2d0f6d
Mastiff dump with master: https://gist.github.com/ekmartin/41f3c4eb56ccfb3caa26191605d8097b
Mastiff dump with this branch: https://gist.github.com/ekmartin/e0f47e44cb9d0142e09a87f8de8e949b

### DIFF
--- a/src/slice_transform.rs
+++ b/src/slice_transform.rs
@@ -13,11 +13,9 @@
 // limitations under the License.
 
 use std::ffi::CString;
-use std::mem;
-use std::ptr;
 use std::slice;
 
-use libc::{self, c_char, c_void, size_t};
+use libc::{c_char, c_void, size_t};
 
 use ffi;
 
@@ -41,15 +39,15 @@ impl SliceTransform {
         transform_fn: TransformFn,
         in_domain_fn: Option<InDomainFn>,
     ) -> SliceTransform {
-        let cb = Box::new(TransformCallback {
+        let cb = Box::into_raw(Box::new(TransformCallback {
             name: CString::new(name.as_bytes()).unwrap(),
             transform_fn,
             in_domain_fn,
-        });
+        }));
 
         let st = unsafe {
             ffi::rocksdb_slicetransform_create(
-                mem::transmute(cb),
+                cb as *mut c_void,
                 Some(slice_transform_destructor_callback),
                 Some(transform_callback),
                 // this is ugly, but I can't get the compiler
@@ -82,18 +80,17 @@ impl SliceTransform {
     }
 }
 
-pub type TransformFn = fn(&[u8]) -> Vec<u8>;
+pub type TransformFn<'a> = fn(&'a [u8]) -> &'a [u8];
 pub type InDomainFn = fn(&[u8]) -> bool;
 
-pub struct TransformCallback {
+pub struct TransformCallback<'a> {
     pub name: CString,
-    pub transform_fn: TransformFn,
+    pub transform_fn: TransformFn<'a>,
     pub in_domain_fn: Option<InDomainFn>,
 }
 
 pub unsafe extern "C" fn slice_transform_destructor_callback(raw_cb: *mut c_void) {
-    let transform: Box<TransformCallback> = mem::transmute(raw_cb);
-    drop(transform);
+    Box::from_raw(raw_cb as *mut TransformCallback);
 }
 
 pub unsafe extern "C" fn slice_transform_name_callback(raw_cb: *mut c_void) -> *const c_char {
@@ -109,16 +106,9 @@ pub unsafe extern "C" fn transform_callback(
 ) -> *mut c_char {
     let cb = &mut *(raw_cb as *mut TransformCallback);
     let key = slice::from_raw_parts(raw_key as *const u8, key_len as usize);
-    let mut result = (cb.transform_fn)(key);
-    result.shrink_to_fit();
-
-    // copy the result into a C++ destroyable buffer
-    let buf = libc::malloc(result.len() as size_t);
-    assert!(!buf.is_null());
-    ptr::copy(result.as_ptr() as *mut c_void, &mut *buf, result.len());
-
-    *dst_length = result.len() as size_t;
-    buf as *mut c_char
+    let prefix = (cb.transform_fn)(key);
+    *dst_length = prefix.len() as size_t;
+    prefix.as_ptr() as *mut c_char
 }
 
 pub unsafe extern "C" fn in_domain_callback(
@@ -128,10 +118,6 @@ pub unsafe extern "C" fn in_domain_callback(
 ) -> u8 {
     let cb = &mut *(raw_cb as *mut TransformCallback);
     let key = slice::from_raw_parts(raw_key as *const u8, key_len as usize);
-
-    if (cb.in_domain_fn.unwrap())(key) {
-        1
-    } else {
-        0
-    }
+    let in_domain = cb.in_domain_fn.unwrap();
+    in_domain(key) as u8
 }
--- a/tests/test_slice_transform.rs
+++ b/tests/test_slice_transform.rs
@@ -13,8 +13,8 @@ pub fn test_slice_transform() {
         let b1: Box<[u8]> = key(b"bbb1");
         let b2: Box<[u8]> = key(b"bbb2");
 
-        fn first_three(k: &[u8]) -> Vec<u8> {
-            k.iter().take(3).cloned().collect()
+        fn first_three<'a>(k: &'a [u8]) -> &'a [u8] {
+            &k[..3]
         }
 
         let prefix_extractor = SliceTransform::create("first_three", first_three, None);


ekmartin:
On a related note, I was trying to write a test for `in_domain` but I couldn't actually trigger a case where `in_domain_fn` got called. I can dig a little further later, but perhaps it's only used when iterating through what's on disk?

Failing test that I'd expect to pass, on master:

```rust
#[test]
pub fn test_in_domain() {
    let n = DBPath::new("_rust_rocksdb_slicetransform_in_domain_test");
    {
        let a1: Box<[u8]> = key(b"aaa1");
        let a2: Box<[u8]> = key(b"aaa2");
        let b1: Box<[u8]> = key(b"bbb1");
        let b2: Box<[u8]> = key(b"bbb2");

        fn first_three(k: &[u8]) -> Vec<u8> {
            k.iter().take(3).cloned().collect()
        }

        fn in_domain(k: &[u8]) -> bool {
            k.starts_with(b"a")
        }

        let prefix_extractor = SliceTransform::create("first_three", first_three, Some(in_domain));

        opts.create_if_missing(true);
        opts.set_prefix_extractor(prefix_extractor);

        let db = DB::open(&opts, &n).unwrap();

        assert!(db.put(&*a1, &*a1).is_ok());
        assert!(db.put(&*a2, &*a2).is_ok());
        assert!(db.put(&*b1, &*b1).is_ok());
        assert!(db.put(&*b2, &*b2).is_ok());

        fn cba(input: &[u8]) -> Box<[u8]> {
            input.to_vec().into_boxed_slice()
        }

        fn key(k: &[u8]) -> Box<[u8]> {
            k.to_vec().into_boxed_slice()
        }

        {
            let expected = vec![(cba(&a1), cba(&a1)), (cba(&a2), cba(&a2))];
            let a_iterator = db.prefix_iterator(b"aaa");
            assert_eq!(a_iterator.collect::<Vec<_>>(), expected)
        }

        {
            let expected = vec![];
            let b_iterator = db.prefix_iterator(b"bbb");
            // My understanding was that this would be empty here, but it seems
            // like it's actually the entire list of "bbb" prefixes instead.
            // Maybe it's reading straight from the memtable and ignoring
            // `in_domain`?
            assert_eq!(b_iterator.collect::<Vec<_>>(), expected)
        }
    }
}
```

jonhoo:
Maybe the return value should be `Cow<'a, [u8]>`? That way users could choose whether to allocate or not depending on whether they want to create a new key or not.

ekmartin:
But who would free the allocations in that case? `transform_callback` is called from RocksDB through `rocksdb_slicetransform_t::Transform` and from what I can tell it doesn't really seem like it tries to free the resulting `Slice` afterwards—I might be missing a way to make this work though.

Here's `valgrind --leak-check=yes` on master:

```code
==330== HEAP SUMMARY:
==330==     in use at exit: 3,601,868 bytes in 200,091 blocks
==330==   total heap usage: 1,469,901 allocs, 1,269,810 frees, 372,726,172 bytes allocated
==330==
==330== 1,548 bytes in 86 blocks are definitely lost in loss record 6 of 8
==330==    at 0x4C2BE2D: malloc (vg_replace_malloc.c:299)
==330==    by 0x207021: rocksdb::slice_transform::transform_callback (in /target/release/rocksdb-leak)
==330==    by 0x208979: rocksdb_slicetransform_t::Transform(rocksdb::Slice const&) const (in /target/release/rocksdb-leak)
==330==    by 0x26E4AE: rocksdb::DBIter::FindNextUserEntryInternal(bool, bool) (in /target/release/rocksdb-leak)
==330==    by 0x26F078: rocksdb::DBIter::Next() (in /target/release/rocksdb-leak)
==330==    by 0x20788D: <rocksdb::db::DBIterator as core::iter::traits::iterator::Iterator>::next (in /target/release/rocksdb-leak)
==330==    by 0x203C84: <alloc::vec::Vec<T> as alloc::vec::SpecExtend<T,I>>::from_iter (in /target/release/rocksdb-leak)
==330==    by 0x202D9C: rocksdb_leak::main (in /target/release/rocksdb-leak)
==330==    by 0x206912: std::rt::lang_start::{{closure}} (in /target/release/rocksdb-leak)
==330==    by 0x473AB2: {{closure}} (rt.rs:49)
==330==    by 0x473AB2: std::panicking::try::do_call (panicking.rs:293)
==330==    by 0x475FD9: __rust_maybe_catch_panic (lib.rs:87)
==330==    by 0x4745BC: try<i32,closure> (panicking.rs:272)
==330==    by 0x4745BC: catch_unwind<closure,i32> (panic.rs:388)
==330==    by 0x4745BC: std::rt::lang_start_internal (rt.rs:48)
==330==
==330== 1,800,000 bytes in 100,000 blocks are definitely lost in loss record 7 of 8
==330==    at 0x4C2BE2D: malloc (vg_replace_malloc.c:299)
==330==    by 0x207021: rocksdb::slice_transform::transform_callback (in /target/release/rocksdb-leak)
==330==    by 0x208979: rocksdb_slicetransform_t::Transform(rocksdb::Slice const&) const (in /target/release/rocksdb-leak)
==330==    by 0x26F960: rocksdb::DBIter::Seek(rocksdb::Slice const&) (in /target/release/rocksdb-leak)
==330==    by 0x20A8CB: rocksdb_iter_seek (in /target/release/rocksdb-leak)
==330==    by 0x207669: rocksdb::db::DBIterator::new (in /target/release/rocksdb-leak)
==330==    by 0x202D31: rocksdb_leak::main (in /target/release/rocksdb-leak)
==330==    by 0x206912: std::rt::lang_start::{{closure}} (in /target/release/rocksdb-leak)
==330==    by 0x473AB2: {{closure}} (rt.rs:49)
==330==    by 0x473AB2: std::panicking::try::do_call (panicking.rs:293)
==330==    by 0x475FD9: __rust_maybe_catch_panic (lib.rs:87)
==330==    by 0x4745BC: try<i32,closure> (panicking.rs:272)
==330==    by 0x4745BC: catch_unwind<closure,i32> (panic.rs:388)
==330==    by 0x4745BC: std::rt::lang_start_internal (rt.rs:48)
==330==    by 0x2030A1: main (in /target/release/rocksdb-leak)
==330==
==330== 1,800,000 bytes in 100,000 blocks are definitely lost in loss record 8 of 8
==330==    at 0x4C2BE2D: malloc (vg_replace_malloc.c:299)
==330==    by 0x207021: rocksdb::slice_transform::transform_callback (in /target/release/rocksdb-leak)
==330==    by 0x208979: rocksdb_slicetransform_t::Transform(rocksdb::Slice const&) const (in /target/release/rocksdb-leak)
==330==    by 0x26E4AE: rocksdb::DBIter::FindNextUserEntryInternal(bool, bool) (in /target/release/rocksdb-leak)
==330==    by 0x26F652: rocksdb::DBIter::Seek(rocksdb::Slice const&) (in /target/release/rocksdb-leak)
==330==    by 0x20A8CB: rocksdb_iter_seek (in /target/release/rocksdb-leak)
==330==    by 0x207669: rocksdb::db::DBIterator::new (in /target/release/rocksdb-leak)
==330==    by 0x202D31: rocksdb_leak::main (in /target/release/rocksdb-leak)
==330==    by 0x206912: std::rt::lang_start::{{closure}} (in /target/release/rocksdb-leak)
==330==    by 0x473AB2: {{closure}} (rt.rs:49)
==330==    by 0x473AB2: std::panicking::try::do_call (panicking.rs:293)
==330==    by 0x475FD9: __rust_maybe_catch_panic (lib.rs:87)
==330==    by 0x4745BC: try<i32,closure> (panicking.rs:272)
==330==    by 0x4745BC: catch_unwind<closure,i32> (panic.rs:388)
==330==    by 0x4745BC: std::rt::lang_start_internal (rt.rs:48)
==330==
==330== LEAK SUMMARY:
==330==    definitely lost: 3,601,548 bytes in 200,086 blocks
==330==    indirectly lost: 0 bytes in 0 blocks
==330==      possibly lost: 0 bytes in 0 blocks
==330==    still reachable: 320 bytes in 5 blocks
==330==         suppressed: 0 bytes in 0 blocks
==330== Reachable blocks (those to which a pointer was found) are not shown.
```

jonhoo:
I guess I was thinking that whatever _currently_ frees the `Vec` would free the `Cow`? If you're talking about the memory the `&'a [u8]` is pointing into, then isn't that just the same as the proposed signature where you return `&'a [u8]` directly?

ekmartin:
The problem is that there's nothing that currently fees the `Vec`. If you look at the `valgrind` output in the comment above none of the `transform_callback` allocations are actually freed, which is why I think changing the signature to `-> &'a [u8]` is the right choice here. Sorry, buried the lede in the original post—I wasn't sure it was a leak, whereas I'm fairly confident it is now.

jonhoo:
Ah, I see, I missed that part! Yes, if nothing currently frees, I agree with you that the current signature is just plain wrong, and `&'a [u8]` is the right thing to do. I do wonder whether there might be a way to _make_ it free, though that may be a question for a separate PR. If we _do_ figure out a way to make it free, then `Cow<'a, [u8]>` seems like a good combination of both approaches.

iSynaptic:
The [code comments](https://github.com/facebook/rocksdb/blob/master/include/rocksdb/slice_transform.h) for `SliceTransform` provided some insight on this. Its use is intended for providing prefixes of existing keys, not to generate arbitrarily new keys (ie. provide slices of existing memory, not allocate new memory). Based on that, I think this change makes sense, since without this change it's possible to allocate memory that leaks.

jonhoo:
@iSynaptic :+1: I _think_ this is a backwards-incompatible change, so may be worth flagging that.

iSynaptic:
@jonhoo Agreed, but also necessary, since this (we believe) is the contract intended by the underlying RocksDB library. I suspect there will be many more backwards-incompatible changes in the near future as we drive towards a 1.0 release.  Better to make those changes now and get us to a stable footing...

jonhoo:
I think this is also the last remaining change necessary for us to move [Noria](https://github.com/mit-pdos/noria) onto upstream `rocksdb` again :tada:

iSynaptic:
@jonhoo Awesome. I'd love to hear more about where you are going with Noria and where this library presents roadblocks. I've got a big PR (#268) in the works and I'm trying to keep the backwards-incompatible changes to a minimum. Maybe Noria would be a good test case to vet the level of work needed to upgrade once #268 is merged/released.

jonhoo:
@ekmartin has the best insight into that I believe, so he can answer to that better than I can.

src/slice_transform.rs
```diff
@@ -41,15 +39,15 @@ impl SliceTransform {
         transform_fn: TransformFn,
         in_domain_fn: Option<InDomainFn>,
     ) -> SliceTransform {
-        let cb = Box::new(TransformCallback {
+        let cb = Box::into_raw(Box::new(TransformCallback {
             name: CString::new(name.as_bytes()).unwrap(),
```
aleksuss: I'd propose to handle the `CString` creation. IMO, 'unwrap()` is a bad idea to use here.

src/slice_transform.rs
```diff
@@ -128,10 +118,6 @@ pub unsafe extern "C" fn in_domain_callback(
 ) -> u8 {
     let cb = &mut *(raw_cb as *mut TransformCallback);
     let key = slice::from_raw_parts(raw_key as *const u8, key_len as usize);
-
-    if (cb.in_domain_fn.unwrap())(key) {
-        1
-    } else {
-        0
-    }
+    let in_domain = cb.in_domain_fn.unwrap();
```
aleksuss: Same here

ekmartin: This is called by RocksDB when we've provided an `in_domain` callback. `in_domain_fn` being `None` here should result in a loud error. If you think it makes it better we could call `unreachable!()`.

src/slice_transform.rs
ekmartin: I agree, but I'm not sure if making this even more of a breaking change is worth it from a pragmatic perspective. What do you think?

ekmartin:
@iSynaptic: shouldn't be too bad, but if you ping me when #268 is ready I can try upgrading and leave some feedback.

src/slice_transform.rs
aleksuss: @ekmartin you mean if we add a handle then we should change a return type to `Option<SliseTransform>` or something like this ?

iSynaptic:
What are the outstanding issues with this PR? Removing `.unwrap()` calls?

aleksuss:
Let's leave it as is. But in the future we need to review the code and remove `unwrap()`s and `expect()`s where it's possible. Agree ?

iSynaptic:
@aleksuss Agreed.

Issue #273, opened 2019-03-03
Is a new release upcoming?


elichai:
Hi,
Just wanted to ask if there's an ETA for the next release so I could stop using the old fork(https://github.com/elichai/rust-rocksdb/) that let's me pass ReadOptions for the DBIterator :)

iSynaptic:
I don't think we've got a specific date picked for the next release. We've got a few PRs in flight - I suspect once they are merged we'll have a release.

elichai:
Thanks!

iSynaptic:
@elichai I just release 0.12.0, which should have the changes you are looking for. Let us know if it works out for you.

elichai:
@iSynaptic I just looked at it and I see that `DB` has a `iterator_opt()` function but only Snapshot has a `iterator_cf_opt()`, am I missing something or making a cf iterator with options is missing?
If so I can make a PR for it quickly.

Thanks.

iSynaptic:
@elichai I think you are right. Feel free to submit a PR. Otherwise, it will be addressed by #268.

PR #284, opened 2019-03-17
Added a function to create iterator with specific CF and ReadOptions

elichai:
Added after Discussion with @iSynaptic https://github.com/rust-rocksdb/rust-rocksdb/issues/273

### DIFF
--- a/src/db.rs
+++ b/src/db.rs
@@ -1093,6 +1093,15 @@ impl DB {
         DBIterator::new(self, &readopts, mode)
     }
 
+    pub fn iterator_cf_opt(
+        &self,
+        cf_handle: ColumnFamily,
+        readopts: &ReadOptions,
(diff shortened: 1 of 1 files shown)


src/db.rs
```diff
@@ -1093,6 +1093,15 @@ impl DB {
         DBIterator::new(self, &readopts, mode)
     }
 
+    pub fn iterator_cf_opt(
```
aleksuss: @elichai It would be nice to add some description for public method

elichai: Done.

elichai:
Thanks!
@iSynaptic If you could publish a new version it would be great :)

iSynaptic:
See #287

iSynaptic:
Just released 0.12.1.

PR #288, opened 2019-03-28
Added raw_iterator_opt method to DB

xuchaoqian:
Now we can use the DBRawIterator with customized opts directly:

```
...
        let mut opts = ReadOptions::default();
        opts.set_prefix_same_as_start(true);
        let mut iter: DBRawIterator = db.raw_iterator_opt(opts);
        iter.seek(b"prefix");
...
```

### DIFF
--- a/src/db.rs
+++ b/src/db.rs
@@ -1162,6 +1162,10 @@ impl DB {
         DBRawIterator::new(self, &opts)
     }
 
+    pub fn raw_iterator_opt(&self, opts: ReadOptions) -> DBRawIterator {
+        DBRawIterator::new(self, &opts)
+    }
+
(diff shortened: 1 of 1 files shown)


src/db.rs
```diff
@@ -1162,6 +1162,10 @@ impl DB {
         DBRawIterator::new(self, &opts)
     }
 
+    pub fn raw_iterator_opt(&self, opts: ReadOptions) -> DBRawIterator {
```
aleksuss: Why do you pass `ReadOptions` by value ?

```suggestion
    pub fn raw_iterator_opt(&self, opts: &ReadOptions) -> DBRawIterator {
```

iSynaptic: Oh, good catch!

aleksuss:
A source branch doesn't exist.

PR #289, opened 2019-03-28
Update DB::compact_range_cf() arguments to be consistent with other methods

romanz:
Following https://github.com/rust-rocksdb/rust-rocksdb/commit/21b9a9e98aa5f2d6d3a04cab2cc2dedfa3011fb6.

### DIFF
--- a/src/db.rs
+++ b/src/db.rs
@@ -1362,8 +1362,16 @@ impl DB {
         }
     }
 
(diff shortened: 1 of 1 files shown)


iSynaptic:
@aleksuss I agree this is a breaking change, but we have been making breaking changes in an effort to make the API more consistent and flexible. I'd like to see this change in before we reach 1.0 when breaking changes are much more difficult (ie. requires a new major release).

Issue #134, opened 2017-08-27
add support for delete_range


benoitc:
would be cool to have the support for this feature to be able to delete in a range directly instead of launching an iterator then delete.

iSynaptic:
Fixed by #290.

PR #290, opened 2019-03-28
Updates RocksDB to 5.18.3 & implement delete_range and delete_range_cf

baptistejamin:
I updated  RocksDB to 5.18.3.

The 5.18.X version includes a patch solving a deadlock.

The 5.18.3 version also includes new features like delete_range and delete_range_cf (https://github.com/facebook/rocksdb/wiki/DeleteRange)


iSynaptic:
Fixes #134.

src/db.rs
```diff
@@ -1649,6 +1649,40 @@ impl WriteBatch {
         }
     }
 
+    pub fn delete_range<K: AsRef<[u8]>>(&mut self, start_key: K,  end_key: K) -> Result<(), Error> {
```
iSynaptic: Can you add some doc comments here and either a test or doc example so the CI build can validate this works?

src/db.rs
```diff
@@ -1649,6 +1649,40 @@ impl WriteBatch {
...
+                start_key.as_ptr() as *const c_char,
+                start_key.len() as size_t,
+                end_key.as_ptr() as *const c_char,
+                end_key.len() as size_t,
+            );
+            Ok(())
+        }
+    }
+
+    pub fn delete_range_cf<K: AsRef<[u8]>>(&mut self, cf: ColumnFamily, start_key: K,  end_key: K) -> Result<(), Error> {
```
iSynaptic: Can you add some doc comments here and either a test or doc example so the CI build can validate this works?

valeriansaliou:
Thanks for the quick reply @iSynaptic ; I'll handle this on behalf of @baptistejamin

valeriansaliou:
All done. I added a descriptive comment to `delete_range()` and tests.

src/db.rs
```diff
@@ -1649,6 +1649,48 @@ impl WriteBatch {
         }
     }
 
+    /// Remove database entries from start key to end key.
+    ///
+    /// Removes the database entries in the range ["begin_key", "end_key"), i.e.,
+    /// including "begin_key" and excluding "end_key". It is not an error if no
+    /// keys exist in the range ["begin_key", "end_key").
+    pub fn delete_range<K: AsRef<[u8]>>(&mut self, start_key: K, end_key: K) -> Result<(), Error> {
```
aleksuss: What do you think to rename params like this ?
```suggestion
    pub fn delete_range<K: AsRef<[u8]>>(&mut self, from: K, to: K) -> Result<(), Error> {
```

valeriansaliou: Go for it!

aleksuss:
@valeriansaliou Apply `cargo fmt` to the code please.

valeriansaliou:
Done!

src/db.rs
```diff
@@ -1649,6 +1649,47 @@ impl WriteBatch {
...
+                start_key.as_ptr() as *const c_char,
+                start_key.len() as size_t,
+                end_key.as_ptr() as *const c_char,
+                end_key.len() as size_t,
+            );
+            Ok(())
+        }
+    }
+
+    pub fn delete_range_cf<K: AsRef<[u8]>>(
```
aleksuss: Sorry if I'm too carping but I'd like to see some doc-comments for this PUBLIC method. It's because the `docs.rs` service won't be able to create a description for the public method. And I suppose it's bad. Any thoughts ?

iSynaptic: Agreed. I think it's the last thing we need before a merge.

aleksuss: Exactly !

valeriansaliou: Sure, I'm on it!

valeriansaliou: All done.

PR #292, opened 2019-04-01
add comment for ReadOptions.set_iterate_upper_bound method.

xiaobogaga:
the method `ReadOptions.set_iterate_upper_bound` method seems confusing, because the upper_bound is not returned when doing iteration.  some comment are added for this method.

### DIFF
--- a/src/db.rs
+++ b/src/db.rs
@@ -1768,9 +1768,10 @@ impl ReadOptions {
         }
     }
 
+    /// set_iterate_upper_bound sets the upper bound for an iterator, and the upper bound itself
+    /// is not included on the iteration result.
     pub fn set_iterate_upper_bound<K: AsRef<[u8]>>(&mut self, key: K) {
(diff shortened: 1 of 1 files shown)


xiaobogaga:
@iSynaptic could you please take a look, thanks.

src/db.rs
```diff
@@ -1768,9 +1768,10 @@ impl ReadOptions {
         }
     }
 
+    /// set_iterate_upper_bound sets the upper bound for an iterator, and the upper bound itself
```
aleksuss: IMO `set_iterate_upper_bound` is redundant here, isn't it ?

iSynaptic: Yes, I would say it is.

xiaobogaga: thanks, I would fix it.

xiaobogaga:
@aleksuss PTAL again, thanks.

PR #294, opened 2019-04-16
Add lifetimes for `DBIterator` return types 

ngotchac:
Without them, the added simple test case doesn't compile.

### DIFF
--- a/src/db.rs
+++ b/src/db.rs
@@ -559,7 +559,7 @@ impl<'a> Snapshot<'a> {
         }
     }
 
-    pub fn iterator(&self, mode: IteratorMode) -> DBIterator {
+    pub fn iterator(&self, mode: IteratorMode) -> DBIterator<'a> {
         let readopts = ReadOptions::default();
         self.iterator_opt(mode, readopts)
     }
@@ -573,7 +573,7 @@ impl<'a> Snapshot<'a> {
         self.iterator_cf_opt(cf_handle, readopts, mode)
     }
 
-    pub fn iterator_opt(&self, mode: IteratorMode, mut readopts: ReadOptions) -> DBIterator {
+    pub fn iterator_opt(&self, mode: IteratorMode, mut readopts: ReadOptions) -> DBIterator<'a> {
         readopts.set_snapshot(self);
         DBIterator::new(self.db, &readopts, mode)
     }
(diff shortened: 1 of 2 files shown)


iSynaptic:
Restarting build since the failure seemed to be a transient issue.

iSynaptic:
Closing and re-opening to trigger build...

iSynaptic:
This code compiles without any changes (there was a transient build server issue that has resolved itself).  This adds lifetimes to `DBIterator`, which I believe is correct since an iterators life is tied to the life of the `DB`, but that means it's a breaking change. Better to get the lifetimes right now than want to change them after a `1.0` release.

@aleksuss @vitvakatu Thoughts?

iSynaptic:
Worth noting I made the same change in my draft PR #268:

https://github.com/rust-rocksdb/rust-rocksdb/pull/268/files#diff-d1e45ae73f0ad58210abebd18b410da6R115

nbdd0121:
Can we also get the get APIs sorted out properly as well, consider that this PR introduces a breaking change anyway? We should remove the existing `get` APIs and rename the `get_pinned` once to `get` (see #337). `get_pinned` is much faster than `get`, except that it returns `'self` instead `'static`.

aleksuss:
@nbdd0121 Have you done benchmarks to measure that a `get_pinned` is really faster then `get` ?

nbdd0121:
> @nbdd0121 Have you done benchmarks to measure that a `get_pinned` is really faster then `get` ?

Yes. It has very large performance advantage:
```
test get_large                 ... bench:     153,760 ns/iter (+/- 5,751)
test get_pinned_large          ... bench:       3,147 ns/iter (+/- 203)
test get_pinned_small          ... bench:         141 ns/iter (+/- 5)
test get_pinned_to_owned_large ... bench:       6,804 ns/iter (+/- 485)
test get_pinned_to_owned_small ... bench:         157 ns/iter (+/- 5)
test get_small                 ... bench:         156 ns/iter (+/- 9)
```

Small payload is 5 bytes, and large payload is the full lorem ipsum text repeated 100 times. "get_pinned_to_owned" uses "slice.as_ref().to_owned()" which converts `DBPinnableSlice` to `Vec<u8>`. Do you want me to create a PR which switchs `get` to use pinned APIs, and deprecate `get_pinned`?

PR #295, opened 2019-04-21
enable sse4.2/pclmul for accelerated crc32c

yjh0502:
Same with https://github.com/rust-rocksdb/rust-rocksdb/pull/167, some comments are borrows from the PR.

### DIFF
--- a/librocksdb-sys/build.rs
+++ b/librocksdb-sys/build.rs
@@ -90,6 +90,18 @@ fn build_rocksdb() {
         .filter(|file| *file != "util/build_version.cc")
         .collect::<Vec<&'static str>>();
 
+    if cfg!(target_arch = "x86_64") {
+        // This is needed to enable hardware CRC32C. Technically, SSE 4.2 is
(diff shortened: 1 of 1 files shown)


librocksdb-sys/build.rs
```diff
@@ -90,6 +90,18 @@ fn build_rocksdb() {
         .filter(|file| *file != "util/build_version.cc")
         .collect::<Vec<&'static str>>();
 
+    if cfg!(target_arch = "x86_64") {
+        // This is needed to enable hardware CRC32C. Technically, SSE 4.2 is
+        // only available since Intel Nehalem (about 2010) and AMD Bulldozer
+        // (about 2011).
+        config.define("HAVE_PCLMUL", Some("1"));
+        config.define("HAVE_SSE42", Some("1"));
+        config.flag("-msse2");
```
aleksuss: Is there any reason to use `.flag_if_supported()` here or it's redundant and every cpu with arch `x86_64` supports these instructions ?

PR #296, opened 2019-04-21
Add `rocksdb_flush` method on DB

valeriansaliou:
As `rust-rocksdb` already had the `disable_wal` method on `WriteOptions` (see: https://docs.rs/rocksdb/0.12.1/rocksdb/struct.WriteOptions.html#method.disable_wal), it makes sense to implement the `DB::flush()` method that lets the user perform a manual memtable flush to SST files.

This is useful when performing all writes to RocksDB in `disable_wal=true` mode; in such case upon process shutdown (eg. SIGTERM) memtables should be manually flushed to commit changes to disk and avoid data loss.

This `disable_wal` + manual `flush` pattern can be useful on busy servers using RocksDB on SSDs as to minimize SSD write wear.

### DIFF
--- a/src/db.rs
+++ b/src/db.rs
@@ -15,7 +15,7 @@
 
 use ffi;
 use ffi_util::opt_bytes_to_ptr;
-use {ColumnFamily, ColumnFamilyDescriptor, Error, Options, WriteOptions, DB};
+use {ColumnFamily, ColumnFamilyDescriptor, Error, Options, FlushOptions, WriteOptions, DB};
 
 use libc::{self, c_char, c_int, c_uchar, c_void, size_t};
 use std::collections::BTreeMap;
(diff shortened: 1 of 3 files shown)


valeriansaliou:
Implemented and tested on Sonic as of https://github.com/valeriansaliou/sonic/commit/4e26de0963cb5e3a9d752ff9b638d875c61e06b4

src/db.rs
```diff
@@ -836,6 +836,17 @@ impl DB {
         &self.path.as_path()
     }
 
+    pub fn flush_opt(&self, flushopts: &FlushOptions) -> Result<(), Error> {
```
aleksuss: No docs for public method

valeriansaliou: Done.

src/db_options.rs
```diff
@@ -1208,6 +1216,28 @@ impl Default for Options {
     }
 }
 
+impl FlushOptions {
+    pub fn new() -> FlushOptions {
+        FlushOptions::default()
+    }
+
+    pub fn set_wait(&mut self, wait: bool) {
```
aleksuss: And here as well please. Also, It could be cool to add a doc-test here.

valeriansaliou: Done.

PR #298, opened 2019-04-30
Revert back to `&mut self` for `create_cf`/`drop_cf`

dvdplm:
ref. https://github.com/rust-rocksdb/rust-rocksdb/issues/279

Essentially reverts https://github.com/rust-rocksdb/rust-rocksdb/pull/197 and makes `create_cf`/`drop_cf` take `&mut self` to ensure nothing else can update the column family map.

The biggest breaking change here is that `create_cf` does not return a `ColumnFamily`. See [this comment](https://github.com/rust-rocksdb/rust-rocksdb/issues/279#issuecomment-487902919) on the original ticket for details.

### DIFF
--- a/src/db.rs
+++ b/src/db.rs
@@ -28,7 +28,6 @@ use std::path::Path;
 use std::ptr;
 use std::slice;
 use std::str;
-use std::sync::{Arc, RwLock};
 
 unsafe impl Send for DB {}
 unsafe impl Sync for DB {}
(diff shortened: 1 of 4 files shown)


vitvakatu:
Superseded by https://github.com/rust-rocksdb/rust-rocksdb/pull/314, I believe

aleksuss:
Close due to #314

PR #300, opened 2019-04-30
Add raw_iterator_cf_opt to the DB API

rnarubin:
The underlying functionality already existed, but it wasn't exposed to
the top-level API.

This is essentially the same as #288 but exposes the column family iterator instead of the default one

### DIFF
--- a/src/db.rs
+++ b/src/db.rs
@@ -588,21 +588,25 @@ impl<'a> Snapshot<'a> {
         DBIterator::new_cf(self.db, cf_handle, &readopts, mode)
     }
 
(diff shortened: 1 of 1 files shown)


src/db.rs
```diff
@@ -1167,6 +1167,14 @@ impl DB {
         DBRawIterator::new_cf(self, cf_handle, &opts)
     }
 
+    pub fn raw_iterator_cf_opt(
```
aleksuss: @rnarubin  I'd like to see some description for this **public** method

rnarubin: @aleksuss you got it, pushed a change

PR #302, opened 2019-05-03
Expose the `status` method on iterators

rnarubin:
This allows handling errors when they're encountered during iteration of
the DB. The details of RocksDB iterator error reporting is described in
https://github.com/facebook/rocksdb/wiki/Iterator#error-handling

This method alone is not especially ergonomic or idiomatic for rust:
users would have to call `status` on their iterators whenever they
receive a `None` item to check whether iteration actually failed instead
of simply ended. A more complete approach is to change the iterator
signature to `Item = Result<..., Error>` and internally have the
iterator check its status. This better reflects the fact that an error
can be encountered during iteration, and that the user should handle or
propagate them explicitly. Such an approach would however be a breaking
change, and should be approached as a larger undertaking than this
changeset's simple addition.

### DIFF
--- a/src/db.rs
+++ b/src/db.rs
@@ -232,9 +232,26 @@ impl<'a> DBRawIterator<'a> {
         }
     }
 
-    /// Returns true if the iterator is valid.
+    /// Returns `true` if the iterator is valid. An iterator is invalidated when
+    /// it reaches the end of its defined range, or when it encounters an error.
+    ///
+    /// To check whether the iterator encountered an error after `valid` has
+    /// returned `false`, use the [`status`](DBRawIterator::status) method. `status` will never
+    /// return an error when `valid` is `true`.
     pub fn valid(&self) -> bool {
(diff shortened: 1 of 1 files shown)


vitvakatu:
IMO, the approach with iterator over `Result`s is much more appropriate here. I've seen this in different Rust libraries - it's not nice, but works well and forces the user to actually check for possible errors.

rnarubin:
@aleksuss I can add some sunny-day unit tests fairly easily. Writing a test that encounters an error strikes me as more difficult though -- what's a reliable/portable way to induce an error during iteration? It might be possible to generate an error through a custom Env, but as far as I can tell the C api doesn't provide a means of implementing an Env.

@vitvakatu I agree using Results would be more appropriate. This is a small change to support error checking _at all_ in the mean time (previously errors were completely ignored!), because overhauling the DBIterator to use Results would be a large breaking change. It's worth making that break in my opinion, but it's a large scope

vitvakatu:
I think we can merge it as-is and create a task for further improvements.

PR #313, opened 2019-06-25
Implement Send and Sync for Snapshot

pavel-mukhanov:
As far it is safe to share `Snapshot` between threads I've implemented `Send` and `Sync` for it.

[Snapshot documentations from sources.](https://github.com/facebook/rocksdb/blob/189f0c27aaecdf17ae7fc1f826a423a28b77984f/include/rocksdb/snapshot.h#L15)

### DIFF
--- a/tests/test_db.rs
+++ b/tests/test_db.rs
@@ -12,8 +12,6 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
-#![feature(thread_spawn_unchecked)]
-
 extern crate libc;
 extern crate rocksdb;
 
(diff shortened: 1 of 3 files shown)


aleksuss:
@pavel-mukhanov Could you please add some tests for that ?

tests/test_db.rs
```diff
@@ -163,6 +165,45 @@ fn snapshot_test() {
     }
 }
 
+#[derive(Clone)]
```
aleksuss: #[cfg(test)] ??? We don't need a `SnapshotWrapper` in a production code, do we ?

pavel-mukhanov: `SnapshotWrapper` is in the tests directory, therefore, it is not visible in the main crate.

aleksuss: Oh. It's ok then.

PR #314, opened 2019-06-26
Change a column family storing

aleksuss:
Column families store in `BTreeMap` without `Arc<RwLock>`. So methods `create_cf` and `drop_cf` become mutable. Also `cf_handle` returns a reference to `ColumnFamily` struct.


src/lib.rs
```diff
@@ -283,10 +281,8 @@ pub struct WriteOptions {
 
 /// An opaque type used to represent a column family. Returned from some functions, and used
 /// in others
-#[derive(Copy, Clone)]
-pub struct ColumnFamily<'a> {
+pub struct ColumnFamily {
```
vitvakatu: I wonder if the removed lifetime constrain may affect the safety of an API. Are there any case when usage of invalid ColumnFamily instance is possible?

aleksuss: @vitvakatu Now public API returns a reference to `ColumnFamily` only. So it's not possible.

PR #318, opened 2019-07-19
Implement Send and Sync for useful types

rnarubin:
The Send and Sync traits aren't auto-implemented on the many types which
wrap a raw pointer to some RocksDB struct. Most of these types are
nevertheless safe to use in Send and Sync context, so they should be
marked as such

### DIFF
--- a/src/db.rs
+++ b/src/db.rs
@@ -199,8 +199,6 @@ pub struct DBIterator<'a> {
     just_seeked: bool,
 }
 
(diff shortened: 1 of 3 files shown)


vitvakatu:
Thank you for your PR, please rebase onto latest master - you have some merge conflicts.

rnarubin:
Sure thing, rebased

PR #319, opened 2019-07-22
Build RocksDB with support for unicode filenames on Windows

### LINKED ISSUE #85: [Windows] Unicode path issue
On Windows, creating a database with unicode characters in the filename creates two directories: an empty one with the correct name, and one with the actual database in it but with a mangled name.

For example:
```rust
DB::open_default("_rust_rocksdb_u͔n͈̰̎i̙̮͚̦c͚̉o̼̩̰͗d͔̆̓ͥé");
```

Creates two folders:
 - ``_rust_rocksdb_u͔n͈̰̎i̙̮͚̦c͚̉o̼̩̰͗d͔̆̓ͥé`` (empty)
 - ``_rust_rocksdb_uÍ”nÌŽÍˆÌ°iÌ™Ì®ÍšÌ¦cÌ‰ÍšoÍ—Ì¼Ì©Ì°dÌ†ÍƒÍ¥Í”eÌ``  (with data inside it)

Linux works as expected (creates one folder with correct name)

udoprog:
This causes RocksDB to wrap all fs operations using the wide filesystem apis:
<https://github.com/facebook/rocksdb/blob/master/port/win/port_win.h#L345>

Fixes #85 - even though it's already closed.

This problem was originally [punted by RocksDB developers](https://github.com/facebook/rocksdb/issues/3408), but a solution that doesn't require a custom environment was merged in [October last year](https://github.com/facebook/rocksdb/pull/4469).

### DIFF
--- a/librocksdb-sys/build.rs
+++ b/librocksdb-sys/build.rs
@@ -123,6 +123,7 @@ fn build_rocksdb() {
         link("rpcrt4", false);
         link("shlwapi", false);
         config.define("OS_WIN", Some("1"));
+        config.define("ROCKSDB_WINDOWS_UTF8_FILENAMES", Some("1"));
 
         // Remove POSIX-specific sources
         lib_sources = lib_sources


udoprog:
Not sure if build failure is caused by this change. I've seen similar things happen using build caching before.

aleksuss:
@udoprog It happens sometime on Travis. I've restarted bad job and It's ok now.

Issue #322, opened 2019-08-26
Cross compilation fails.


bddap:
## To reproduce (macos)

```bash
brew install FiloSottile/musl-cross/musl-cross
rustup target add x86_64-unknown-linux-musl
CXX_x86_64-unknown-linux-musl=x86_64-linux-musl-gcc CC_x86_64-unknown-linux-musl=x86_64-linux-musl-g++ cargo build --release --target x86_64-unknown-linux-musl
```

<details>

<summary>Error log</summary>

```
   Compiling librocksdb-sys v6.1.2 (/Users/a/d/tmp/rust-rocksdb/librocksdb-sys)
error: failed to run custom build command for `librocksdb-sys v6.1.2 (/Users/a/d/tmp/rust-rocksdb/librocksdb-sys)`

Caused by:
  process didn't exit successfully: `/Users/a/d/tmp/rust-rocksdb/target/release/build/librocksdb-sys-dd7419cf82e92f3f/build-script-build` (exit code: 1)
--- stdout
cargo:rerun-if-changed=build.rs
cargo:rerun-if-changed=rocksdb/
cargo:rerun-if-changed=snappy/
cargo:rerun-if-changed=lz4/
cargo:rerun-if-changed=zstd/
cargo:rerun-if-changed=zlib/
cargo:rerun-if-changed=bzip2/
cargo:warning=couldn't execute `llvm-config --prefix` (error: No such file or directory (os error 2))
cargo:warning=set the LLVM_CONFIG_PATH environment variable to a valid `llvm-config` executable
NUM_JOBS = Some("8")
TARGET = Some("x86_64-unknown-linux-musl")
OPT_LEVEL = Some("3")
HOST = Some("x86_64-apple-darwin")
CXX_x86_64-unknown-linux-musl = Some("x86_64-linux-musl-gcc")
CXXFLAGS_x86_64-unknown-linux-musl = None
CXXFLAGS_x86_64_unknown_linux_musl = None
TARGET_CXXFLAGS = None
CXXFLAGS = None
CRATE_CC_NO_DEFAULTS = None
DEBUG = Some("false")
CARGO_CFG_TARGET_FEATURE = Some("crt-static,fxsr,sse,sse2")
[...]

PR #323, opened 2019-08-26
enable cross compilation

bddap:
Fixes cross-compilation by addressing #322.

### DIFF
--- a/librocksdb-sys/build.rs
+++ b/librocksdb-sys/build.rs
@@ -168,7 +168,7 @@ fn build_rocksdb() {
 
 fn build_snappy() {
     let target = env::var("TARGET").unwrap();
-    
+
     let mut config = cc::Build::new();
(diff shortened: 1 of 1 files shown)


vitvakatu:
Thank you for a PR. Could you please fix trailing spaces on the added blank line to make `rustfmt` happy?

librocksdb-sys/build.rs
```diff
@@ -45,6 +45,8 @@ fn bindgen_rocksdb() {
 }
 
 fn build_rocksdb() {
+    let target = env::var("TARGET").unwrap();
```
aleksuss: If we don't set env variable "TARGET" we would get a panic. Is it really ok here ? What if get a target from `cfg!(target_arch)` by default, but if "TARGET" variable was set then get the value from it?

PR #328, opened 2019-09-07
Update rocksdb submodule to 6.1.2

vitvakatu:
_Actually_ upodates used rocksdb to 6.1.2


vitvakatu:
I spend some time updating list of rocksdb sources. It compiles and passes tests now, but I'm not sure if some files might be redundant. Also I splitted list in lines, because seeing differences between different versions was a complete nightmare. 

Btw, I used the following script to gather files (might be useful in future updates):
```bash
find . -type f -name '*.cc' | sed -e "s/^\.\///" | sed '/java/d' | sed '/win/d' | sed '/tools/d' | sed '/test\.cc/d'
```

It requires some manual editing after, but looks good in general.

aleksuss:
@vitvakatu did you use `make gen_lib_sources` command in `rust-rocksdb/librocskdb-sys` directory for generating the list of rocksdb sources?
By the way. The actual version of rocksdb is 6.2.2. What do you think about to update to this version ?

vitvakatu:
@aleksuss didn't see it 😄 

Right now there are some problems with compilation on Windows, so I'd make update to 6.2.2 in the future PR.

Btw are you sure we need to update to a versions that are not currently supported by any of the system package managers? Homebrew has rocksdb 6.1.2, on Ubuntu the situation is much worse.

PR #332, opened 2019-09-19
Add get_updates_since() wrapper and its iterator

nlfiedler:
In concert with the recently added latest_sequence_number(), this returns an
iterator for the write batches that have occurred since the given sequence.
This API from RocksDB is based on callbacks, and as a result the wrapper
relies on the application implementing a trait that acts as a callback.

cargo test passes

### DIFF
--- a/src/db.rs
+++ b/src/db.rs
@@ -162,6 +162,69 @@ pub struct DBRawIterator<'a> {
     db: PhantomData<&'a DB>,
 }
 
+/// Iterates the batches of writes since a given sequence number.
+///
+/// `DBWALIterator` is returned by `DB::get_updates_since()` and will return the
+/// batches of write operations that have occurred since a given sequence number
+/// (see `DB::latest_sequence_number()`). This iterator cannot be constructed by
+/// the application.
+///
+/// The iterator item type is a tuple of (`u64`, `WriteBatch`) where the first
+/// value is the sequence number of the associated write batch.
+///
+pub struct DBWALIterator {
+    inner: *mut ffi::rocksdb_wal_iterator_t,
+}
+
+impl DBWALIterator {
+    /// Returns `true` if the iterator is valid. An iterator is invalidated when
+    /// it reaches the end of its defined range, or when it encounters an error.
+    ///
+    /// To check whether the iterator encountered an error after `valid` has
+    /// returned `false`, use the [`status`](DBWALIterator::status) method.
+    /// `status` will never return an error when `valid` is `true`.
+    pub fn valid(&self) -> bool {
+        unsafe { ffi::rocksdb_wal_iter_valid(self.inner) != 0 }
+    }
+
+    /// Returns an error `Result` if the iterator has encountered an error
+    /// during operation. When an error is encountered, the iterator is
+    /// invalidated and [`valid`](DBWALIterator::valid) will return `false` when
+    /// called.
+    pub fn status(&self) -> Result<(), Error> {
+        unsafe {
+            ffi_try!(ffi::rocksdb_wal_iter_status(self.inner,));
+        }
+        Ok(())
+    }
+}
+
+impl Iterator for DBWALIterator {
+    type Item = (u64, WriteBatch);
+
+    fn next(&mut self) -> Option<(u64, WriteBatch)> {
+        // Seek to the next write batch.
+        unsafe {
+            ffi::rocksdb_wal_iter_next(self.inner);
+        }
+        if self.valid() {
+            let mut seq: u64 = 0;
+            let inner = unsafe { ffi::rocksdb_wal_iter_get_batch(self.inner, &mut seq) };
+            Some((seq, WriteBatch { inner }))
+        } else {
+            None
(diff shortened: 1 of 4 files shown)


nlfiedler:
The RocksDB API for iterating the write batch is based on callbacks, so the Rust wrapper is unlike proper Rust iterators. I'm happy to make any changes at all, such as rearranging the new code, rewriting the API docs, adding more tests, etc.

vitvakatu:
Travis CI failed because of formatting issues - please run `cargo +stable fmt`. You can do it after the review

tests/test_db.rs
```diff
@@ -251,3 +251,88 @@ fn test_sequence_number() {
...
+#[test]
+fn test_get_updates_since() {
+    let path = DBPath::new("_rust_rocksdb_test_get_updates_since");
+    {
+        let db = DB::open_default(&path).unwrap();
+        //
+        // get_updates_since() on an empty database
+        //
+        let result = db.get_updates_since(0);
+        assert!(result.is_ok());
```
vitvakatu: You can save some lines by `unwrap`ing result immediately - you don't need to check `is_ok` if you unwraping it right on the next line.

So `let iter = db.get_updates_since(0).unwrap();` is two lines less and much easier to read.

tests/test_db.rs
```diff
@@ -251,3 +251,88 @@ fn test_sequence_number() {
...
+        let mut iter = result.unwrap();
+        assert!(iter.next().is_none());
+
+        //
+        // add some records and collect sequence numbers,
+        // verify 3 batches of 1 put each were done
+        //
+        let _ = db.put(b"key1", b"value1");
```
vitvakatu: Here and later you're ignoring possible errors while performing `put` operation. The correct syntax would be `db.put(....).unwrap();`

vitvakatu: This also applies to other places where you have `let _ = `

tests/test_db.rs
```diff
@@ -251,3 +251,88 @@ fn test_sequence_number() {
...
+    fn put(&mut self, _key: Box<[u8]>, _value: Box<[u8]>) {
+        self.puts += 1;
+    }
+    fn delete(&mut self, _key: Box<[u8]>) {
+        self.deletes += 1;
+    }
+}
+
+#[test]
+fn test_get_updates_since() {
```
vitvakatu: nit: I'd split this test into 5 smaller tests, just for readability. Also I think you can move it to the separate file into `tests` directory - I'm literally dying because of thousands of lines in this crate.

src/db.rs
```diff
@@ -1588,6 +1651,66 @@ impl DB {
...
+}
+
+unsafe extern "C" fn writebatch_put_callback(
+    state: *mut c_void,
+    k: *const c_char,
+    klen: usize,
+    v: *const c_char,
+    vlen: usize,
+) {
+    // coerce the raw pointer back into a box, but "leak" it so we prevent
```
vitvakatu: I do not understand the purpose of leaking box here. Shouldn't it be available for the whole time of `rocksdb_writebatch_iterate` function?

src/db.rs
```diff
@@ -1588,6 +1651,66 @@ impl DB {
...
+        value.to_vec().into_boxed_slice(),
+    );
+}
+
+unsafe extern "C" fn writebatch_delete_callback(state: *mut c_void, k: *const c_char, klen: usize) {
+    // coerce the raw pointer back into a box, but "leak" it so we prevent
+    // freeing the resource before we are done with it
+    let boxed_cb = Box::from_raw(state as *mut &mut dyn WriteBatchIterator);
+    let leaked_cb = Box::leak(boxed_cb);
+    let key = slice::from_raw_parts(k as *const u8, klen as usize);
```
vitvakatu: Vec has [from_raw_parts](https://doc.rust-lang.org/std/vec/struct.Vec.html#method.from_raw_parts) method, so I think you can use it here to avoid convertation into slice.

tests/test_db.rs
nlfiedler: Yeah, I was copying other functions in that test crate, but I agree that splitting is good.

src/db.rs
nlfiedler: Because the boxed callback thing has to be passed through C and back again, we need the raw pointer. But when you call `Box::from_raw()` you get a _managed_ box. Now the callback thing would get freed. On the second call, we crash because of a double-free. To prevent that, we "leak" the box, telling the compiler to "forget" about it. The box is managed in the `WriteBatch.iterate()`, which is why I had the `Box::from_raw()` there, properly disposing of the box when we are completely done with it.

src/db.rs
nlfiedler: Actually, this is exactly the situation where calling `Vec::from_raw_parts()` is _not_ safe, the element types are `char` and the length is `size_t`. And indeed, it crashes when I try to use that function. Using the slice is the only safe way that I have found to do this.

aleksuss:
Well done 👍

Issue #149, opened 2017-11-07
Tailing iterator


andrenth:
Hi

Does rust-rocksdb support RocksDB's [Tailing iterator](https://github.com/facebook/rocksdb/wiki/Tailing-Iterator)?

Thanks.

PR #333, opened 2019-09-23
Add set_tailing to the ReadOptions struct to support tailing iterators.

cjbradfield:
Fixes: #149

### DIFF
--- a/src/db.rs
+++ b/src/db.rs
@@ -1998,6 +1998,15 @@ impl ReadOptions {
             ffi::rocksdb_readoptions_set_readahead_size(self.inner, v as size_t);
         }
     }
+
+    /// If true, create a tailing iterator. Note that tailing iterators
+    /// only support moving in the forward direction. Iterating in reverse
+    /// or seek_to_last are not supported.
+    pub fn set_tailing(&mut self, v: bool) {
+        unsafe {
+            ffi::rocksdb_readoptions_set_tailing(self.inner, v as c_uchar);
+        }
+    }
 }
 
 impl Default for ReadOptions {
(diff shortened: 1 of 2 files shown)


src/db.rs
```diff
@@ -2270,6 +2279,35 @@ fn iterator_test() {
...
+                str::from_utf8(&*k).unwrap(),
+                str::from_utf8(&*v).unwrap()
+            );
+            i = i + 1;
+        }
+        assert_eq!(i, 3);
+    }
+    let opts = Options::default();
+    assert!(DB::destroy(&opts, path).is_ok());
+}
```
aleksuss: I'd prefer to see something like this instead: 
```rust
#[test]
fn iterator_test_tailing() {
    let path = "_rust_rocksdb_iteratortest_tailing";
    {
        let data = [(b"k1", b"v1"), (b"k2", b"v2"), (b"k3", b"v3")];
        let mut ro = ReadOptions::default();
        ro.set_tailing(true);
        let db = DB::open_default(path).unwrap();

        let (key, value) = data[0];
        let p = db.put(key, value);
        assert!(p.is_ok());

        let iter = db.iterator_opt(IteratorMode::Start, &ro);

        let (key, value) = data[1];
        let p = db.put(key, value);
        assert!(p.is_ok());

        let (key, value) = data[2];
        let p = db.put(key, value);
        assert!(p.is_ok());

        for (i, (k, v)) in iter.enumerate() {
            let (real_key, real_value) = data[i];
            assert_eq!(
                (real_key.to_vec(), real_value.to_vec()),
                (k.to_vec(), v.to_vec())
            );
        }
    }
    assert!(DB::destroy(&opts, path).is_ok());
}
```

What do you thinks ?

cjbradfield: Sure, I can definitely do something like that. I had something similar at first, but then abandoned it to be like the iterator test above it. I'll update the PR this evening, PST time.

aleksuss:
@cjbradfield just one more thing. Could you please update a CHANGELOG file in this PR. Thanks.

cjbradfield:
> @cjbradfield just one more thing. Could you please update a `CHANGELOG.md` file in this PR. Thanks.

sure thing.

PR #342, opened 2019-10-22
Fix rocksdb build on mingw.

vldm:


librocksdb-sys/build.rs
```diff
@@ -129,6 +129,13 @@ fn build_rocksdb() {
         link("shlwapi", false);
         config.define("OS_WIN", Some("1"));
         config.define("ROCKSDB_WINDOWS_UTF8_FILENAMES", Some("1"));
+        if &target == "x86_64-pc-windows-gnu" {
+            // Tell Mingw to create localtime_r wrapper of localtime_s function.
```
vitvakatu:
```suggestion
            // Tell MinGW to create localtime_r wrapper of localtime_s function.
```

librocksdb-sys/build.rs
```diff
@@ -129,6 +129,13 @@ fn build_rocksdb() {
...
+            config.define("_POSIX_C_SOURCE", None);
+            // Tell Mingw to use atleast vista headers instead of winxp.
```
vitvakatu:
```suggestion
            // Tell MinGW to use at least Windows Vista headers instead of the ones of Windows XP.
```

vldm:
@aleksuss @vitvakatu CI for some reasons reports error. But i think this problem is not related to my patch. Could you try to restart job?

vitvakatu:
Restarted, Travis does weird things lately

PR #345, opened 2019-10-30
Always use pinnable slice for get operations

### LINKED ISSUE #336: DBVector and DBPinnableSlice's Send/Sync
I'm using rust-rocksdb in async environment so it involves potentially moving DBVector and DBPinnableSlice between multiple threads, and I noticed that both of them are automatically `!Send` and `!Sync` due to usage of raw pointer. This is overly restrictive.

Both of them are definitely `Sync` as they just points to a chunk of immutable memory.

DBVector is just a thin wrapper around a buffer allocated in C and is freed by `libc::free`, so it is indeed `Send`. I haven't concluded whether DBPinnableSlice is `Send` as I don't quite understand how it destructor interacts with multi-threading, but it probably is `Send` as well.

### LINKED ISSUE #337: A new approach for DB::get
I've digged through source code of rocksdb, and realize that if we use DB::get, then:
* Rust calls C API's rocksdb_get
* rocksdb_get create a C++ std::string, and calls C++ API's DB::Get
* DB::Get uses pinned api internally and copy it to the given string
* rocksdb_get copies the string to the malloc'ed slice

This looks incredibly inefficient to me. I think it'll be more efficient if we instead make `DBVector` a thin wrapper around `Box<[u8]>` (wrapper is necessary to not break API compatibility), and then implement `DB::get` using `DB::get_pinned` instead, and do the copying once in Rust rather than twice in C++.

nbdd0121:
Fixes #336 
Fixes #337
Related #294

As posted in #294, `get_pinned` has a huge performance advantage over `get` (current master)
```
test get_large                 ... bench:     153,760 ns/iter (+/- 5,751)
test get_pinned_large          ... bench:       3,147 ns/iter (+/- 203)
test get_pinned_small          ... bench:         141 ns/iter (+/- 5)
test get_pinned_to_owned_large ... bench:       6,804 ns/iter (+/- 485)
test get_pinned_to_owned_small ... bench:         157 ns/iter (+/- 5)
test get_small                 ... bench:         156 ns/iter (+/- 9)
```

There are almost no case where you don't want a pinned get. So to make API simpler, we should always use pinned get and name it `get`. (#337 analyzes the control flow of current get, which internally uses pinned get as well). This PR updates `get` APIs to use pinned get, and deprecate `get_pinned` APIs.

Technically it's a breaking change as it adds lifetime, but similar to #294, its effect should be fairly small. Users who want a `'static` struct could simply do `slice.as_ref().to_owned()`, and even doing such a clone is better than current `get` in terms of performance.

### DIFF
--- a/src/db.rs
+++ b/src/db.rs
@@ -1995,6 +1995,15 @@ pub struct DBPinnableSlice<'a> {
     db: PhantomData<&'a DB>,
 }
 
+// Safety note: auto-implementing Send on most db-related types is prevented by the inner FFI
+// pointer. In most cases, however, this pointer is Send-safe because it is never aliased and
+// rocksdb internally does not rely on thread-local information for its user-exposed types.
+unsafe impl<'a> Send for DBPinnableSlice<'a> {}
+
+// Sync is similarly safe for many types because they do not expose interior mutability, and their
+// use within the rocksdb library is generally behind a const reference
+unsafe impl<'a> Sync for DBPinnableSlice<'a> {}
+
 impl<'a> AsRef<[u8]> for DBPinnableSlice<'a> {
     fn as_ref(&self) -> &[u8] {
         // Implement this via Deref so as not to repeat ourselves
--- a/src/db.rs
+++ b/src/db.rs
@@ -949,7 +949,8 @@ impl DB {
         key: K,
         readopts: &ReadOptions,
     ) -> Result<Option<Vec<u8>>, Error> {
-        self.get_pinned_opt(key, readopts).map(|x| x.map(|v| v.as_ref().to_vec()))
+        self.get_pinned_opt(key, readopts)
+            .map(|x| x.map(|v| v.as_ref().to_vec()))
     }
 
     /// Return the bytes associated with a key value. If you only intend to use the vector returned
@@ -968,7 +969,8 @@ impl DB {
         key: K,
         readopts: &ReadOptions,
     ) -> Result<Option<Vec<u8>>, Error> {
-        self.get_pinned_cf_opt(cf, key, readopts).map(|x| x.map(|v| v.as_ref().to_vec()))
+        self.get_pinned_cf_opt(cf, key, readopts)
+            .map(|x| x.map(|v| v.as_ref().to_vec()))
     }
 
     /// Return the bytes associated with a key value and the given column family. If you only
--- a/src/db.rs
+++ b/src/db.rs
@@ -941,6 +941,9 @@ impl DB {
         self.write_opt(batch, &wo)
     }
 
+    /// Return the bytes associated with a key value with read options. If you only intend to use
+    /// the vector returned temporarily, consider using [`get_pinned_opt`](#method.get_pinned_opt)
+    /// to avoid unnecessary memory copy.
     pub fn get_opt<K: AsRef<[u8]>>(
         &self,
         key: K,
@@ -949,11 +952,16 @@ impl DB {
         self.get_pinned_opt(key, readopts).map(|x| x.map(|v| v.as_ref().to_vec()))
     }
 
-    /// Return the bytes associated with a key value
+    /// Return the bytes associated with a key value. If you only intend to use the vector returned
+    /// temporarily, consider using [`get_pinned`](#method.get_pinned) to avoid unnecessary memory
+    /// copy.
     pub fn get<K: AsRef<[u8]>>(&self, key: K) -> Result<Option<Vec<u8>>, Error> {
         self.get_opt(key.as_ref(), &ReadOptions::default())
     }
 
+    /// Return the bytes associated with a key value and the given column family with read options.
+    /// If you only intend to use the vector returned temporarily, consider using
+    /// [`get_pinned_cf_opt`](#method.get_pinned_cf_opt) to avoid unnecessary memory.
     pub fn get_cf_opt<K: AsRef<[u8]>>(
         &self,
         cf: &ColumnFamily,
@@ -963,6 +971,9 @@ impl DB {
         self.get_pinned_cf_opt(cf, key, readopts).map(|x| x.map(|v| v.as_ref().to_vec()))
     }
 
+    /// Return the bytes associated with a key value and the given column family. If you only
+    /// intend to use the vector returned temporarily, consider using
+    /// [`get_pinned_cf`](#method.get_pinned_cf) to avoid unnecessary memory.
     pub fn get_cf<K: AsRef<[u8]>>(
         &self,
         cf: &ColumnFamily,
--- a/src/db.rs
+++ b/src/db.rs
@@ -690,7 +690,7 @@ impl<'a> Snapshot<'a> {
         DBRawIterator::new_cf(self.db, cf_handle, &readopts)
     }
 
-    pub fn get<K: AsRef<[u8]>>(&self, key: K) -> Result<Option<DBVector>, Error> {
+    pub fn get<K: AsRef<[u8]>>(&self, key: K) -> Result<Option<Vec<u8>>, Error> {
         let readopts = ReadOptions::default();
         self.get_opt(key, readopts)
     }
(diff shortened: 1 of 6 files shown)


iSynaptic:
Just checked, we do already support getting pinned data:

https://github.com/rust-rocksdb/rust-rocksdb/commit/dbd2ca6e4f5bd4a268cddacace8a7d4132ee955d

I recommend we close without merging.

nbdd0121:
Rocksdb's get are more like legacy APIs. If you look into the C++ implementation, you will see it basically just uses its pinned API and does a copy (details in #337). There are absolutely no reasons to use the get API at all.

nbdd0121:
I disagree that we should provide a 1-to-1 mapping. 1-1 mapping should be for the -sys crate, not a crate that tries to do high-level abstraction and do it in the Rust way.

iSynaptic:
@nbdd0121 I understand the motivation for this PR - a desire to improve performance by not needlessly coping data around in memory.  We are in full agreement that we should provide ways to minimize/eliminate poor memory utilization.

re: 1-to-1 mapping: I view the `librocksdb-sys` create as a mechanical translation of the C-API into Rust. It's always going to be 1-to-1 with the C-API - it has to be.  However, the `rocksdb` crate I view as being a 1-to-1 mapping to the C++ API / conceptual OO model that the underlying RocksDB library implements.  Creating a more opinionated API should be left to yet another crate, built on top of either `librocksdb-sys` or `rocksdb`.

re: the inefficiencies of `get`: Thank you for the research done as part of #337.  I'd like to understand more about your findings.  I'm not sure by your description if you are saying there are 2 or 3 buffers being used when calling the normal `get` method.  I expected 2 - 3 would be a surprise.  I will dig into the underlying code as you did and share my findings.

That said, calling `get` means the data lives in a buffer that can live for a longer period of time because it's a copy. Calling `get_pinned` means the data lives in a buffer you should not hold onto for very long.  That distinction needs to hold true.  The way you've implemented your changes returns a `DBVector` that can hold on to pinned data with no indication to the caller that this is the case.  Additionally, pinned data needs to be [cleaned up differently](https://github.com/rust-rocksdb/rust-rocksdb/blob/master/src/db.rs#L2137-L2143), which DBVector [does not do](https://github.com/rust-rocksdb/rust-rocksdb/blob/master/src/db.rs#L2061-L2067).

If there are indeed 3 buffers in play when calling the normal `get` method, we may be able to get that down to 2 buffers by internally using `get_pinned`, but we would have to implement that differently than you have here (eg. create our own buffer and copy the data from the pinned buffer into it).

nbdd0121:
If you look at the C++ API instead of the C one, you will note that there is a single `Get` API, which can take either `PinnableSlice` or a `std::string`. The C API call the first variant `get_pinned` because you cannot overload a method in C. So If we want to provide a 1-1 mapping to the C++ API, rather than the C API, `get_pinned` shouldn't exist.

RocksDB's [change log](https://github.com/facebook/rocksdb/blob/master/HISTORY.md#public-api-change-22) mentions that the `Get` API with `std::string` is **discouraged**. It is still provided for backward-compatibility.

RocksDB (or rather all LSM databases) internally uses immutable data structure extensively. Therefore not destructing a PinnableSlice will not create a big burden to RocksDB itself. The only drawback is that the entire block may not be freed if there is an active PinnableSlice pointing to that block.

iSynaptic:
@nbdd0121 As you noted, the C++ API provides the an overload - a choice between a pinned buffer or a new buffer with data being copied into it.  Rust does not allow for method level overloads, so the current approach to expose this choice same is to provide to methods, `get_` and `get_pinned_`.

Regardless of how those are implemented in the Rust library (eg. calling the underlying `get` and `get_pinned` respectively, or both using pinned data is some way), `get_` needs to return a buffer that does not hold on to pinned data, whereas `get_pinned_` can.  We should not hide any trade-offs or drawbacks or make the choice for the library user - we should expose the choice and allow them to make the trade-off decisions.

nbdd0121:
> As you noted, the C++ API provides an overload - a choice between a pinned buffer or a new buffer with data being copied into it. Rust does not allow for method level overloads, so the current approach to expose this same choice is to provide two methods, `get_` and `get_pinned_`.

It does provide a overload, but the specific overload that we used for `get_` methods is **discouraged**. Being a binding that does not need to take backward compatibility burdens, we don't need to provide a binding to an deprecated API.

> Regardless of how those are implemented in the Rust library (eg. calling the underlying `get` and `get_pinned` respectively, or both using pinned data in some way), `get_` needs to return a buffer that does not hold on to pinned data, whereas `get_pinned_` can. We should not hide any trade-offs or drawbacks or make the choice for the library user - we should expose the choice and allow them to make the trade-off decisions.

By providing only one API, we are not hiding any trade-offs or make the choice for the library user. In contrast, we allow users of this library to make a decision about how to deal with these. If the user wants to hold the data for longer, the user can use `.as_ref().to_owned()` to create a clone. Probably we just need to add a note to the doc that if they plan to hold the data for a while, this is what they could do.

PS. using `.as_ref().to_owned()` will give the user `Vec<u8>`, which allows the user to mutate the data if it wants, so that's a bonus.
PPS. `get_pinned` isn't a good name, as it might be confused with `std::pin`.
PPPS. In most cases the return value of `get` is used only temporarily. So having `get` defaulted to the `PinnableSlice` is desirable. If we absolutely need to provide the deprecated non-pinned API to the user, it'd better have another name.

iSynaptic:
> It does provide a overload, but the specific overload that we used for get_ methods is discouraged.

I get that it is discouraged - because it is wasteful. However, it is not deprecated, it is still "supported".

 > If the user wants to hold the data for longer, the user can use .as_ref().to_owned() to create a clone.

This is just an alternative way of providing the choice: two methods or one method with clone operations on the return value. Historically, it's been two methods.  Having it be two methods makes the choice more explicit from the start, just like having two C++ method overloads makes the choice explicit from the start. Relying on a `.as_ref().to_owned()` approach defers the choice and can lead to accidentally using pinned data for long periods of time when copied data is what was needed.  

Given that the Rust library has had two methods for some time, and that most, if not all, language bindings expose two methods/overloads, switching to this alternative approach does not support the original motivation for this PR - performance increase by not copying memory needlessly. It argues for an alternative API to provide the exact same functionality.  Granted, it is trying to steer users towards a more performant approach, but adopts a _less_ safe default. By less safe, I mean now by default you need to know you are holding on to pinned data when you call `get` and you need to know that you need to call `.as_ref().to_owned()` if you plan on holding onto that data for long.  This does not encourage people to fall into ["the pit of success"](https://blog.codinghorror.com/falling-into-the-pit-of-success/).

@aleksuss @vitvakatu @spacejam Thoughts on this discussion?

aleksuss:
> We should expose our own get_pinned method.

I agree with that statement. I think a developer should have a possibility to choose what is more appropriate for him `get` or `get_pinned`.

vitvakatu:
I'm sorry for such a long response. I agree with @iSynaptic that we do not need to restrict user and create ambiguity in API by removing the old behavior of `get` method. 
So:
> Regardless of how those are implemented in the Rust library (eg. calling the underlying get and get_pinned respectively, or both using pinned data in some way), get_ needs to return a buffer that does not hold on to pinned data, whereas get_pinned_ can. We should not hide any trade-offs or drawbacks or make the choice for the library user - we should expose the choice and allow them to make the trade-off decisions.

nbdd0121:
Following the reviews I made the update to keep both API series around but let the `get` API uses `get_pinned` APIs internally. The return types of `get` APIs are changed from `DBVector` to `Vec<u8>`.

iSynaptic:
Thank you @nbdd0121. Are you able to provide new benchmark data with this new approach? I'd like to see how this performs and if it still contributes to a performance boost.

aleksuss:
By the way. Does it make sense to add such behaviour to snapshot implementation ?

iSynaptic:
> 
> 
> By the way. Does it make sense to add such behaviour to snapshot implementation ?

IMO, no. I plan on submitting a PR removing all "PIGMED" operations from `Snapshot`s since it doesn't match with the C++ API and had greatly complicated #268.

iSynaptic:
@nbdd0121 Thanks for the updated benchmark.

What I get from this new benchmark is that `get_small` has roughly the same performance, but `get_large` is much improved, going from the original 153μs to 6μs; quite the impressive boost.

Assuming all the CI build completes successfully, I'd like to see this get merged. It provides a great performance boost without changing the semantic difference between `get` and `get_pinned`.

Thanks for working through all of this @nbdd0121.

PR #346, opened 2019-10-30
Avoid copy in DBRawIterator::{key, value}

nbdd0121:
Thanks to Rust's lifetime and borrow checker, it's completely safe to access internal buffer without copying, as any seek operation will need to take `&mut self`.

This PR therefore removes the copying `key` and `value` functions, and replace them with the originally unsafe `key_inner` and `value_inner`.

### DIFF
--- a/src/db.rs
+++ b/src/db.rs
@@ -488,52 +488,38 @@ impl<'a> DBRawIterator<'a> {
         }
     }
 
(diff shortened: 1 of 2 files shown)


iSynaptic:
Thanks @nbdd0121!

PR #360, opened 2019-12-04
feat(rocksdb): upgrade rocksdb to v6.4.6

FlorentinDUBOIS:


vitvakatu:
You need to update `build_version.cc` file as well.

FlorentinDUBOIS:
Thanks, I will do that!

PR #372, opened 2020-01-10
Add flush_cf method to DB

wqfish:
The current implementation allows flushing the default column family, but not the rest. This change adds the ability to flush a given column family.

### DIFF
--- a/src/db.rs
+++ b/src/db.rs
@@ -911,7 +911,7 @@ impl DB {
         &self.path.as_path()
     }
 
-    /// Flush database memtable to SST files on disk (with options).
+    /// Flushes database memtables to SST files on the disk.
     pub fn flush_opt(&self, flushopts: &FlushOptions) -> Result<(), Error> {
         unsafe {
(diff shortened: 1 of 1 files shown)


src/db.rs
```diff
@@ -924,6 +924,19 @@ impl DB {
         self.flush_opt(&FlushOptions::default())
     }
 
+    /// Flush database memtable to SST files on disk for given column family.
```
```suggestion
    /// Flushes database memtables to SST files on the disk for a given column family.
```

src/db.rs
```diff
@@ -924,6 +924,19 @@ impl DB {
...
+    pub fn flush_cf_opt(&self, cf: &ColumnFamily, flushopts: &FlushOptions) -> Result<(), Error> {
+        unsafe {
+            ffi_try!(ffi::rocksdb_flush_cf(self.inner, flushopts.inner, cf.inner));
+        }
+        Ok(())
+    }
+
+    /// Flush database memtable to SST files on disk for given column family using default option.
```
```suggestion
    /// Flushes database memtables to SST files on the disk for a given column family using default options.

Issue #224, opened 2018-11-14
Transition to Rust 2018


iSynaptic:
Rust 2018 is [slated to land in December 2018](https://blog.rust-lang.org/2018/07/27/what-is-rust-2018.html). Perhaps we should transition to it?  Thoughts?

vitvakatu:
I propose to wait just a bit, to avoid some unpredictable bugs and regressions and make sure that the whole ecosystem will make a transition together with us. (one of the examples - `failure` wasn't ready to `dyn Trait` syntax and caused a lot of problems in various projects which depend on it)

iSynaptic:
@vitvakatu Sounds good. I'll leave this open as a reminder to reconsider after some time has passed.

iSynaptic:
@vitvakatu @aleksuss There have been 2 releases since Rust 2018 landed in December. Any objections to moving over to it?

vitvakatu:
No objections from me,  Rust 2018 looks stable enough now

PR #375, opened 2020-01-13
Move to Rust 2018

wqfish:
It has been a while since Rust 2018 released and we should be able to do it now :)

Fixes https://github.com/rust-rocksdb/rust-rocksdb/issues/224


tests/test_compationfilter.rs
```diff
@@ -32,7 +30,7 @@ fn test_filter(level: u32, key: &[u8], value: &[u8]) -> CompactionDecision {
 
 #[test]
 fn compaction_filter_test() {
-    use {Options, DB};
+    use crate::{Options, DB};
```
aleksuss: It's a redundant use and could be removed.

Issue #352, opened 2019-11-16
batch.delete doesn't return error when key not exist and thus batch isn't atomic


ailisp:
In this snippet:
```rust
use rocksdb::{Options, WriteBatch, DB};

fn main() {
    // NB: db is automatically closed at end of lifetime
    let path = "bo_test_rocksdb.db";
    {
        let db = DB::open_default(path).unwrap();
        db.put(b"a", b"b").unwrap();

        let mut batch = WriteBatch::default();
        batch.put(b"a", b"bb").unwrap();
        batch.delete(b"c").unwrap();
        db.write(batch).unwrap();
        match db.get(b"a") {
            Ok(Some(value)) => assert_eq!(value, b"b"),
            _ => panic!(),
        }
        db.delete(b"a").unwrap();
    }
    let _ = DB::destroy(&Options::default(), path);
}
```
I expect `batch.delete(b"c")` returns an error, as its doc saying:
```rust
impl WriteBatch {
//...
    /// Remove the database entry for key.
    ///
    /// Returns an error if the key was not found.
    pub fn delete<K: AsRef<[u8]>>(&mut self, key: K) -> Result<(), Error> {
```
But that unwrap doesn't panic.

I then expect the `db.get(b"a")` will get `Some(b"b")` because `WriteBatch` is documented as
```rust
/// An atomic batch of write operations.
///
/// Making an atomic commit of several writes:
// ...
///     db.write(batch); // Atomically commits the batch
// ...
pub struct WriteBatch {
```

But it panics at `assert_eq!`, saying value is `b"bb"`, this indicates the `put` in batch committed to db but delete fail just silently ignored. Expected result: the whole result isn't committed to db.

Also as a separate question, is there other cases I can test to confirm the batch is atomic? Thanks!

wqfish:
This behavior is expected -- see https://github.com/facebook/rocksdb/blob/1dd7873e08c1f55dabb8451f1b8795fcdb012887/include/rocksdb/write_batch.h#L85

The doc needs to be fixed.

PR #376, opened 2020-01-15
Fix doc for WriteBatch::delete

wqfish:
The function does nothing if the key is not found in DB. See https://github.com/facebook/rocksdb/blob/1dd7873e08c1f55dabb8451f1b8795fcdb012887/include/rocksdb/write_batch.h#L85

Fixes https://github.com/rust-rocksdb/rust-rocksdb/issues/352

Also note that these `WriteBatch` functions never return `Err`, so maybe we should change all of them to not return `Result`?

### DIFF
--- a/src/db.rs
+++ b/src/db.rs
@@ -1765,9 +1765,7 @@ impl WriteBatch {
         }
     }
 
-    /// Remove the database entry for key.
-    ///
-    /// Returns an error if the key was not found.
+    /// Removes the database entry for key. Does nothing if the key was not found.
     pub fn delete<K: AsRef<[u8]>>(&mut self, key: K) -> Result<(), Error> {
         let key = key.as_ref();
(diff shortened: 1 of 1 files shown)


src/db.rs
```diff
@@ -1766,7 +1766,7 @@ impl WriteBatch {
 
     /// Remove the database entry for key.
```
```suggestion
    /// Removes the database entry for the key.
```

aleksuss:
@wqfish Yes. I agree that we should change a signature of the `WriteBatch::delete` method. A `Result<(), Error>` is redundant, I suppose.

wqfish:
> @wqfish Yes. I agree that we should change a signature of the `WriteBatch::delete` method. A `Result<(), Error>` is redundant, I suppose.

I'll send a separate PR for that.

PR #387, opened 2020-02-17
Bump uuid and bindgen dependencies

### LINKED ISSUE #384: Old version of bindgen brings in old dependencies
The newest version of `bindgen` is `0.52.0`. It'd be good to update to it so that various transitive dependencies also get updated.

jonhoo:
Fixes #384.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -26,4 +26,4 @@ bzip2 = ["librocksdb-sys/bzip2"]
 
 [dependencies]
 libc = "0.2"
-librocksdb-sys = { path = "librocksdb-sys", version = "6.4.6" }
+librocksdb-sys = { path = "librocksdb-sys", version = "6.4.7" }
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -26,4 +26,4 @@ bzip2 = ["librocksdb-sys/bzip2"]
 
 [dependencies]
 libc = "0.2"
-librocksdb-sys = { path = "librocksdb-sys", version = "6.4.7" }
+librocksdb-sys = { path = "librocksdb-sys", version = "6.4.6" }
--- a/librocksdb-sys/Cargo.toml
+++ b/librocksdb-sys/Cargo.toml
@@ -1,6 +1,6 @@
 [package]
 name = "librocksdb-sys"
-version = "6.4.6"
+version = "6.4.7"
 edition = "2018"
 authors = ["Karl Hobley <karlhobley10@gmail.com>", "Arkadiy Paronyan <arkadiy@ethcore.io>"]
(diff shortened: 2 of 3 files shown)


jonhoo:
Also note that irrespective of this PR, I think a crates.io release should be made to `librocksdb-sys`, since the latest version on crates.io is 6.2.4.

jonhoo:
If the version needs to stay `6.4.6`, I would recommend merging this and _then_ publishing 6.4.6. It's not entirely clear whether this change is backwards-compatible (I don't know if the generated code is "the same"), but if the versioning of the crate is tied to the versioning of the rocksdb lib, we may not have much of a choice.

jonhoo:
Haha, no, sadly not — this is just a part of my everyday workday :sweat_smile:

jonhoo:
Interesting, these build failures look mostly unrelated to this change, but I'll dig a little: https://travis-ci.org/rust-rocksdb/rust-rocksdb/jobs/651573400?utm_medium=notification&utm_source=github_status

jonhoo:
Yeah, the same build failure occurs on `master` on Windows:
https://travis-ci.org/rust-rocksdb/rust-rocksdb/branches

aleksuss:
@jonhoo never mind. It's known issue on windows host. Everything is ok.

jonhoo:
Seems to pass on the other two platforms at least :tada:

jonhoo:
@aleksuss Nice! What's the next steps for getting out a release of `librocksdb-sys`?

aleksuss:
@jonhoo Usually, I publish a new `librocksdb-sys` release together with the main crate. To be honest, I haven't decided when to publish a new release yet.

jonhoo:
Is there a particular reason you don't want to release a new minor release of `librocksdb-sys` without also doing a new release of `rocksdb`? Presumably the latter has more breaking changes and such than the former, and the releases are mostly independent?

jonhoo:
Alternatively, would you consider doing a patch release of `rocksdb` (so `0.13.1`) that includes the bump to `librocksdb-sys`?

aleksuss:
Yeah. I will think about it. I suppose it's a good idea.

jonhoo:
I expect it will cut compile times for certain crates pretty drastically, as they no longer need to build multiple versions of `syn`, `quote`, and `bindgen`.

jonhoo:
@aleksuss have you had a chance to think some more about whether to release new versions of `librocksdb-sys`?

aleksuss:
Done https://crates.io/crates/librocksdb-sys I hope it would be enough for you ? 🙂

jonhoo:
Beautiful, thank you!

Issue #386, opened 2020-02-08
Lifetime handling on iterators appears incorrect, leads to crashes and is likely undefined behavior


shepmaster:
You are able to return an iterator that has zero relation to the database it comes from:

```rust
use rocksdb::{DBIterator, IteratorMode, DB};

fn example() -> DBIterator<'static> {
    let db = DB::open_default("storage").unwrap();
    db.iterator(IteratorMode::Start)
}

fn main() {
    for i in example() {
        println!("{:?}", i);
    }
}
```

When running this purely-safe Rust code, an error occurs and the program crashes:

```
% cargo run
    Finished dev [unoptimized + debuginfo] target(s) in 0.09s
     Running `target/debug/repro`
pthread lock: Invalid argument
```

valgrind has this to say:

```
==3415== Warning: unimplemented fcntl command: 1036
==3415== Invalid read of size 4
==3415==    at 0x41824F: fetch_sub (atomic_base.h:524)
==3415==    by 0x41824F: rocksdb::SuperVersion::Unref() (column_family.cc:361)
==3415==    by 0x16F506: rocksdb::(anonymous namespace)::CleanupIteratorState(void*, void*) (db_impl.cc:1262)
==3415==    by 0x2127D8: rocksdb::Cleanable::DoCleanup() (cleanable.h:68)
==3415==    by 0x38CC7B: rocksdb::Cleanable::~Cleanable() (iterator.cc:22)
==3415==    by 0x23BC27: rocksdb::InternalIteratorBase<rocksdb::Slice>::~InternalIteratorBase() (internal_iterator.h:24)
==3415==    by 0x25DDD3: rocksdb::MemTableIterator::~MemTableIterator() (memtable.cc:295)
==3415==    by 0x21B4FF: rocksdb::IteratorWrapperBase<rocksdb::Slice>::DeleteIter(bool) (iterator_wrapper.h:52)
==3415==    by 0x214523: rocksdb::DBIter::~DBIter() (db_iter.cc:165)
==3415==    by 0x21A219: rocksdb::ArenaWrappedDBIter::~ArenaWrappedDBIter() (db_iter.cc:1483)
==3415==    by 0x21A263: rocksdb::ArenaWrappedDBIter::~ArenaWrappedDBIter() (db_iter.cc:1483)
==3415==    by 0x15046B: rocksdb_iter_destroy (c.cc:1234)
[...]

Issue #356, opened 2019-11-25
Lacking lifetime on iterator_cf(_opt)


ailisp:
DB.iterator_cf and DB.iterator_cf_opt lacks a similar lifetime definition like iterator_opt/iterator_opt:
```
    pub fn iterator<'a, 'b: 'a>(&'a self, mode: IteratorMode) -> DBIterator<'b> {
```
Lack of this make you unable to write a function that takes a &DB and return a DBIterator

nearmax:
Actually, I think it does not make sense for `'b` to have a larger scope than `'a` and `iterator_cg` and `iterator_cf_opt` should be redone. The iterator should not outlive the object it is iterating over (the database). What am I missing?

PR #391, opened 2020-02-18
Fix lifetime parameter for iterators

wqfish:
The current implementation says that the lifetime of the iterator (parameter `'b`) lives at least as long as the lifetime of the DB (parameter `'a`), which is the opposite of what we want. Make it `'a: 'b` instead.

Also add lifetime parameters to a few extra iterator related functions.

Fixes https://github.com/rust-rocksdb/rust-rocksdb/issues/386
Fixes https://github.com/rust-rocksdb/rust-rocksdb/issues/356

### DIFF
--- a/src/db.rs
+++ b/src/db.rs
@@ -1106,12 +1106,12 @@ impl DB {
         self.cfs.get(name)
     }
 
-    pub fn iterator<'a, 'b: 'a>(&'a self, mode: IteratorMode) -> DBIterator<'b> {
+    pub fn iterator<'a: 'b, 'b>(&'a self, mode: IteratorMode) -> DBIterator<'b> {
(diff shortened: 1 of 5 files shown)


src/db.rs
```diff
@@ -1176,28 +1184,28 @@ impl DB {
     }
 
     /// Opens a raw iterator over the database, using the default read options
-    pub fn raw_iterator(&self) -> DBRawIterator {
+    pub fn raw_iterator<'a: 'b, 'b>(&self) -> DBRawIterator<'b> {
```
vitvakatu: Probably a typo here, `'a` is unused.

wqfish: good catch

PR #393, opened 2020-03-06
Fix building librocksdb-sys with system libraries

basvandijk:
This PR:

 * Uses the correct names for the zlib and bzip2 libraries. The zlib library is called `libz.a` and the bzip2 library is called `libbz2.a`.

* Only fails on empty sub-module directories when the librocksdb-sys is configured to build them.

* Parameterize the rocksdb include directory.

### DIFF
--- a/librocksdb-sys/build.rs
+++ b/librocksdb-sys/build.rs
@@ -26,7 +26,7 @@ fn fail_on_empty_directory(name: &str) {
     }
 }
 
-fn rocksdb_include_dir() -> std::string::String {
+fn rocksdb_include_dir() -> String {
     match env::var("ROCKSDB_INCLUDE_DIR") {
         Ok(val) => return val,
         Err(_) => return "rocksdb/include".to_string(),
--- a/librocksdb-sys/build.rs
+++ b/librocksdb-sys/build.rs
@@ -323,10 +323,10 @@ fn main() {
     if cfg!(feature = "zstd") && !try_to_find_and_link_lib("ZSTD") {
         build_zstd();
     }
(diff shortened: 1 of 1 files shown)


librocksdb-sys/build.rs
```diff
@@ -26,9 +26,16 @@ fn fail_on_empty_directory(name: &str) {
     }
 }
 
+fn rocksdb_include_dir() -> std::string::String {
```
vitvakatu: Why is it needed? Do you have a specific use-case.

basvandijk: I would like to build the `rocksdb` crate with a system-provided `rocksdb` library. So in my build recipe I'm setting:

`ROCKSDB_INCLUDE_DIR = "${rocksdb}/include";`

where `rocksdb` is the path to the `rocksdb` on my system.

aleksuss: By the way. A system-provided library places its headers with paths like `/usr/include` in linux case or `/usr/local/include` in macos case. A compiler uses these paths by default AFAIK 🤔 This var could be useful when you use sources from other different places.

basvandijk: Indeed. In my case I set:
```
ROCKSDB_INCLUDE_DIR=/nix/store/m9xx9jq9b6vz42zy8xdzj03grskdlm89-rocksdb-6.6.4/include
```
since I use the Nix build system.

aleksuss: Yes. I think it could be useful in some cases.

PR #250, opened 2019-02-01
add transaction support

hjiayz:

PR #394, opened 2020-03-08
Make DBPath use tempfile.

jder:
This ensures no collisions when running tests in a multithreaded manner,
and avoids cluttering up the repo's directory with temp files.

Step one in allowing `cargo watch` to work correctly; more cleanup is
needed before we stop making any files in the repo's directory.

This also helps to prevent bugs where DBPath's drop() could run early
because the DBPath itself (not a reference) was passed into something
like DB::open and dropped immediately, destroying the database.

---

I'm planning to try and get functionality like #250 merged, starting with 
things that are easy to peel off like this. (This is a slightly different approach
 from #250 which I hope seems reasonable.)

Easy follow-ups I'm doing in https://github.com/jder/rust-rocksdb/tree/db-path-temp-full
(could be this PR if preferred; not sure how much people are OK with in one PR):
  - [x] delete tests from db.rs which are duplicated in tests/test_db.rs
  - [x] move remaining tests over to use DBPath where possible, tempfile where not
  - [x] remove `_rust_rocksdb_` from all the DBPath names (and from .gitignore)
  - [x] rename DBPath to TemporaryDBPath as in #250. 
  - [x] remove path argument from TemporaryDBPath as in #250

### DIFF
--- a/tests/test_iterator.rs
+++ b/tests/test_iterator.rs
@@ -253,7 +253,7 @@ fn test_prefix_iterator_uses_full_prefix() {
 
 #[test]
 fn test_full_iterator() {
-    let path = "_rust_rocksdb_fulliteratortest";
+    let path = DBPath::new("fulliteratortest");
     {
         let a1: Box<[u8]> = key(b"aaa1");
         let a2: Box<[u8]> = key(b"aaa2");
@@ -273,7 +273,7 @@ fn test_full_iterator() {
         opts.set_allow_concurrent_memtable_write(false);
         opts.set_memtable_factory(factory);
 
-        let db = DB::open(&opts, path).unwrap();
+        let db = DB::open(&opts, &path).unwrap();
 
         assert!(db.put(&*a1, &*a1).is_ok());
         assert!(db.put(&*a2, &*a2).is_ok());
@@ -295,8 +295,6 @@ fn test_full_iterator() {
         let a_iterator = db.full_iterator(IteratorMode::Start);
         assert_eq!(a_iterator.collect::<Vec<_>>(), expected)
     }
-    let opts = Options::default();
-    assert!(DB::destroy(&opts, path).is_ok());
 }
 
 fn custom_iter<'a>(db: &'a DB) -> impl Iterator<Item = usize> + 'a {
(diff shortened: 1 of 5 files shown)


tests/test_iterator.rs
```diff
@@ -295,8 +295,6 @@ fn test_full_iterator() {
         let a_iterator = db.full_iterator(IteratorMode::Start);
         assert_eq!(a_iterator.collect::<Vec<_>>(), expected)
     }
-    let opts = Options::default();
-    assert!(DB::destroy(&opts, path).is_ok());
```
jder: These are unnecessary since it happens in DBPath::drop.

tests/util/mod.rs
```diff
@@ -1,39 +1,40 @@
...
+        DB::destroy(&opts, &self.path).expect("Failed to destroy temporary DB");
     }
 }
 
-impl AsRef<Path> for DBPath {
+/// Convert a DBPath ref to a Path ref.
+/// We don't implement this for DBPath values because we want them to
+/// exist until the end of their scope, not get passed in to functions and
+/// dropped early.
+impl AsRef<Path> for &DBPath {
```
jder: Note the change here to `&DBPath` to encourage passing a reference around to e.g. DB::open.

tests/util/mod.rs
```diff
@@ -1,39 +1,40 @@
...
-            "{}.{}.{}",
-            prefix,
-            current_time.as_secs(),
-            current_time.subsec_nanos()
-        );
+        let dir = tempfile::Builder::new()
+            .prefix(prefix)
+            .tempdir()
+            .expect("Failed to create temporary path for db.");
+        let path = dir.path().join("db");
```
aleksuss: Is `.join("db")` really necessary ?

jder: Most of the tests are OK with opening `dir.path()` directly, but the checkpoint tests expect there to be no directory at this path before they run. So, I could remove this and then not move those tests over to use this mechanism, but it seems harmless to me.

jder:
@aleksuss thanks for the review. Can we merge this or should I include some of the other items I listed in the description first?

aleksuss:
@jder I suppose we can. Let's wait for @vitvakatu review.

jder:
@vitvakatu would you be able to take a look at this PR?

PR #395, opened 2020-03-14
Refactor db.rs and lib.rs into smaller pieces

jder:
As part of trying to get the changes from #250/#268 merged, this PR includes just the mechanical code movement from #268 of things from db.rs and lib.rs into separate files/modules. I believe this leaves the public interface entirely unchanged. (None of the tests needed edits.) 

I suggest reviewing each commit individually. They each:
* move some related types out of db.rs and/or lib.rs to match #268
* update imports in lib.rs and db.rs
* make some things `pub(crate)` as needed

Cases where I'm aware of differences from #268 are:
* This doesn't add the `prelude` module (could be a follow-up).
* I didn't make the `column_family` module public since none of the others are.
* I didn't make `ffi_util::to_cpath` public (to avoid changing the public interface).

I also chose not to move the unit tests out of db.rs along with the related types (I'd like to merge #394 first to reduce the number of conflicts). 

cc @aleksuss @iSynaptic @vitvakatu if you're interested in reviewing. Thanks!


jder:
Thanks for the approval @vitvakatu! Would we be able to merge this now or are you looking for another review?

PR #396, opened 2020-03-24
check for endianness when building snappy

knarz:
Hey,

snappy's build process usually involves a configuration step which detects whether it is build on a big or little endian system. Since the `librocksdb-sys` uses a `build.rs` that step is skipped and snappy ends up being built for little endian even if on a big endian system. This small patch should change this behaviour.

### DIFF
--- a/librocksdb-sys/build.rs
+++ b/librocksdb-sys/build.rs
@@ -174,6 +174,7 @@ fn build_rocksdb() {
 
 fn build_snappy() {
     let target = env::var("TARGET").unwrap();
+    let endianness = env::var("CARGO_CFG_TARGET_ENDIAN").unwrap();
 
     let mut config = cc::Build::new();
     config.include("snappy/");
@@ -185,6 +186,10 @@ fn build_snappy() {
         config.flag("-EHsc");
     } else {
         config.flag("-std=c++11");
+    }
+
+    if endianness == "big" {
+        config.define("WORDS_BIGENDIAN", Some("1"));
     }
 
     config.file("snappy/snappy.cc");


librocksdb-sys/build.rs
```diff
@@ -187,6 +188,10 @@ fn build_snappy() {
         config.flag("-std=c++11");
     }
 
+    if endianness == "big" {
+        config.define("WORDS_BIGENDIAN", Some("1"));
```
aleksuss: I couldn't find a usage of the `WORDS_BIGENDIAN` var. Is it really correct name ?

dvdplm: I think you're right, good catch. @knarz should it be `SNAPPY_IS_BIG_ENDIAN`?

knarz: The snappy version pulled in here is over two years old and has `#ifdef WORDS_BIGENDIAN` as the only only reference to endianness.
Your suggestion seems to be correct for a current version of snappy

aleksuss: Yes. In a revision we use really presents `WORDS_BIGENDIAN`. So. Maybe update snappy for fresher revision like [this](https://github.com/google/snappy/commit/537f4ad6240e586970fe554614542e9717df7902) for using an actual var `SNAPPY_IS_BIG_ENDIAN` ?

aleksuss: It's really up to you. Actually, I can merge it as is, but I think to update snappy revision in any case even in other PR.

knarz: I'd say merge this as is and update snappy later.

Updating snappy would definitely be in order, seeing as the version used is very old. Unfortunately I don't have the time to work on that at the moment—I assume it's a bigger task.

PR #399, opened 2020-03-29
Add some extra config options

casualjim:
adds options that I have in use in a go version of the current library I'm working on: 

* set_pipelined_write,
* memtable_huge_page_size 
* optimize_universal_style_compaction

### DIFF
--- a/src/db_options.rs
+++ b/src/db_options.rs
@@ -186,7 +186,7 @@ impl Drop for ReadOptions {
 }
 
 impl BlockBasedOptions {
-    /// Approximate size of user data packed per block.  Note that the
+    /// Approximate size of user data packed per block. Note that the
     /// block size specified here corresponds to uncompressed data. The
     /// actual size of the unit read from disk may be smaller if
     /// compression is enabled. This parameter can be changed dynamically.
--- a/src/db_options.rs
+++ b/src/db_options.rs
@@ -187,9 +187,9 @@ impl Drop for ReadOptions {
 
 impl BlockBasedOptions {
     /// Approximate size of user data packed per block.  Note that the
-    /// block size specified here corresponds to uncompressed data.  The
+    /// block size specified here corresponds to uncompressed data. The
     /// actual size of the unit read from disk may be smaller if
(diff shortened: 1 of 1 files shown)


src/db_options.rs
```diff
@@ -192,6 +192,12 @@ impl BlockBasedOptions {
         }
     }
 
+    pub fn set_metadata_block_size(&mut self, size: usize) {
```
aleksuss: Could you add some description for the public method ?

src/db_options.rs
```diff
@@ -201,6 +207,15 @@ impl BlockBasedOptions {
         }
     }
 
+    pub fn set_lru_cache_compressed(&mut self, size: size_t) {
```
aleksuss: Here as well.

casualjim:
I took care of this now

src/db_options.rs
```diff
@@ -186,12 +186,41 @@ impl Drop for ReadOptions {
 }
 
 impl BlockBasedOptions {
+    /// Approximate size of user data packed per block.  Note that the
```
aleksuss: Sorry for nagging but decrease amount of spaces between sentences please.

src/db_options.rs
```diff
@@ -201,6 +231,19 @@ impl BlockBasedOptions {
         }
     }
 
+    /// When configured: use the specified cache for compressed blocks.
+    /// Otherwise rocksdb will not use a compressed block cache.
+    /// Note: though it looks similar to `block_cache`, RocksDB doesn't put the
+    ///       same type of object there.
```
aleksuss:
```suggestion
    /// same type of object there.
```

src/db_options.rs
```diff
@@ -186,12 +186,42 @@ impl Drop for ReadOptions {
 }
 
 impl BlockBasedOptions {
+    /// Approximate size of user data packed per block.  Note that the
```
aleksuss:
```suggestion
    /// Approximate size of user data packed per block. Note that the
```

PR #402, opened 2020-04-09
Add support for open_for_read_only APIs

wqfish:
This change exposes the C APIs that open RocksDB for read only.

### DIFF
--- a/src/db.rs
+++ b/src/db.rs
@@ -50,26 +50,26 @@ unsafe impl Send for DB {}
 // use within the rocksdb library is generally behind a const reference
 unsafe impl Sync for DB {}
 
-// Option that specifies whether open DB for read only.
-enum RWOption {
+// Specifies whether open DB for read only.
+enum AccessType {
     ReadWrite,
     ReadOnly { error_if_log_file_exist: bool },
 }
 
 impl DB {
-    /// Open a database with default options.
+    /// Opens a database with default options.
     pub fn open_default<P: AsRef<Path>>(path: P) -> Result<DB, Error> {
         let mut opts = Options::default();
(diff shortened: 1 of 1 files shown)


src/db.rs
```diff
@@ -76,11 +91,51 @@ impl DB {
...
+            opts,
+            path,
+            cfs,
+            RWOption::ReadOnly {
+                error_if_log_file_exist,
+            },
+        )
     }
 
     /// Open a database with the given database options and column family descriptors.
```
aleksuss, line 123:
```suggestion
    /// Opens a database with the given database options and column family descriptors.
```

src/db.rs
```diff
@@ -50,6 +50,12 @@ unsafe impl Send for DB {}
 // use within the rocksdb library is generally behind a const reference
 unsafe impl Sync for DB {}
 
+// Option that specifies whether open DB for read only.
+enum RWOption {
```
aleksuss, line 54: Does it make sense to add derives Clone and Copy here ?

aleksuss: And I'd rename it to `AccessType`.

wqfish: > Does it make sense to add derives Clone and Copy here ?

Probably not necessary right now. It's not a public type so we can always add later if needed. I'll rename it.

calavera:
@aleksuss I have implemented `OpenAsSecondary` on top of the AccessType enum introduced in this pull request. I was wonder when you're going to merge it, so I can open a new pull request without complicating the review more than necessary.

aleksuss:
@calavera I'm waiting for your PR. I think to include it in the 0.14.0 release.

PR #423, opened 2020-05-03
Allow to build RocksDB with a different stdlib.

calavera:
RocksDB supports up to C++20, so this should be configurable.

I've left the default value as it currently is.

Signed-off-by: David Calavera <david.calavera@gmail.com>

### DIFF
--- a/librocksdb-sys/build.rs
+++ b/librocksdb-sys/build.rs
@@ -312,16 +312,13 @@ fn try_to_find_and_link_lib(lib_name: &str) -> bool {
 }
 
 fn cxx_standard() -> String {
-    env::var("ROCKSDB_CXX_STDLIB").map_or_else(
-        |_| "-std=c++11".to_owned(),
-        |lib| {
-            if !lib.starts_with("-std=") {
-                format!("-std={}", lib)
-            } else {
-                lib
-            }
-        },
-    )
+    env::var("ROCKSDB_CXX_STD").map_or("-std=c++11".to_owned(), |cxx_std| {
+        if !cxx_std.starts_with("-std=") {
+            format!("-std={}", cxx_std)
+        } else {
+            cxx_std
+        }
+    })
 }
 
 fn main() {
--- a/librocksdb-sys/build.rs
+++ b/librocksdb-sys/build.rs
@@ -163,7 +163,7 @@ fn build_rocksdb() {
     if target.contains("msvc") {
         config.flag("-EHsc");
     } else {
-        config.flag(&stdlib_flag());
+        config.flag(&cxx_standard());
         // this was breaking the build on travis due to
         // > 4mb of warnings emitted.
         config.flag("-Wno-unused-parameter");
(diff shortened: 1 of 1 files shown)


librocksdb-sys/build.rs
```diff
@@ -309,6 +311,15 @@ fn try_to_find_and_link_lib(lib_name: &str) -> bool {
     false
 }
 
+fn stdlib_flag() -> String {
+    let lib = env::var("ROCKSDB_CXX_STDLIB").unwrap_or_else(|_| "-std=c++11".to_owned());
+    if !lib.starts_with("-std=") {
```
aleksuss, line 316: What about to use a regex validation here ?

librocksdb-sys/build.rs
```diff
@@ -309,6 +311,15 @@ fn try_to_find_and_link_lib(lib_name: &str) -> bool {
     false
 }
 
+fn stdlib_flag() -> String {
```
aleksuss, line 314: I'd name this function `cxx_standart`.

librocksdb-sys/build.rs
calavera: what should it validate?

librocksdb-sys/build.rs
methyl: I guess it should be `cxx_standard`

librocksdb-sys/build.rs
aleksuss: Validate that value is something like `-std=c++14` or `-std=gnu++11`. If value is different than return `-std=c++11`. Because if user makes an error than he would get compile error. So reason is to prevent such cases.

calavera: I'm thinking about this, and I wonder if it's not better to let the build fail if the flag is not valid. 

If you change this value, and you made a typo, and we use the default c++11, you'd be building with a std version that you don't want, which might be more unexpected, and you might not even notice.

If you change this value, and the build fails fast, you can identify the problem right away.

librocksdb-sys/build.rs
```diff
@@ -309,6 +311,19 @@ fn try_to_find_and_link_lib(lib_name: &str) -> bool {
     false
 }
 
+fn cxx_standard() -> String {
+    env::var("ROCKSDB_CXX_STDLIB").map_or_else(
```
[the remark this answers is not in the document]
aleksuss: Is it clippy suggestion to use `map_or_else` ? It seems `map_or` is more appropriate.
```rust
fn cxx_standard() -> String {
    env::var("ROCKSDB_CXX_STD").map_or("-std=c++11".to_owned(), |cxx_std| {
        if !cxx_std.starts_with("-std=") {
            format!("-std={}", cxx_std)
        } else {
            cxx_std
        }
    })
}
```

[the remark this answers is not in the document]
calavera: sounds great, thanks for the suggestion.

PR #425, opened 2020-05-06
Expose `open_with_ttl`.

calavera:
This api allows you to open a database with a TTL filter. 
This filter is applied during compaction, and entries get dropped if their TTL has expired.

This is based on the current RocksDB C bindings that don't expose additional functionality from the ttl utilities.

Signed-off-by: David Calavera <david.calavera@gmail.com>

### DIFF
--- a/src/db.rs
+++ b/src/db.rs
@@ -31,6 +31,7 @@ use std::path::PathBuf;
 use std::ptr;
 use std::slice;
 use std::str;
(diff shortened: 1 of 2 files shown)


mraerino:
is there a schedule for releasing this?

mraerino:
@aleksuss could you cut a release, so we can use this properly?

PR #431, opened 2020-05-24
Add operations as traits

acrrd:
This PR try to bring #250 and #268 forward.

The implementation of the traits does not make options (`ReadOptions`, `WriteOptions`) generic (or an associated type) since seems that is only useful if we want `Snapshot` implement `Get` and `GetCF`.
`Open` and `OpenCF` are not implemented is this PR.

Each trait is added in a separate commit to make the review simpler.


ssloboda:
Just want to say that I'm very excited about the possibility of this PR being accepted! Thanks to everyone who is working on it right now and to the maintainers for reviewing it!

As a user of `rust-rocksdb`, I think there is a compelling use case for at least leaving the door open to `impl Get/GetCF for Snapshot` without requiring a breaking API change. The use case in particular I am thinking of is the following:

A transaction (for an optimistic transaction DB) is built and committed, and some actions need to be taken when in response to the key-value pairs that were changed. The actions that need to be taken would like to have access to a snapshot of the DB at the point in time immediately after the transaction was committed. `impl Get/GetCF for Snapshot` would be very useful to allow that to happen.

Very excited about all the progress that has been made on mainline `rust-rocksdb` recently - thanks again to everyone for your hard work!

acrrd:
> A transaction (for an optimistic transaction DB) is built and committed, and some actions need to be taken when in response to the key-value pairs that were changed. The actions that need to be taken would like to have access to a snapshot of the DB at the point in time immediately after the transaction was committed. `impl Get/GetCF for Snapshot` would be very useful to allow that to happen while still writing generic code.

Keep in mind that at the moment `get_opt` of `DB` and `Snapshot` already have two different signature: the first take `&ReadOptions` the second `ReadOptions` so your code must be generic also over the type of option of the trait.

Said that, I'll be happy to add that implementation in some way for `Snapshot` but then
I think that we should also talk about if we want `Put, PutCF` and all other traits implemented for `WriteBatch` and in future `Transaction`. The problem here is that for this two structs the corresponding `ffi` functions never take a `WriteOptions` as input. PR #250 solve this by implementing the traits specifying `()` as the `WriteOptions` type. I don't like very much this solution, I'll prefer to have one trait per operation (eg. `Put`, `PutOpt`, `PutCF`, `PutCFOpt`) but this will add more  boilerplate.

Anyhow if there is interest in bringing this forward I can implement the solution we seems fit better.

acrrd:
I made `Get` and `GetCF` generic over the option type. Now `Snapshot` implements both.

src/prelude.rs
```diff
@@ -0,0 +1 @@
+pub use crate::{ops::*, ColumnFamily, Error, Options, ReadOptions, WriteOptions, DB};
```
aleksuss, line 1: A header about licence was missed.

src/handle.rs
```diff
@@ -0,0 +1,20 @@
+// Copyright 2019 Tyler Neely
```
aleksuss, line 1: Nit: Year is 2020 now 😄

acrrd:
> I'm still reviewing ... But question for now is: Do we really need a separate trait for every operation ? What's profit ?

Cleaner api and type safety. The implementation of `put_opt` for `Transaction` that you find in #250 is forced to use `()` to signal that the method is not supported. With a trait per op we can just avoid to implement the `PutOpt` for `Transaction`. 
I thought also about splitting the methods based on `opt` instead of `cf` but we have the same problem.
If you open the database without declaring column families does not make sense to have, for example, `put_cf` implemented for it.
I have to check but I think we don't need to have on trait per op for methods that take `ReadOptions` as parameter.

src/handle.rs
acrrd: But Tyler did the work back in 2019, I think his copyright hold since then.

acrrd:
Adding some small (pseudo)code examples.
With this PR we can implement `Put` for `Transaction` like:
```
impl Put for Transaction {
  fn put(key: K, value: V) {
    ffi::rocksdb_transaction_put(key, value)
  }
  fn put_opt(key: K, value: V, _writeopts: WriteOptions) {
    self.put(key, value)
  }
}
```
but I think is misleading to accept and option and silently don't use it.
Another option with this PR could be to just panic in `put_opt` like happen
if you are calling it on a read-only database.

In #250 `Put` for `Transaction` is implemented as:
```
impl Put<()> for Transaction {
  fn put(key: K, value: V) {
    ffi::rocksdb_transaction_put(key, value)
  }
  fn put_opt(key: K, value: V, writeopts: ()) {
    self.put(key, value)
  }
}
```
here at least is explicit that the option is not used.

With a trait per operation `Put` for `Transaction` can be implemented like:
```
impl Put<()> for Transaction {
  fn put(key: K, value: V) {
    ffi::rocksdb_transaction_put(key, value)
  }
}
```
this is the cleaner version in my opinion.

I think the solution implementable with this PR is the worse (sorry to not have noticed it before), we need to decide between the other two (or both).

DarkEld3r:
I agree that having small traits for operations results in a better (cleaner) api.

By the way, thanks for the effort. I feel sorry for not reacting to the pull request sooner, so I will try to review it in the nearest future.

acrrd:
If we reach an agreement on one trait per op I can implement that so you can review that directly.

aleksuss:
@acrrd Yeah. I suppose we do.

acrrd:
I'll work on it on the weekend!

src/handle.rs
spacejam: I don't have a preference what happens here

spacejam:
I think that this might be pretty confusing for people, as traits can add a lot of friction for users. They are harder to find docs for, and it can be confusing when an error about the method not being found happens. I think making `use rocksdb::prelude::*` an idiom that is added to every example is important for partially countering the second part, but it will still add some friction to looking things up in the documentation.

I think we should really think hard before introducing this much friction to the user experience.

mappum:
>I think that this might be pretty confusing for people, as traits can add a lot of friction for users. They are harder to find docs for, and it can be confusing when an error about the method not being found happens.

A solid solution to these points is to give the concrete types their own methods for each operation (with the same name and signature), which simply calls the trait method implementations.

This both adds the docs for the method to the obvious place on the concrete type, and allows calling the methods without pulling in the trait. Using the types generically with the trait all still works as expected.

Example:

```rust
mod stuff {
    pub trait Print {
        fn print(&self) { println!("printing") }
    }
    
    pub struct Foo;
    
    impl Print for Foo {}
    
    impl Foo {
        pub fn print(&self) { Print::print(self) }
    }
}

fn main() {
    // use stuff::Print; // this is optional
    stuff::Foo.print();
}
```

DarkEld3r:
Personally, I think that issues with documentation are a little less relevant for this project, because usually I check the original api and documentation first. After that it is easy to find the corresponding methods in the rust api.

acrrd:
The problem with the current implementation of traits is that each trait is implemented for every `T` that implements `Handle<rocksdb_t>` , `Handle` is private to the crate so is not in the documentation and is impossible for the user to discover which traits are implemented.
I will try something with the crates `deletegate` and `ambassador`.

DarkEld3r:
@acrrd I understand that this is a huge peace of work and don't want to bother you, but how are thing going? Despite the complications I think that the new api is much better and would like to see this merged (when it is ready).

acrrd:
I renamed the `DB` to `DBInner`. All the traits are implemented for `DBInner` through `Handle<_>`.
A new struct `DB` implements the traits by delegating  to `DBInner`. In the same way will be possible to create `ReadOnlyDB` or other variants.

src/db.rs
```diff
@@ -354,1054 +256,442 @@ impl DB {
...
-            if value.is_null() {
-                return Ok(None);
+macro_rules! make_new_db_with_traits {
+    ($struct_name:ident, [$($d:ty),+])  => (
+        #[derive(Delegate)]
+        $(#[delegate($d)])+
+        pub struct $struct_name(DBInner);
+
+        impl $struct_name {
+            delegate! {
```
stanislav-tkach, line 460: I'm not sure that the delegation is that useful herel. Personally, I would replace this delegation with a manual implementation, it wouldn't require much more boilerplate, right?

stanislav-tkach (approved):
Thanks for the effort! I really like the new api state. I have only several minor comments, but I have doubts about the `delegate` crate usage. As far as I can see, the `ambassador` crate really helps to reduce the boilerplate, but I cannot say the same about `delegate`.

src/db.rs
```diff
@@ -15,14 +15,37 @@
...
     ffi,
-    ffi_util::{from_cstr, opt_bytes_to_ptr, raw_data, to_cpath},
+    ffi_util::{from_cstr, raw_data, to_cpath},
+    handle::Handle,
+    ops::{
+        backup::BackupInternal,
+        checkpoint::CheckpointInternal,
+        column_family::{CreateColumnFamily, DropColumnFamily, GetColumnFamily},
+        compact_range::{CompactRangeCFOpt, CompactRangeOpt},
+        delete::{DeleteCFOpt, DeleteOpt},
```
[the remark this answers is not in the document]
acrrd: My thinking for the naming here was that this traits are not exposing methods that the end user will be using but they are only used internally to create/destroy the corresponding object. This traits are `pub (crate)` and they expose unsafe methods that return pointer so I wanted to make it clear with the suffix `Internal`. The other traits that we have are traits that the end user can use. I'm ok with a different naming.

src/handle.rs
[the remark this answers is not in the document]
acrrd: Thanks, this was already mentioned but I forgot to remove them.

src/db.rs
acrrd: Yes, not that much, each method will expand to 4 lines, for example:
`pub fn path(&self) -> & Path;`  expands to
```
#[always(inline)]
pub fn path(&self) -> & Path {
  self.0.path()
}
```

acrrd: If we know that the compiler is already inlining this in release mode we can remove the attribute.

src/db.rs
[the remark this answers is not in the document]
stanislav-tkach: It seems I have missed `pub (crate)`. In that case I'm OK with the current naming.

acrrd:
In the mean while I worked on the transaction here: https://github.com/acrrd/rust-rocksdb/tree/transaction
This api seems to work nicely with it.

src/checkpoint.rs
```diff
@@ -17,34 +17,37 @@
...
+use std::marker::PhantomData;
 use std::path::Path;
 
 /// Undocumented parameter for `ffi::rocksdb_checkpoint_create` function. Zero by default.
 const LOG_SIZE_FOR_FLUSH: u64 = 0_u64;
 
 /// Database's checkpoint object.
 /// Used to create checkpoints of the specified DB from time to time.
-pub struct Checkpoint {
+pub struct Checkpoint<'a> {
```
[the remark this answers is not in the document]
acrrd: It is a bit different for those. `Checkpoint` and `Snapshot` are created from the database while `Cache` and `Env` are optional input to create a database. We will need to modify both `Options` and `DB` to hold lifetimes. I tried to create an example [here](https://play.rust-lang.org/?version=stable&mode=debug&edition=2018&gist=2be014c57b2fde9e8783268bf23a5c96)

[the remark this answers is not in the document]
acrrd: This will work only if you can have only one `Cache` and `Env` per database.

acrrd:
I renamed the open methods of each database since `SecondaryDB::open_as_secondary` does not make to much sense at this point. But a part from the issue with `Cache` and `Env` I think this is ready. I'll open a PR with transaction support when this is merged but all the work on the branch is done.

For what concern #462, if it is ok to use `Option<'cache, 'env>` and `DB<'opt>` as I proposed in the previous comment I can implement that in this PR or one just after this.

spacejam:
I still think requiring this many traits is a significant negative impact on the user interface, but given that I am not really involved in this project beyond appointing maintainers, it's not really so much my decision. Do the other maintainers have consensus that this significant change is one that could possibly be accepted? This is a lot of work that keeps happening without it being clear to me that the maintainers have signaled that it is likely to be mergeable over time.

stanislav-tkach:
@acrrd 
It definitely makes sense to move `Cache` and `Env` changes to a separate pull request. This one is already too big and I suppose that not everyone would like complications with lifetimes.

@spacejam
I guess most of current maintainers are in the similar position as you. Personally, I do welcome these changes: current API is a little bit too "C-like" for my taste. Though I need to say that currently I'm not actively using RocksDB.

acrrd:
@stanislav-tkach Ok I'll do another PR for #462. And I'll make a separate PR to introduce transaction keeping the current structure of the project while we see what to with this.

@spacejam 
If your concern is about the documentation we can do what @mappum suggested, to have the methods directly show up.

acrrd:
I don't think it is necessary to keep this around. The main objective was to add transactions that have already been added.

PR #435, opened 2020-06-03
Export DEFAILT_COLUMN_FAMILY_NAME constant

DarkEld3r:
The changes are covered by the `test_column_family` test.

### DIFF
--- a/src/column_family.rs
+++ b/src/column_family.rs
@@ -14,6 +14,12 @@
 
 use crate::{ffi, Options};
 
+/// The name of the default column family.
+///
+/// The column family with this name is created implicitly whenever column
+/// families are used.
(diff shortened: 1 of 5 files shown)


src/column_family.rs
```diff
@@ -14,6 +14,12 @@
 
 use crate::{ffi, Options};
 
+/// The name of the default column family.
+///
+/// The column family with this name is created implicitly whenever column
+/// families are used.
+pub static DEFAULT_COLUMN_FAMILY_NAME: &str = "default";
```
aleksuss, line 21: Why `static` rather than `const` ?

DarkEld3r: No reason. :upside_down_face: Fixed.

PR #437, opened 2020-06-10
Don't segfault on failed a merge operator.

nelhage:
RocksDB requires that we populate the returned length even on failure:
[0]. Make sure we set length to zero before returning, and add a test
that confirms we see the error.

[0]: https://github.com/facebook/rocksdb/blob/02db03af8db7af03b8b026bc40d86e04ac741958/db/c.cc#L370


src/merge_operator.rs
```diff
@@ -451,4 +454,36 @@ mod test {
         }
         assert!(DB::destroy(&opts, path).is_ok());
     }
+
+    #[test]
+    fn failed_merge_test() {
```
aleksuss, line 459: What is purpose of the test ? 
P.S. The test is not passed.

src/merge_operator.rs
```diff
@@ -451,4 +454,36 @@ mod test {
...
+            let res = db.merge(b"key", b"new value");
+            match res.and_then(|_e| db.get(b"key")) {
+                Ok(val) => panic!("expected merge failure to propogate, got: {:?}", val),
+                Err(e) => {
+                    assert!(e.into_string().contains("Could not perform merge."));
+                }
+            }
+            // We can't DB::destroy, because that will detect the
+            // corruption and fail.
+            assert!(fs::remove_dir_all(path).is_ok());
```
aleksuss, line 478: If a panic occurs above `remove_dif_all` wouldn't be called.

src/merge_operator.rs
```diff
@@ -451,4 +454,36 @@ mod test {
...
+
+        let path = "_rust_rocksdb_failed_merge_test";
+        let mut opts = Options::default();
+        opts.create_if_missing(true);
+        opts.set_merge_operator("test operator", test_failing_merge, None);
+        {
+            let db = DB::open(&opts, path).expect("open with a merge operator");
+            db.put(b"key", b"value").expect("put_ok");
+            let res = db.merge(b"key", b"new value");
+            match res.and_then(|_e| db.get(b"key")) {
```
aleksuss, line 470: Could you explain why you use `and_then` here ?

src/merge_operator.rs
```diff
@@ -451,4 +454,36 @@ mod test {
...
+                Ok(val) => panic!("expected merge failure to propogate, got: {:?}", val),
```
aleksuss, line 471:
```suggestion
                Ok(val) => panic!("expected merge failure to propagate, got: {:?}", val),
```

src/merge_operator.rs
aleksuss: You could write something like:
```rust
let res = db.merge(b"key", b"new value");
assert!(res.is_ok());

match db.get(b"key") {
    Ok(val) => panic!("expected merge failure to propogate, got: {:?}", val),
    Err(e) => {
        assert!(e.as_ref().contains("Could not perform merge."));
    }
}
```

src/merge_operator.rs
```diff
@@ -451,4 +454,36 @@ mod test {
...
+                    assert!(e.into_string().contains("Could not perform merge."));
+                }
+            }
+            // We can't DB::destroy, because that will detect the
+            // corruption and fail.
+            assert!(fs::remove_dir_all(path).is_ok());
+        }
+    }
+
+    fn test_failing_merge(
```
aleksuss, line 482: The function could be in a scope of the test because it uses only there.

src/merge_operator.rs
nelhage: My understanding is that RocksDB is within its rights to perform the merge operation *either* at write time or at read time (and return the error then), and so I wanted to assert that the "merge+read" operation failed a whole, without asserting where Rocks actually returned the error.

In practice the code today does surface it at read time, so I could reorganize the code that way if you preferred.

src/merge_operator.rs
```diff
@@ -451,4 +454,35 @@ mod test {
...
+            let res = db.merge(b"key", b"new value");
+            match res.and_then(|_e| db.get(b"key")) {
+                Ok(val) => panic!("expected merge failure to propagate, got: {:?}", val),
+                Err(e) => {
+                    assert!(e.into_string().contains("Could not perform merge."));
+                }
+            }
+            // We can't DB::destroy, because that will detect the
+            // corruption and fail.
+            assert!(fs::remove_dir_all(path).is_ok());
```
DarkEld3r, line 485: Using `expect` or `unwrap` here would be better because we can see the reason of failure.

Does this test pass for you locally?

nelhage: Updated!

This is passing for me locally, yes.

DarkEld3r: Same for me. 

At the first glance, `DestroyDB` [should](https://github.com/facebook/rocksdb/blob/master/db/db_impl/db_impl.cc#L3665) delete everything even if something went wrong and report the reason of the first failure. But as far as I can see, there are still some files remains in this case.

Could you please try calling `DB::destroy` (and ignore the return value) before calling the fs::remove_dir_all?

src/merge_operator.rs
```diff
@@ -451,4 +454,40 @@ mod test {
...
+                }
+            }
+            // DB::destroy detects the corruption and returns an error
+            // without deleting all of the files, so we ignore its
+            // return value and then call `fs::remove_dir_all` to be
+            // sure we've cleaned up.
+            match DB::destroy(&opts, path) {
+                _ => (),
+            };
+            fs::remove_dir_all(path).expect("deleting test database");
```
aleksuss, line 490:
```rust
if DB::destroy(&opts, path).is_err() {
    fs::remove_dir_all(path).expect("deleting test database");
}
```
???

DarkEld3r: As far as I can see, the reason of the failure is a locked file, so I doubt it will solve the issue, but we can try. :upside_down_face:

src/merge_operator.rs
DarkEld3r: Honestly, I have run out of ideas.

Perhaps, `TempDir` from the `tempfile` crate can be used. At least it generates a directory with a random name, so it won't affect consequent test runs and the temp directory should be removed later even in case of a locked file.

src/merge_operator.rs
aleksuss: Time to move all tests to the `tests` module and use `DBPath` everywhere 😄

nelhage: `DBPath` would also fail here, for the same a `DB::destroy` alone doesn't work, but we could amend it to handle this case somehow. Do you want me to make that change?

DarkEld3r: Ideally, we need to find out why this test is having issues with deleting files, but as a quick solution personally I would use `TempDir`.

nelhage: Pushed a fix here! I realized now that I've been confused why `DB::destroy` was failing; I'd been blaming on the corrupted database, but the real problem was just that I was calling it while `db` was still alive, and so it couldn't lock the database. So now we use a tempfile, and also `DB::destroy().expect()`. Still passing locally.

DarkEld3r: Cool. :+1: 

Though as the problem is resolved I would prefer to use `DBPath` instead of tempdir + manual call to `DB::destroy()`.

nelhage: I moved all the merge_operator tests to `tests/` and to use `DBPath`; let me know if you'd rather I roll that back and send another PR.

PR #438, opened 2020-06-11
Fix slice transformers with no in_domain callback 

nelhage:
The C API [has no provision for this being `NULL`](https://github.com/facebook/rocksdb/blob/27e593fbe10efa5a562ca79b1939ffb69d25163f/db/c.cc#L445-L447)


src/slice_transform.rs
```diff
@@ -118,6 +111,46 @@ pub unsafe extern "C" fn in_domain_callback(
 ) -> u8 {
     let cb = &mut *(raw_cb as *mut TransformCallback);
     let key = slice::from_raw_parts(raw_key as *const u8, key_len as usize);
-    let in_domain = cb.in_domain_fn.unwrap();
-    in_domain(key) as u8
+    if let Some(in_domain) = cb.in_domain_fn {
+        in_domain(key) as u8
+    } else {
+        0xff_u8
```
aleksuss, line 117: Nit: The postfix `_u8` is needless because the compiler knows a type of a return value of the function.

src/slice_transform.rs
```diff
@@ -118,6 +111,46 @@ pub unsafe extern "C" fn in_domain_callback(
...
+        } else {
+            slice
+        }
+    }
+
+    #[test]
+    fn prefix_test() {
+        use crate::{Options, DB};
+
+        let path = "_rust_rocksdb_prefix_test";
```
DarkEld3r, line 137: You can use `DBPath` here in order not to destroy the database manually.

src/slice_transform.rs
```diff
@@ -118,6 +111,46 @@ pub unsafe extern "C" fn in_domain_callback(
...
+            None,
+        ));
+        opts.set_memtable_prefix_bloom_ratio(0.1);
+
+        {
+            let db = DB::open(&opts, path).unwrap();
+            db.put(b"key_sfx1", b"a").unwrap();
+            db.put(b"key_sfx2", b"b").unwrap();
+
+            assert!(db.get(b"key_sfx1").unwrap().unwrap() == b"a");
```
DarkEld3r, line 152: I suppose `assert_eq` would make more sense here.

nelhage:
Fixed both comments!

PR #441, opened 2020-06-30
Add compaction filter factory API

unrealhoang:
And other option configuration.

### DIFF
--- a/src/compaction_filter.rs
+++ b/src/compaction_filter.rs
@@ -31,8 +31,36 @@ pub enum Decision {
     Change(&'static [u8]),
 }
 
+/// CompactionFilter allows an application to modify/delete a key-value at
+/// the time of compaction.
 pub trait CompactionFilter {
+    /// The compaction process invokes this
+    /// method for kv that is being compacted. The application can inspect
+    /// the existing value of the key and make decision based on it.
+    ///
+    /// Key-Values that are results of merge operation during compaction are not
+    /// passed into this function. Currently, when you have a mix of Put()s and
+    /// Merge()s on a same key, we only guarantee to process the merge operands
+    /// through the compaction filters. Put()s might be processed, or might not.
+    ///
+    /// When the value is to be preserved, the application has the option
+    /// to modify the existing_value and pass it back through new_value.
+    /// value_changed needs to be set to true in this case.
+    ///
+    /// Note that RocksDB snapshots (i.e. call GetSnapshot() API on a
+    /// DB* object) will not guarantee to preserve the state of the DB with
+    /// CompactionFilter. Data seen from a snapshot might disppear after a
+    /// compaction finishes. If you use snapshots, think twice about whether you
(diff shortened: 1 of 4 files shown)


src/compaction_filter_factory.rs
```diff
@@ -0,0 +1,128 @@
...
+    ffi::rocksdb_compactionfilter_create(
+        filter_ptr as *mut c_void,
+        Some(compaction_filter::destructor_callback::<F::Filter>),
+        Some(compaction_filter::filter_callback::<F::Filter>),
+        Some(compaction_filter::name_callback::<F::Filter>),
+    )
+}
+
+#[cfg(test)]
+mod test {
```
DarkEld3r, line 76: Please rename the module to `tests` - it should fix the Clippy error.

src/compaction_filter.rs
```diff
@@ -51,19 +56,32 @@ where
...
+    fn filter(&mut self, level: u32, key: &[u8], value: &[u8]) -> Decision {
+        (self.filter_fn)(level, key, value)
+    }
+}
+
+pub unsafe extern "C" fn destructor_callback<F>(raw_cb: *mut c_void)
+where
+    F: CompactionFilter,
+{
+    let _: Box<F> = mem::transmute(raw_cb as *mut F);
```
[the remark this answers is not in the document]
unrealhoang: This is the old code, I'm not really sure about the intention so I kept it the same. But I can change as you suggested.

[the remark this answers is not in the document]
DarkEld3r: I think it is better to use `Box::from_raw` in both places.

DarkEld3r:
The documentation suggests that `wildcard_imports` lint shouldn't be triggered in any module which name contains "test". I have no clues why it isn't working in our case. I have submitted the [issue](https://github.com/rust-lang/rust-clippy/issues/5764), lets hope someone will suggest a workaround.

DarkEld3r:
The issue with the Clippy is fixed, but the release should be available next week, so I guess it is OK to suppress this warning for now.

src/compaction_filter.rs
```diff
@@ -32,6 +31,11 @@ pub enum Decision {
     Change(&'static [u8]),
 }
 
+pub trait CompactionFilter {
```
DarkEld3r, line 34: It would be nice to have documentation for this trait.

src/compaction_filter_factory.rs
```diff
@@ -0,0 +1,129 @@
+use std::ffi::CStr;
+
+use libc::{self, c_char, c_void};
+
+use crate::{
+    compaction_filter::{self, CompactionFilter},
+    ffi,
+};
+
+pub trait CompactionFilterFactory {
```
DarkEld3r, line 10: Same here. :upside_down_face:

unrealhoang:
@DarkEld3r Thank you for your review, can you take a another look?

PR #442, opened 2020-06-30
Adding read/write/db/compaction options

linxGnu:
This PR:
- Includes many options for read/write/db/compaction which is up to date with rocksdb 6.8.1. 
- Also added comments for various settings.
- Comments taken from c++ upstream


src/cache.rs
```diff
@@ -0,0 +1,59 @@
...
+use libc::{self, size_t};
+
+use crate::ffi;
+
+// Cache is a cache used to store data read from data in memory.
+pub struct Cache {
+    pub(crate) inner: *mut ffi::rocksdb_cache_t,
+}
+
+#[allow(dead_code)]
```
DarkEld3r, line 24: I suppose this is no longer needed because methods are public.

src/cache.rs
```diff
@@ -0,0 +1,59 @@
...
+impl Cache {
+    /// Create a lru cache with capacity
+    pub fn new_lru_cache(capacity: size_t) -> Cache {
+        let cache = unsafe { ffi::rocksdb_cache_create_lru(capacity) };
+        if cache.is_null() {
+            panic!("Could not create Cache");
```
DarkEld3r, line 30: Personally I would prefer to return `Result` instead of panicking here, but I'm not 100% sure because we do have very similar code with panic for `Options`. On the other hand, `DB::open` returns a result.

src/db_options.rs
```diff
@@ -1872,17 +2370,34 @@ impl Default for WriteOptions {
...
     // that struct needs an "iterator" impl too.
+
+    /// Specify whether the "data block"/"index block"/"filter block"
+    /// read for this iteration should be cached in memory?
+    /// Callers may wish to set this field to false for bulk scans.
+    ///
+    /// Default: true
     #[allow(dead_code)]
-    fn fill_cache(&mut self, v: bool) {
+    pub fn fill_cache(&mut self, v: bool) {
```
DarkEld3r, line 2392: I suppose `allow(dead_code)` can be removed.

src/db_options.rs
```diff
@@ -1872,17 +2370,34 @@ impl Default for WriteOptions {
     }
 }
 
+#[derive(Debug, Copy, Clone, PartialEq)]
+pub enum ReadTier {
+    /// Reads data in memtable, block cache, OS cache or storage.
+    All = 0 as isize,
+    /// Reads data in memtable or block cache.
+    BlockCache = 1 as isize,
```
DarkEld3r, line 2378: Are these `as isize` casts required? As far as I can see, these values are transformed into `c_int` that isn't the same as `isize`. I guess `i32` is more appropriate here, but perhaps these casts can be simply omitted?

src/db_options.rs
```diff
@@ -2108,6 +2716,237 @@ pub enum DBRecoveryMode {
     SkipAnyCorruptedRecord = ffi::rocksdb_skip_any_corrupted_records_recovery as isize,
 }
 
+#[derive(Debug, Copy, Clone, PartialEq)]
+pub enum CompactionAccessPattern {
+    None = 0 as isize,
```
DarkEld3r, line 2721: Same here (redundant `as isize`).

src/db_options.rs
```diff
@@ -2108,6 +2716,237 @@ pub enum DBRecoveryMode {
...
+    /// Default: ::Total
+    pub fn set_stop_style(&mut self, style: UniversalCompactionStopStyle) {
+        unsafe {
+            ffi::rocksdb_universal_compaction_options_set_stop_style(self.inner, style as c_int);
+        }
+    }
+}
+
+#[derive(Debug, Copy, Clone, PartialEq)]
+#[allow(dead_code)]
```
DarkEld3r, line 2877: Please remove `allow(dead_code)`.

src/db_options.rs
```diff
@@ -2108,6 +2716,237 @@ pub enum DBRecoveryMode {
...
+pub enum BottommostLevelCompaction {
+    /// Skip bottommost level compaction
+    Skip = 0 as isize,
```
DarkEld3r, line 2880: I suppose `u8` is more appropriate here, but I think that instead of using `as isize` for each value it is easier to use `repr(u8)` for the whole enum.

src/perf.rs
```diff
@@ -0,0 +1,175 @@
...
+
+use libc::{c_char, c_int, c_uchar, c_void};
+use std::ffi::CStr;
+
+use crate::ffi;
+
+#[derive(Debug, Copy, Clone, PartialEq)]
+pub enum PerfStatsLevel {
+    /// Unknown settings
+    Uninitialized = 0 as isize,
```
DarkEld3r, line 23: Similar here.

src/perf.rs
```diff
@@ -0,0 +1,175 @@
...
+    // 	rocksdb_env_get_file_size_nanos,
+    // 	rocksdb_env_get_file_modification_time_nanos,
+    // 	rocksdb_env_rename_file_nanos,
+    // 	rocksdb_env_link_file_nanos,
+    // 	rocksdb_env_lock_file_nanos,
+    // 	rocksdb_env_unlock_file_nanos,
+    // 	rocksdb_env_new_logger_nanos,
+    // 	rocksdb_total_metric_count = 68
+    // }
+    pub fn metric(&self, id: c_int) -> u64 {
```
DarkEld3r, line 172: There are a lot of variants here, but I would still prefer to use an enum (with the `#[non_exhaustive]` attribute).

linxGnu:
Hello @DarkEld3r , thank you very much for your review :+1:

I have addressed your comments. Please kindly take a look on updates.

src/db_options.rs
```diff
@@ -2108,6 +2716,238 @@ pub enum DBRecoveryMode {
...
+            ffi::rocksdb_universal_compaction_options_set_stop_style(self.inner, style as c_int);
+        }
+    }
+}
+
+#[derive(Debug, Copy, Clone, PartialEq)]
+#[repr(u8)]
+pub enum BottommostLevelCompaction {
+    /// Skip bottommost level compaction
+    Skip = 0,
```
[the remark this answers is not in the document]
linxGnu: Thank you very much for guiding. I think export in `lib.rs` is a good choice. Will update soon and let you know.
I have just realized that we are missing manual compaction methods which accepts CompactOptions :thinking: 
I will add those and tests as well.

linxGnu:
@DarkEld3r I have just added test for compaction, perf context stats and cache. PTAL, thank you

src/db.rs
```diff
@@ -933,6 +934,27 @@ impl DB {
         }
     }
 
+    pub fn compact_range_opt<S: AsRef<[u8]>, E: AsRef<[u8]>>(
```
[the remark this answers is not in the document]
linxGnu: Thank you, I am adding them right now as well as tests for prefix_extract functionality. I found the test is missing too.

linxGnu:
@DarkEld3r @aleksuss PTAL, thank you. I have just added tests for prefix_extract as well as comments for public methods

PR #443, opened 2020-07-10
Link C++ stdlib when using prebuilt rocksdb

unrealhoang:
This fix build when using prebuilt RocksDB

### DIFF
--- a/librocksdb-sys/build.rs
+++ b/librocksdb-sys/build.rs
@@ -333,6 +333,14 @@ fn main() {
         println!("cargo:rerun-if-changed=rocksdb/");
         fail_on_empty_directory("rocksdb");
         build_rocksdb();
+    } else {
+        let target = env::var("TARGET").unwrap();
+        // according to https://github.com/alexcrichton/cc-rs/blob/master/src/lib.rs#L2189
+        if target.contains("apple") || target.contains("freebsd") || target.contains("openbsd") {
+            println!("cargo:rustc-link-lib=dylib=c++");
+        } else if target.contains("linux") {
+            println!("cargo:rustc-link-lib=dylib=stdc++");
+        }
     }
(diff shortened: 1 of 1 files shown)


librocksdb-sys/build.rs
```diff
@@ -359,4 +360,15 @@ fn main() {
...
+    if link_sys {
+        let target = env::var("TARGET").unwrap();
+        // according to https://github.com/alexcrichton/cc-rs/blob/master/src/lib.rs#L2189
+        if target.contains("apple") || target.contains("freebsd") || target.contains("openbsd") {
+            println!("cargo:rustc-link-lib=dylib=c++");
+        } else if target.contains("msvc") {
+        } else {
+            println!("cargo:rustc-link-lib=dylib=stdc++");
+        }
+    }
```
aleksuss, line 373: Overall looks good. But couple nitpicks.

1. I'd get rid of the additional variable `link_sys` and just would use `if/else`.
2.  The construction:
 ```rust
} else if target.contains("msvc") {
} else {
```
looks a bit weird. By the way. There are targets with `netbsd` suffix exist. I suppose such targets should be linked with `c++` library rather than `stdc++`. So I'd propose something like: 
```rust
    if !try_to_find_and_link_lib("ROCKSDB") {
        println!("cargo:rerun-if-changed=rocksdb/");
        fail_on_empty_directory("rocksdb");
        build_rocksdb();
    } else {
        let target = env::var("TARGET").unwrap();
        // according to https://github.com/alexcrichton/cc-rs/blob/master/src/lib.rs#L2189
        if target.contains("apple") || target.contains("freebsd") || target.contains("openbsd") {
            println!("cargo:rustc-link-lib=dylib=c++");
        } else if target.contains("linux") {
            println!("cargo:rustc-link-lib=dylib=stdc++");
        }
    }
```

PR #446, opened 2020-07-16
Support fetching sst files metadata, delete files in range, get mem usage

linxGnu:
This PR:
- Support fetching sst files metadata via api
- Support delete file in range api
- Option to set memtable whole key filtering
- Support api to get approximate memory usage

### DIFF
--- a/src/db.rs
+++ b/src/db.rs
@@ -15,7 +15,7 @@
 
 use crate::{
     ffi,
-    ffi_util::{opt_bytes_to_ptr, to_cpath},
+    ffi_util::{from_cstr, opt_bytes_to_ptr, raw_data, to_cpath},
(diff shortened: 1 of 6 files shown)


src/db.rs
```diff
@@ -1270,6 +1270,97 @@ impl DB {
             Ok(())
         }
     }
+
+    /// Returns a list of all table files with their level, start key
+    /// and end key
+    pub fn livefiles(&self) -> Result<Vec<LiveFile>, Error> {
```
DarkEld3r, line 1276: Personally I would prefer to name this function `live_files`.

linxGnu: I addressed your comment, PTAL :+1:

PR #449, opened 2020-07-25
librocksdb-sys: do not set rerun-if-changed=build.rs

xu-cheng:
This is unnecessary as per the Cargo book:

> Cargo automatically handles whether or not the script itself needs to
> be recompiled, and of course the script will be re-run after it has
> been recompiled. Otherwise, specifying build.rs is redundant and unnecessary.

https://doc.rust-lang.org/cargo/reference/build-scripts.html#rerun-if-changed

Additionally, this sometimes causes the rocksdb being unnecessarily
recomputed whenever I make any change in the downstream project.
And it takes a long time to finish.

### DIFF
--- a/librocksdb-sys/build.rs
+++ b/librocksdb-sys/build.rs
@@ -325,8 +325,6 @@ fn cxx_standard() -> String {
 }
 
 fn main() {
-    println!("cargo:rerun-if-changed=build.rs");
-
     bindgen_rocksdb();
 
     if !try_to_find_and_link_lib("ROCKSDB") {


DarkEld3r:
> If the build script inherently does not need to re-run under any circumstance, then emitting cargo:rerun-if-changed=build.rs is a simple way to prevent it from being re-run.

Perhaps I'm missing something, but doesn't this mean that the script will be re-run only if the `build.rs` itself is changed?

xu-cheng:
> > If the build script inherently does not need to re-run under any circumstance, then emitting cargo:rerun-if-changed=build.rs is a simple way to prevent it from being re-run.

> Perhaps I'm missing something, but doesn't this mean that the script will be re-run only if the build.rs itself is changed?

I think that refers to the case that there isn't any `cargo:rerun-if-changed=`. In this case, the cargo will run the build script all the times. Therefore, it is recommended to emit at least one `cargo:rerun-if-changed=`. You can find the relevant doc in the below:

>  By default, it takes a conservative approach of always re-running the build script if any file within the package is changed (or the list of files controlled by the exclude and include fields). For most cases, this is not a good choice, so it is recommended that every build script emit at least one of the rerun-if instructions (described below). If these are emitted, then Cargo will only re-run the script if the given value has changed.

PR #453, opened 2020-08-05
Add Send and Sync implementation for WriteBatch

### LINKED ISSUE #347: Can we implement Send on WriteBatch?
In Rust, I am writing code where I construct WriteBatch objects in a worker thread, then I use a channel to send WriteBatches to a different thread for writing to RocksDB.

Because a WriteBatch does not depend on external state (it does not reference a DB object), I think it would be safe to add an implementation for Send to WriteBatch.

Is this correct?

DarkEld3r:
Closes #347.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,7 @@
 ## [Unreleased]
 
 * Export the `DEFAULT_COLUMN_FAMILY_NAME` constant.
+* Added `Sync` and `Send` implementations to `WriteBatch`.
 
 ## 0.14.0 (2020-04-22)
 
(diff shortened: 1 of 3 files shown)


kaimast:
I cannot review this but it looks good to me. Was about to issue an almost identical pull request.

DarkEld3r:
According to [this](https://github.com/facebook/rocksdb/blob/fdf882ded218344c136c97daf76dfb59e4bc155f/include/rocksdb/write_batch.h#L20), `WriteBatch` shouldn't implement `Sync`:

> // Multiple threads can invoke const methods on a WriteBatch without
// external synchronization, but if any of the threads may call a
// non-const method, all threads accessing the same WriteBatch must use
// external synchronization.

But I suppose that `Send` can still be applied.

PR #456, opened 2020-08-23
Extend github actions

stanislav-tkach:
Summary:
- Travis is removed.
- Github actions are used for all testing.
- Additionally to Windows and Linux tests are now run on MacOS.

Clippy check now fails because the pull request is made from the fork, so it will be fixed when (and if) this is merged.


.github/workflows/rust.yml
```diff
@@ -1,42 +1,89 @@
...
+          args: --all -- --check
 
-      - name: Install dependencies
-        run: choco install llvm -y
+  clippy:
+    name: Clippy
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout sources
+        uses: actions/checkout@v2
```
aleksuss, line 30: Obviously, you need to add a fetching submodules as well:
```yaml
with:
    submodules: true
```

.github/workflows/rust.yml
```diff
@@ -1,42 +1,89 @@
...
+        uses: actions-rs/toolchain@v1
+        with:
+          toolchain: stable
+          components: clippy
+          profile: minimal
+          override: true
+      - name: Run clippy
+        uses: actions-rs/clippy-check@v1
+        with:
+          token: ${{ secrets.GITHUB_TOKEN }}
```
aleksuss, line 41: Just out of curiosity. Is that token really needed ?

stanislav-tkach: According to the documentation, it is required: https://github.com/actions-rs/clippy-check#inputs

aleksuss (approved):

PR #460, opened 2020-08-30
bump bindgen version

jensim:
Nothing tested, but for the build tests

### DIFF
--- a/librocksdb-sys/Cargo.toml
+++ b/librocksdb-sys/Cargo.toml
@@ -9,7 +9,6 @@ readme = "README.md"
 repository = "https://github.com/rust-rocksdb/rust-rocksdb"
 keywords = [ "bindings", "ffi", "rocksdb" ]
 categories = [ "api-bindings", "database", "external-ffi-bindings" ]
-
 links = "rocksdb"
 
 [features]
--- a/librocksdb-sys/Cargo.toml
+++ b/librocksdb-sys/Cargo.toml
@@ -30,5 +30,5 @@ uuid = { version = "0.8", features = ["v4"] }
 
 [build-dependencies]
 cc = { version = "^1.0", features = ["parallel"] }
-bindgen = "0.54"
+bindgen = "0.55"
 glob = "0.3"


jensim:
@aleksuss Having read the [contributing guidelines](CONTRIBUTING.md), I thought that I'd make this a `draft pull request`, as i have not tested the lib locally in any way. 
I just found that this library/binary was on the top 1000 list of crates.io and that it has an outdated transitive dependency to clang-sys through bindgen.

aleksuss:
@jensim it's ok. Make the PR **ready for review** state.

jensim:
> @jensim it's ok. Make the PR **ready for review** state.

Done!

I also rebased on updates in master

aleksuss (approved):

silasb:
New to rust and cargo, but do we need to bump this and publish this?

Looks like my install of `librocksdb-sys` is pulling in `bindgen 0.54.0`

jensim:
> New to rust and cargo, but do we need to bump this and publish this?
> 
> Looks like my install of `librocksdb-sys` is pulling in `bindgen 0.54.0`

@silasb glad to see new rustaceans 😁🥳
If you are a library consumer of librocksdb-sys, this is not to be considered a huge problem in any way.

I am a bit of a security buff, so I do a lot of preemptive work of keeping libraries on latest stable version, in case security issues are discovered. This makes patching security issues a lot faster. My little [security spider](https://github.com/jensim/rust-top-audit) will monitor for the latest version of the top 1000 crates, and will therefore keep listing rocksdb as out of date until it publishes a bumped version.

If you are asking about this for the rocksdb crates... Then I would suggest publishing a new version sooner rather than later. For the reason that it contains a library crate, and there may be lib consumer to consider. 


Best regards
Jens

silasb:
Hi @jensim I was asking about this for RocksDB as I was running into issues getting it to build.  Looks like the issue has nothing to do with librocksdb-sys or RocksDB crate at all. Sorry for the noise. I'll comment on my fix in case anyone else makes it here.

My issue has something to do with LinuxBrew `ldd` vs Debian `ldd` that was causing me issues when building `rocksdb`.  Removing `linuxbrew` from my `PATH` made `ldd /usr/lib/llvm-10/lib/libclang-10.so.1` return all the correct shared libraries.

Looks like there isn't an easy way to override the shared library lookup to force `libloading` or `libc` crates to ignore `linuxbrew` paths other than manually changing my `PATH`.

Again, sorry for the noise.

PR #473, opened 2020-11-07
Use matches! to fix Clippy warning

stanislav-tkach:


stanislav-tkach:
I like matches, but I'm not so sure about `map_or_else`.

aleksuss:
I've done this in #466 🙃

aleksuss (approved):

PR #497, opened 2021-02-18
Keep Cache and Env alive with Rc

### LINKED ISSUE #462: Issues with Cache
Hey, we've been updating to v0.15. We were using https://docs.rs/rocksdb/0.15.0/rocksdb/struct.BlockBasedOptions.html#method.set_lru_cache before, but since that is deprecated we are refactoring. We noticed two main issues:

- The dependency of `DB` on the lifetime of the `Cache` is not encoded in the type system. As far as I understand the documentation, dropping the Cache before dropping the DB results in undefined behavior. Perhaps this could be safeguarded by either making `Cache` refcounted or by using lifetimes.
- `Cache` is not `Send` because it contains a raw pointer. I'm not sure if that's desired behavior. The implementation of `set_lru_cache` (https://docs.rs/rocksdb/0.15.0/src/rocksdb/db_options.rs.html#419-426) suggests `Send`ness is not required.

acrrd:
Fix #462.

- Add a lifetime to `Checkpoint` to ensure that it does not outlive the `DB`
- Wrap `Cache` and `Env` in an `Rc` and keep a copy of that in the database to ensure that they are not dropped before the database is.

This add two auxiliary structs where to keep `Cache` and `Env` that are set using `set_row_cache` in `Options` or `set_block_cache`, `set_block_compressed` in `BlockBasedOptions`. When a `DB` is open this structs are cloned and kept in the database to ensure the lifetime requirements are met.

This approach allow us to have an unchanged api (with the exception of `Checkpoint` life time) but the compiler will not help us if we forget to keep track of something. Implementing this with lifetime will require to have four lifetimes parameters in (at least), `Options`, `BlockBasedOptioins`, `DB`, and `ColumnFamilyDescriptor`.

Do we need to keep track of `Cache` and `Env` that are in the `Options` inside `ColumnFamilyDescriptor` when we open a database through them?

### DIFF
--- a/src/db_options.rs
+++ b/src/db_options.rs
@@ -15,7 +15,7 @@
 use std::ffi::{CStr, CString};
 use std::mem;
 use std::path::Path;
-use std::rc::Rc;
+use std::sync::Arc;
 
 use libc::{self, c_char, c_int, c_uchar, c_uint, c_void, size_t};
 
@@ -48,7 +48,7 @@ impl Drop for CacheWrapper {
     }
 }
 
-pub struct Cache(pub(crate) Rc<CacheWrapper>);
+pub struct Cache(pub(crate) Arc<CacheWrapper>);
 
 impl Cache {
     /// Create a lru cache with capacity
@@ -57,7 +57,7 @@ impl Cache {
         if cache.is_null() {
             Err(Error::new("Could not create Cache".to_owned()))
         } else {
-            Ok(Cache(Rc::new(CacheWrapper { inner: cache })))
+            Ok(Cache(Arc::new(CacheWrapper { inner: cache })))
         }
     }
 
@@ -93,6 +93,8 @@ impl Cache {
 ///
 /// Note: currently, C API behinds C++ API for various settings.
 /// See also: `rocksdb/include/env.h`
+pub struct Env(Arc<EnvWrapper>);
+
 struct EnvWrapper {
     inner: *mut ffi::rocksdb_env_t,
 }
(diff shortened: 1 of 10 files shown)


stanislav-tkach (approved):
Looking good to me! 👍

stanislav-tkach:
> Do we need to keep track of Cache and Env that are in the Options inside ColumnFamilyDescriptor when we open a database through them?

I'm not one hundred percent sure, but I'm afraid that yes. I need to check the RocksDB code more thoroughly, though.

src/db_options.rs
```diff
@@ -125,66 +134,105 @@ impl Env {
...
+            block_based: self
+                .block_based
+                .as_ref()
+                .map(BlockBasedOptionsMustOutliveDB::clone),
+        }
+    }
+}
+
+#[derive(Default)]
+pub struct BlockBasedOptionsMustOutliveDB {
```
aleksuss, line 226: Does it really need to be public rather than `pub(crate)`  ? If so then I'd rename it to `BlockBasedCache` or something because it seems weird name as for public API.

acrrd: It could be private!

acrrd:
I'm saving also the `Options` from a `ColumnFAmilyDescriptor`, we can remove this later if we see that is not needed.
We can have two databases crated with the same `Env` and `Cache` and send them to two different thread, so I changed `Rc` to `Arc`. I would like to add the test `is_sync<OptionsMustOutliveDb>` and `is_send<OptionsMustOutliveDb>` but at the moment `Env` and `Cache` are not marker `Send`/Sync`. For `Env` a comment says that is safe to implemented them, I think would be ok to implement them also for `Cache` since it can be used from different database in different thread.

aleksuss:
@acrrd resolve please conflicts. Thanks.

acrrd:
Done,thanks.

aleksuss (approved):
LGTM

Issue #494, opened 2021-02-04
merge operator delete_callback 


zhangsoledad:
I got a case that needs to use `merge` function.
I notice https://github.com/rust-rocksdb/rust-rocksdb/pull/457 introduce `delete_callback` to handle memory release and avoid copy.
https://github.com/xuchen-plus/rust-rocksdb/blob/392d20a4df6c9319822eaabfd517c7697a735377/src/db_options.rs#L945-L952
```rust
            let mo = ffi::rocksdb_mergeoperator_create(
                mem::transmute(cb),
                Some(merge_operator::destructor_callback),
                Some(full_merge_callback),
                Some(partial_merge_callback),
                Some(merge_operator::delete_callback),
                Some(merge_operator::name_callback),
            );
```

But https://github.com/rust-rocksdb/rust-rocksdb/pull/481  overwrite it back.

https://github.com/BoOTheFurious/rust-rocksdb/blob/9881373ed920727523e5c1cb714a2680ba97c1e2/src/db_options.rs#L944-L950

https://github.com/BoOTheFurious/rust-rocksdb/blob/9881373ed920727523e5c1cb714a2680ba97c1e2/src/db_options.rs#L971-L975

```
           let mo = ffi::rocksdb_mergeoperator_create(
                Box::into_raw(cb) as _,
                Some(merge_operator::destructor_callback::<F, F>),
                Some(full_merge_callback::<F, F>),
                Some(partial_merge_callback::<F, F>),
                None,
                Some(merge_operator::name_callback::<F, F>),
            );
```

So my question: Is this mistake? Or this is on purpose？

stanislav-tkach:
Honestly, I have missed that detail in the #481 pull request. Would you like to submit a pull request to fix that issue?

PR #481, opened 2020-11-23
Fix merge fn closure from pr385

BoOTheFurious:
Hello,
I 've got a use case which need to use stateful closure as merge function. As an old PR refers to my needs (PR#385 1 year old) I have rebase on current master. I have also integrated PR#480 to pass the test.
All works fine. I hope It will be accepted quickly.

Thanks @mrobakowski for the works on the initial PR.

Regards

PR #457, opened 2020-08-24
Avoid copy for merge operator result using delete_callback

xuchen-plus:
This PR uses `delete_callback` to release memory of merge operator result.

The `rocksdb_mergeoperator_t` from Rocksdb C API allows a `delete_callback` argument which allows user to handle memory release [db/c.cc](https://github.com/facebook/rocksdb/blob/master/db/c.cc#L372)
The merge result `Vec<u8>` is turned into a boxed slice and turned into raw pointer by `Box::into_raw`. In `delete_callback`, `Box::from_raw` is called to wrap the raw pointer back and the original memory will be released here.

Testing done:
All unit tests passed.

Fixes: #410

PR #500, opened 2021-02-22
fix: merge delete_callback omitted by mistake

zhangsoledad:
https://github.com/rust-rocksdb/rust-rocksdb/pull/457  make a change
> The merge result Vec<u8> is turned into a boxed slice and turned into raw pointer by Box::into_raw. In delete_callback, Box::from_raw is called to wrap the raw pointer back and the original memory will be released here.

https://github.com/rust-rocksdb/rust-rocksdb/pull/481 omit `delete_callback` by mistake.

According
```
if (delete_value_ != nullptr) {
      (*delete_value_)(state_, tmp_new_value, new_value_len);
} else {
      free(tmp_new_value);
}
```
https://doc.rust-lang.org/std/boxed/struct.Box.html#method.into_raw
https://doc.rust-lang.org/nomicon/ffi.html#destructors
I think this is UB.

fix https://github.com/rust-rocksdb/rust-rocksdb/issues/494

### DIFF
--- a/src/db_options.rs
+++ b/src/db_options.rs
@@ -962,7 +962,7 @@ impl Options {
                 Some(merge_operator::destructor_callback::<F, F>),
                 Some(full_merge_callback::<F, F>),
                 Some(partial_merge_callback::<F, F>),
-                None,
+                Some(merge_operator::delete_callback),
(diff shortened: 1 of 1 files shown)


stanislav-tkach (approved):
👍

PR #506, opened 2021-03-26
Remove need for &mut self in create_cf and drop_cf (v2)

ryoqun:
Here's the polished version, finally!

https://github.com/rust-rocksdb/rust-rocksdb/pull/496#discussion_r600965996

> @aleksuss Hi, again. I want this to move forward :) (cc: @carllin)
> 
> > What about to bound an invoke of this with trait ??? I'll try to provide the idea with code soon.
> 
> Maybe, were you thinking about like this?: [ryoqun@1be386d](https://github.com/ryoqun/rust-rocksdb/commit/1be386d3d388901398b29e67ecc5f56d6f59f9cf)
> 
> Sorry for very rough code. But this eliminates bunch of various concerns arisen from this pr's review:
> 
>     * ColumnFamily is bounded in the case of MultiThread with zero runtime overhead (I even stopped using Arc)
> 
>     * ... and no `derive(Clone)` on `ColumnFamily`
> 
>     * will have no `panic!`s (if everything is correctly rewritten)
> 
>     * keeps api compatibility and no multithread methods.
> 
>     * no enum runtime variant branching (although this is cheap)
> 
>     * downside: a bit of internal code churn (I'll promise I'll work hard to fix them all)
> 
> 
> If this direction is really to go from maintainer's perspective, I'll finish this very quickly. :)

PR #509, opened 2021-04-19
Don't leak dropped column families

ryoqun:
## problem

Regardless `SingleThreaded` or `MultiThreaded`, `rust-rocksdb` crate doesn't free files even after `drop_cf()` as long as we're opening `DB` instance. I added `test_no_leaked_column_family` to demonstrate that.

For `SingleThreaded`, it has been existed since the begging. So this isn't regression of #506...

For `MultiThreaded`, we (Solana) originally introduced the `MultiThreaded` for the ultimate goal of freeing dropped cf resources in the middle of in-process cf mutation under multi-threaded context. For that end, we managed to introduce the very `MultiThreaded` mode thanks for persistent review and eventual merge and release of 0.16.0. But due to this existing bug, we still can't realize the ultimate goal.

The root cause of the leak is that `rocksdb_column_family_handle_destroy` isn't called until we `Drop::drop()` the whole DB.

## solution

Properly call `rocksdb_column_family_handle_destroy` at earlier (but proper) timings.

For `SingleThreaded`, it's relatively simple: just define `Drop` for `ColumnFamily`. There is no api incompatibility.

For `MultiThreaded`, the story isn't that simple because of its relaxed lifetime relation between DB and CFs. So, I had to wrap BoundColumnFamily with Arc. (I added small assertion for this concern in `test_no_leaked_column_family` as well) So, I had to introduce api incompatibility to fix this sadly.

Although there is unavoidable api breakage, I approached the problem under the constraint of providing some kind of source-level compatibility code for downstream consumers for the sake of general simpler code maintenace and generic programming. Specifically, leveraging the compiler's auto ref resolution thing, I realized that property using `&&ColumnFamily` for the single threaded. So, all tests compiles adding another extra `&` at call side for the both MultiThreaded and SingleThraded.

Also note that all of existing code should compile as before for default crate feature setting (no `multi-threaded-cf`). I just added `&` across the board to make the test suite compile for the both crate configuration.

As for performance, there is no overhead in SingleThreaded. and there is small Arc one for MultiThreaded.

## (for the record)

Below is draft pr description: 

> first of all, thanks for releasing 0.16.0!!!!
> 
> this pr is still in draft phase. don't waste your time before I properly document things...
> 
> this is very unfortunate but, and after our further testing, we found we need another bugfix: https://github.com/solana-labs/solana/issues/16589


src/db.rs
```diff
@@ -1716,16 +1717,10 @@ impl DBWithThreadMode<MultiThreaded> {
...
-        self.cfs
-            .cfs
-            .read()
-            .unwrap()
-            .get(name)
-            .map(|cf| BoundColumnFamily {
-                inner: cf.inner,
-                multi_threaded_cfs: PhantomData,
-            })
+    pub fn cf_handle(&self, name: &str) -> Option<Arc<BoundColumnFamily>> {
```
ryoqun, line 1720: @carllin I think we can avoid the handler approch you mentioned in this way?

1. take `Arc<CF>`
2. call `DB::drop_cf()`
3. pass the `Arc<CF>` to the background cf cleaner thread
4. the thread waits until `Arc::strong_count == 1`.
5. drop!

tests/test_db.rs
```diff
@@ -536,16 +536,30 @@ fn test_open_cf_with_ttl() {
...
-    let cf = db.cf_handle("test_cf").unwrap();
-    db.put_cf(cf, b"key1", b"value1").unwrap();
-
-    thread::sleep(Duration::from_secs(2));
-    // Trigger a manual compaction, this will check the TTL filter
-    // in the database and drop all expired entries.
-    db.compact_range_cf(cf, None::<&[u8]>, None::<&[u8]>);
-
-    assert!(db.get_cf(cf, b"key1").unwrap().is_none());
+    let outlived_db_or_cf = {
```
ryoqun, line 539: these are just free-riding on unrelated test..

I'll write a proper one.

ryoqun:
long time no see. I'm going to revisit this shortly.

ryoqun:
Hi, I worked on this finally. it took some time write a proper test to detect leak... I'll clean up this wip branch tomorrow: https://github.com/rust-rocksdb/rust-rocksdb/compare/master...ryoqun:fix-dropped-cf-leak-wip?expand=1

ryoqun:
@aleksuss hi, everythihg is done! :) could you review this?

aleksuss:
@ryoqun yes, sure. Thanks for your contribution

src/lib.rs
```diff
@@ -102,7 +102,7 @@ pub use crate::{
         ColumnFamilyRef, DEFAULT_COLUMN_FAMILY_NAME,
     },
     compaction_filter::Decision as CompactionDecision,
-    db::{DBWithThreadMode, LiveFile, MultiThreaded, SingleThreaded, DB},
+    db::{DBWithThreadMode, LiveFile, MultiThreaded, SingleThreaded, ThreadMode, DB},
```
ryoqun, line 105: I think this one line is enough for https://github.com/rust-rocksdb/rust-rocksdb/issues/532

ryoqun:
fyi, this is how you can write generic-over-`ThreadMode` or `{multi,single}-threaded-cf`crate feature agnostic code: https://github.com/solana-labs/solana/compare/master...ryoqun:rocksdb-multi-threaded?expand=1

tests/test_column_family.rs
```diff
@@ -250,3 +250,66 @@ fn test_create_duplicate_column_family() {
...
+            db.drop_cf(&cf_name).unwrap();
+
+            #[cfg(feature = "multi-threaded-cf")]
+            {
+                outlived_cf = Some(cf);
+            }
+        }
+
+        // if we're not leaking, the dir bytes should be well under 10M bytes in total
+        let dir_bytes = fs_extra::dir::get_size(&n).unwrap();
```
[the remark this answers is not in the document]
ryoqun: okay, done: 201d7187c6784c3a9b138d5a6ab7699304c9f2b6

tests/util/mod.rs
```diff
@@ -3,6 +3,7 @@ use std::path::{Path, PathBuf};
 use rocksdb::{Options, DB};
 
 /// Temporary database path which calls DB::Destroy when DBPath is dropped.
+#[derive(Debug)]
```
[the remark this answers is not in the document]
ryoqun: nope. remnant from my dirty `dbg!()`-ing... 201d7187c6784c3a9b138d5a6ab7699304c9f2b6

src/column_family.rs
```diff
@@ -52,33 +52,82 @@ pub struct ColumnFamily {
...
+// but its memory will be exposed after transmute()-ing to BoundColumnFamily.
+// ColumnFamily's lifetime should be bound to DB. But, db holds cfs and cfs can't easily
+// self-reference DB as its lifetime due to rust's type system
+pub(crate) struct UnboundColumnFamily {
+    pub(crate) inner: *mut ffi::rocksdb_column_family_handle_t,
+}
+
+impl UnboundColumnFamily {
+    pub(crate) fn bound_column_family<'a>(
+        self: std::sync::Arc<Self>,
```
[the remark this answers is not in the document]
ryoqun: glad to see a code hygiene like this: 201d7187c6784c3a9b138d5a6ab7699304c9f2b6

I was just lazy...

src/db.rs
```diff
@@ -31,20 +32,20 @@ use std::ffi::{CStr, CString};
...
+use std::sync::Arc;
 use std::sync::RwLock;
 use std::time::Duration;
 
 // Marker trait to specify single or multi threaded column family alternations for DB
 // Also, this is minimum common API sharable between SingleThreaded and
 // MultiThreaded. Others differ in self mutability and return type.
 pub trait ThreadMode {
-    fn new(cf_map: BTreeMap<String, ColumnFamily>) -> Self;
+    fn new(cf_map: BTreeMap<String, *mut crate::ffi::rocksdb_column_family_handle_t>) -> Self;
```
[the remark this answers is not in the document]
ryoqun: According to `git grep`, it looks like those C types tend to be prefixed with `ffi::`. So, I followed the convension instead of `use` just for `rocksdb_column_family_handle_t` (201d718):

```
$ git grep 'rocksdb_.*_t[^_a-z0-9]' src/
src/backup.rs:    inner: *mut ffi::rocksdb_backup_engine_t,
src/backup.rs:    inner: *mut ffi::rocksdb_options_t,
src/backup.rs:    inner: *mut ffi::rocksdb_restore_options_t,
src/backup.rs:        let be: *mut ffi::rocksdb_backup_engine_t;
src/checkpoint.rs:    inner: *mut ffi::rocksdb_checkpoint_t,
src/checkpoint.rs:        let checkpoint: *mut ffi::rocksdb_checkpoint_t;
src/column_family.rs:    pub(crate) inner: *mut ffi::rocksdb_column_family_handle_t,
src/column_family.rs:    pub(crate) inner: *mut ffi::rocksdb_column_family_handle_t,
src/column_family.rs:    pub(crate) inner: *mut ffi::rocksdb_column_family_handle_t,
src/column_family.rs:fn destroy_handle(handle: *mut ffi::rocksdb_column_family_handle_t) {
src/column_family.rs:    fn inner(&self) -> *mut ffi::rocksdb_column_family_handle_t;
src/column_family.rs:    fn inner(&self) -> *mut ffi::rocksdb_column_family_handle_t {
src/column_family.rs:    fn inner(&self) -> *mut ffi::rocksdb_column_family_handle_t {
src/column_family.rs:    fn inner(&self) -> *mut ffi::rocksdb_column_family_handle_t {
src/compaction_filter_factory.rs:    unsafe fn from_raw(ptr: *mut ffi::rocksdb_compactionfiltercontext_t) -> Self {
src/compaction_filter_factory.rs:    context: *mut ffi::rocksdb_compactionfiltercontext_t,
src/db.rs:    fn new(cf_map: BTreeMap<String, *mut ffi::rocksdb_column_family_handle_t>) -> Self;
src/db.rs:    fn new(cfs: BTreeMap<String, *mut ffi::rocksdb_column_family_handle_t>) -> Self {
src/db.rs:    fn new(cfs: BTreeMap<String, *mut ffi::rocksdb_column_family_handle_t>) -> Self {
src/db.rs:        cfopts: &[*const ffi::rocksdb_options_t],
src/db.rs:        cfhandles: &mut Vec<*mut ffi::rocksdb_column_family_handle_t>,
src/db.rs:    ) -> Result<*mut ffi::rocksdb_column_family_handle_t, Error> {
src/db.rs:            // rocksdb_wal_readoptions_t does not appear to have any functions
src/db.rs:            let opts: *const ffi::rocksdb_wal_readoptions_t = ptr::null();
src/db.rs:        cf_inner: *mut ffi::rocksdb_column_family_handle_t,
src/db_iterator.rs:    inner: *mut ffi::rocksdb_iterator_t,
src/db_iterator.rs:        cf_handle: *mut ffi::rocksdb_column_family_handle_t,
src/db_iterator.rs:        cf_handle: *mut ffi::rocksdb_column_family_handle_t,
src/db_iterator.rs:    pub(crate) inner: *mut ffi::rocksdb_wal_iterator_t,
src/db_options.rs:fn new_cache(capacity: size_t) -> *mut ffi::rocksdb_cache_t {
src/db_options.rs:    pub(crate) inner: *mut ffi::rocksdb_cache_t,
src/db_options.rs:    inner: *mut ffi::rocksdb_env_t,
src/db_options.rs:    pub(crate) inner: *mut ffi::rocksdb_options_t,
src/db_options.rs:    pub(crate) inner: *mut ffi::rocksdb_writeoptions_t,
src/db_options.rs:    pub(crate) inner: *mut ffi::rocksdb_flushoptions_t,
src/db_options.rs:    pub(crate) inner: *mut ffi::rocksdb_block_based_table_options_t,
src/db_options.rs:    pub(crate) inner: *mut ffi::rocksdb_readoptions_t,
src/db_options.rs:    pub(crate) inner: *mut ffi::rocksdb_cuckoo_table_options_t,
src/db_options.rs:    pub(crate) inner: *mut ffi::rocksdb_ingestexternalfileoptions_t,
src/db_options.rs:        unsafe { ffi::rocksdb_block_based_options_set_data_block_index_type(self.inner, index_t) }
src/db_options.rs:            .map(|path| path.inner as *const ffi::rocksdb_dbpath_t)
src/db_options.rs:            ffi::rocksdb_readoptions_set_readahead_size(self.inner, v as size_t);
src/db_options.rs:    pub(crate) inner: *mut ffi::rocksdb_fifo_compaction_options_t,
src/db_options.rs:    pub(crate) inner: *mut ffi::rocksdb_universal_compaction_options_t,
src/db_options.rs:    pub(crate) inner: *mut ffi::rocksdb_compactoptions_t,
src/db_options.rs:    pub(crate) inner: *mut ffi::rocksdb_dbpath_t,
src/db_pinnable_slice.rs:    ptr: *mut ffi::rocksdb_pinnableslice_t,
src/db_pinnable_slice.rs:    pub(crate) unsafe fn from_c(ptr: *mut ffi::rocksdb_pinnableslice_t) -> DBPinnableSlice<'a> {
src/perf.rs:    pub(crate) inner: *mut ffi::rocksdb_perfcontext_t,
src/perf.rs:    inner: *mut ffi::rocksdb_memory_usage_t,
src/perf.rs:    inner: *mut ffi::rocksdb_memory_consumers_t,
src/slice_transform.rs:    pub inner: *mut ffi::rocksdb_slicetransform_t,
src/snapshot.rs:    pub(crate) inner: *const ffi::rocksdb_snapshot_t,
src/sst_file_writer.rs:    pub(crate) inner: *mut ffi::rocksdb_sstfilewriter_t,
src/sst_file_writer.rs:    inner: *mut ffi::rocksdb_envoptions_t,
src/sst_file_writer.rs:    fn create_raw(opts: &Options, env_opts: &EnvOptions) -> *mut ffi::rocksdb_sstfilewriter_t {
src/write_batch.rs:    pub(crate) inner: *mut ffi::rocksdb_writebatch_t,

```

ryoqun:
@aleksuss thanks for review! I've cleaned this pr up accordingly.

src/db.rs
```diff
@@ -31,20 +32,20 @@ use std::ffi::{CStr, CString};
...
+use std::sync::Arc;
 use std::sync::RwLock;
 use std::time::Duration;
 
 // Marker trait to specify single or multi threaded column family alternations for DB
 // Also, this is minimum common API sharable between SingleThreaded and
 // MultiThreaded. Others differ in self mutability and return type.
 pub trait ThreadMode {
-    fn new(cf_map: BTreeMap<String, ColumnFamily>) -> Self;
+    fn new(cf_map: BTreeMap<String, *mut ffi::rocksdb_column_family_handle_t>) -> Self;
```
aleksuss, line 48: Is it really good idea to use `ffi` types in public api ?

src/column_family.rs
```diff
@@ -52,33 +54,82 @@ pub struct ColumnFamily {
...
+}
+
 impl<'a> AsColumnFamilyRef for &'a ColumnFamily {
     fn inner(&self) -> *mut ffi::rocksdb_column_family_handle_t {
         self.inner
     }
 }
 
-impl<'a> AsColumnFamilyRef for BoundColumnFamily<'a> {
+impl<'a> AsColumnFamilyRef for Arc<BoundColumnFamily<'a>> {
```
aleksuss, line 132:
```suggestion
impl<'a> AsColumnFamilyRef for ColumnFamilyRef<'a> {
```
?

src/db.rs
```diff
@@ -64,36 +65,39 @@ pub struct SingleThreaded {
...
-            }
-        }
+        // Cause all ColumnFamily objects to be Drop::drop()-ed.
+        self.cfs.clear();
     }
 }
 
 impl ThreadMode for MultiThreaded {
-    fn new(cfs: BTreeMap<String, ColumnFamily>) -> Self {
+    fn new(cfs: BTreeMap<String, *mut ffi::rocksdb_column_family_handle_t>) -> Self {
```
aleksuss, line 88: Can't we create `ffi::rocksdb_column_family_handle_t` inside the method `new` and then a signature of the method could be changed to `fn new(cfs: BTreeSet<String>) -> Self` ? 🤔

src/db.rs
ryoqun: certainly, it's not good idea... but we have to do that... kind of. I added some reasoning here: https://github.com/rust-rocksdb/rust-rocksdb/pull/509/commits/0d5ff023b42536aee8869fac5a504b553ff0fe4d

I googled a bit but found that it's quite bothersome to hide actual trait methods from public traits...

In short, these methods aren't intended to be consumed, nevertheless both pre- and post- this pr.

src/db.rs
ryoqun: technically, yes. but I double its merit...

As I went long explaining here  (https://github.com/rust-rocksdb/rust-rocksdb/commit/0d5ff023b42536aee8869fac5a504b553ff0fe4d), this trait method is called deep inside DB construction at (`open_cf_descriptors_internal`). So, we have to move large part of code inside `ThreadMode`.. Then `ThreadMode` will be significantly fatter than now, although it's not intended to be used for users still as was before. Then this only gives us the ability of getting rid of those `ffi::` types in strictly _public_ APIs.

src/column_family.rs
ryoqun: super good eyes you have to spot these unspoken subtleties... :)

I added some missing docs: eb509fbb596c489580daa763f36c69df2a3fa548

src/db.rs
aleksuss: Should the trait `ThreadMod` be public ??

aleksuss: I got it. I've seen a description a bit later.

ryoqun:
@aleksuss Sorry for delay! I'm feeling this pr is getting polished up from your reviews. Thanks for continued review. :)

ryoqun:
> @aleksuss aleksuss requested review from @stanislav-tkach and @aleksuss 10 days ago 

:dog: (Hachikō) <(I'm waiting for further review comments if any from you)

src/db.rs
```diff
@@ -1753,46 +1788,39 @@ impl DBWithThreadMode<MultiThreaded> {
...
         }
     }
 
     /// Returns the underlying column family handle
-    pub fn cf_handle(&self, name: &str) -> Option<BoundColumnFamily> {
+    pub fn cf_handle(&self, name: &str) -> Option<Arc<BoundColumnFamily>> {
         self.cfs
             .cfs
             .read()
             .unwrap()
```
aleksuss, line 1813:
```suggestion
            .ok()?
```

src/db.rs
```diff
@@ -1753,46 +1788,39 @@ impl DBWithThreadMode<MultiThreaded> {
...
+            Arc::new(UnboundColumnFamily { inner }),
+        );
         Ok(())
     }
 
     /// Drops the column family with the given name by internally locking the inner column
     /// family map. This avoids needing `&mut self` reference
     pub fn drop_cf(&self, name: &str) -> Result<(), Error> {
-        let inner = self.inner;
         if let Some(cf) = self.cfs.cfs.write().unwrap().remove(name) {
```
aleksuss, line 1801: I'd replace here `unwrap()` for something like:
```rust
.map_err(|e| Error::new(e.to_string()))?
```

src/db.rs
```diff
@@ -1753,46 +1788,39 @@ impl DBWithThreadMode<MultiThreaded> {
     /// Creates column family with given name and options
     pub fn create_cf<N: AsRef<str>>(&self, name: N, opts: &Options) -> Result<(), Error> {
         let inner = self.create_inner_cf_handle(name.as_ref(), opts)?;
-        self.cfs
-            .cfs
-            .write()
-            .unwrap()
-            .insert(name.as_ref().to_string(), ColumnFamily { inner });
+        self.cfs.cfs.write().unwrap().insert(
```
aleksuss, line 1791: `unwrap()` -> `.map_err(|e| Error::new(e.to_string()))?` ?

src/db.rs
```diff
@@ -58,47 +70,56 @@ pub struct SingleThreaded {
...
 
-    fn cf_drop_all(&mut self) {
-        for cf in self.cfs.read().unwrap().values() {
-            unsafe {
-                ffi::rocksdb_column_family_handle_destroy(cf.inner);
-            }
-        }
+    fn drop_all_cfs_internal(&mut self) {
+        // Cause all UnboundColumnFamily objects to be Drop::drop()-ed.
+        self.cfs.write().unwrap().clear();
```
aleksuss, line 115: `unwrap()` -> `expect("Couldn't acquire a write lock")` ?

aleksuss (commented):
Sorry for long intervals between reviews. Too busy lately.

src/db.rs
ryoqun: @aleksuss hi! thanks for continued review while you're very busy... :)

in short, I think RwLock's {`.read()`/`.write()`} should be paired with `.unwrap()`.  This topic is called _lock poisoning_:

https://doc.rust-lang.org/nomicon/poisoning.html:

> However if the Mutex contained, say, a BinaryHeap that does not actually have the heap property, it's unlikely that any code that uses it will do what the author intended. As such, the program should not proceed normally. Still, if you're double-plus-sure that you can do something with the value, the Mutex exposes a method to get the lock anyway. It is safe, after all. Just maybe nonsense.

(Replace `BinaryHeap` here with our case: `BTreeMap`.)


Also, some relevant info:

- https://blog.rust-lang.org/2020/12/11/lock-poisoning-survey.html
- https://hackmd.io/@rust-libs/BJZxn8a3D#Which-of-these-is-closest-to-how-you-acquire-locks-Select-all-that-apply

Also, I've quickly skimmed wild usage (using our monorepo as sample..), too:

image

also, notice that rocksdb appears in the above screenshot, because I did `.unwrap()` at my previous pr like this: https://github.com/rust-rocksdb/rust-rocksdb/pull/506/files#r674132675

src/db.rs
ryoqun: thanks for carefully-thought suggestion.

However, according to https://doc.rust-lang.org/std/sync/struct.RwLock.html#errors-2, RwLock::write() return Err if and only if it's poisoned.

> This function will return an error if the RwLock is poisoned. An RwLock is poisoned whenever a writer panics while holding an exclusive lock. An error will be returned when the lock is acquired.

Also, RwLock::write() is blocking indefinitely. so your suggested message could be confusing and just better off propagating the RwLock Err as-is with `.unwrap()`?:

> Locks this rwlock with exclusive write access, blocking the current thread until it can be acquired.
> This function will not return while other writers or other readers currently have access to the lock.

src/db.rs
ryoqun: see my comment: https://github.com/rust-rocksdb/rust-rocksdb/pull/509/files#r674138005

src/db.rs
ryoqun: see my comment: https://github.com/rust-rocksdb/rust-rocksdb/pull/509/files#r674138005

src/db.rs
ryoqun: oops the above comment assumes you read my other comment here: https://github.com/rust-rocksdb/rust-rocksdb/pull/509/files#r674138005

ryoqun:
> Sorry for long intervals between reviews. Too busy lately.

@aleksuss hi! I've replied to your comments. please have a look when convenient. thanks!

aleksuss (approved):

ryoqun:
@aleksuss really thanks very much for merging and prepareing 0.17.0. I noticed it already #539 . :)

Issue #502, opened 2021-03-03
multi_get memory corruption


mikhailOK:
`multi_get_opt` incorrectly calls the `ffi::rocksdb_multi_get`:

ffi.rs:
```
            let keys: [*const c_char; 3] = [cstrp!("box"), cstrp!("foo"), cstrp!("notfound")];
            let keys_sizes: [size_t; 3] = [3, 3, 8];
            let mut vals: [*mut c_char; 3] = [ptr::null_mut(), ptr::null_mut(), ptr::null_mut()];
            let mut vals_sizes: [size_t; 3] = [0, 0, 0];
            let mut errs: [*mut c_char; 3] = [ptr::null_mut(), ptr::null_mut(), ptr::null_mut()];
            rocksdb_multi_get(
                db,
                roptions,
                3,
                keys.as_ptr(),
                keys_sizes.as_ptr(),
                vals.as_mut_ptr(),
                vals_sizes.as_mut_ptr(),
                errs.as_mut_ptr(),
            );
```
multi_get_opt:
```
        let mut values = vec![ptr::null_mut(); keys.len()];
        let mut values_sizes = vec![0_usize; keys.len()];
        unsafe {
            ffi_try!(ffi::rocksdb_multi_get(
                self.inner,
                readopts.inner,
                ptr_keys.len(),
                ptr_keys.as_ptr(),
                keys_sizes.as_ptr(),
                values.as_mut_ptr(),
                values_sizes.as_mut_ptr(),
            ));
        }
```
ffi_try:
```
macro_rules! ffi_try_impl {
    ( $($function:ident)::*( $($arg:expr,)*) ) => {{
        let mut err: *mut ::libc::c_char = ::std::ptr::null_mut();
        let result = $($function)::*($($arg,)* &mut err);
        if !err.is_null() {
            return Err(Error::new($crate::ffi_util::error_message(err)));
        }
        result
    }};
}
```
`ffi_try` works with a single error, but `rocksdb_multi_get` expects an array of errors.

stanislav-tkach:
Thank you for the report! Would you mind submitting a pull request?

PR #511, opened 2021-04-21
Fix multi_get

mikhailOK:
Make multi_get return Vec<Result> instead of Result<Vec>.
See #502 for details.

### DIFF
--- a/src/db.rs
+++ b/src/db.rs
@@ -747,7 +747,7 @@ impl<T: ThreadMode> DBWithThreadMode<T> {
     }
 
     /// Return the values associated with the given keys.
-    pub fn multi_get<K, I>(&self, keys: I) -> Result<Vec<Vec<u8>>, Error>
+    pub fn multi_get<K, I>(&self, keys: I) -> Vec<Result<Option<Vec<u8>>, Error>>
     where
         K: AsRef<[u8]>,
         I: IntoIterator<Item = K>,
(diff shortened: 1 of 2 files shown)


src/db.rs
```diff
@@ -1779,16 +1783,25 @@ fn convert_options(opts: &[(&str, &str)]) -> Result<Vec<(CString, CString)>, Err
         .collect()
 }
 
-fn convert_values(values: Vec<*mut c_char>, values_sizes: Vec<usize>) -> Vec<Vec<u8>> {
+fn convert_values(
+    values: Vec<*mut c_char>,
+    values_sizes: Vec<usize>,
+    errors: Vec<*mut c_char>,
+) -> Vec<Result<Option<Vec<u8>>, Error>> {
```
stanislav-tkach, line 1790: Do you think that `Option<Vec>` would be better than a potentially empty `Vec` here?

Personally, I would use `Ok(value.unwrap_or(Vec::new()))` to simplify api a little bit.

stanislav-tkach (approved):
Thank you for the pull request! It is nice that you reuse the `ffi_util::raw_data` function here, though I'm not sure about `Option<Vec>`.

src/db.rs
mikhailOK: Without `Option` there's no way to distinguish between key -> empty string and key not in the database. This way `multi_get` is consistent with `get` that returns `Result<Option<Vec>>`.

thibault-martinez:
Hello, we're currently having a `signal: 11, SIGSEGV: invalid memory reference` using `multi_get` and it may be due to #502 , do you have an idea when you intend to merge this PR ? Thanks.

aleksuss (approved):

thibault-martinez:
FYI, it did fix out SIGSEGV, thanks!

PR #519, opened 2021-05-01
Add support for the cuckoo table format

rbost:
RocksDB supports a cuckoo table format, for fast point queries on datasets with fixed key and value length. Unfortunately, the current version of `rust-rocksdb` does not support creating a new table with this format. 

This PR adds the necessary Rust APIs for that, and nothing more (as everything was already available from the ffi).

### DIFF
--- a/src/db_options.rs
+++ b/src/db_options.rs
@@ -711,23 +711,48 @@ impl Default for BlockBasedOptions {
 }
 
 impl CuckooTableOptions {
+    /// Determines the utilization of hash tables. Smaller values
+    /// result in larger hash tables with fewer collisions.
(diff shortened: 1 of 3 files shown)


stanislav-tkach (approved):
Looking good to me, but it would be nice to add at least one test that sets some options and opens the database.

rbost:
> Looking good to me, but it would be nice to add at least one test that sets some options and opens the database.

Done in bebddea. It is a very basic test though. Is it good enough for you?

stanislav-tkach:
Yes, I think it is enough, thank you.

aleksuss (approved):

PR #526, opened 2021-05-28
Make SSE inclusion conditional for target features

mbargull:
Since `0.12.3`/gh-295 the underlying library gets compiled with SSE4.1/SSE4.2/PCLMUL instructions unconditionally.
This prevents building the library for backwards compatible binaries, e.g., for `target-cpu=nocona` you'd want a binary without SSE4 instructions.
With the changes proposed here, one still can freely build for `target-cpu=nehalem` (or `target-cpu=westmere` to get PCLMUL), or just `target-cpu=native` given a recent enough machine and for local deployment. And additionally `target-cpu=nocona` would yield the desired compatible build.

(Context here is that `rust-rocksdb` is used by some programs we distribute as binaries for [Bioconda](https://bioconda.github.io) and currently/usually target older CPU architectures there.)

### DIFF
--- a/librocksdb-sys/build.rs
+++ b/librocksdb-sys/build.rs
@@ -103,14 +103,24 @@ fn build_rocksdb() {
         // This is needed to enable hardware CRC32C. Technically, SSE 4.2 is
         // only available since Intel Nehalem (about 2010) and AMD Bulldozer
         // (about 2011).
-        config.define("HAVE_SSE42", Some("1"));
-        config.flag_if_supported("-msse2");
-        config.flag_if_supported("-msse4.1");
-        config.flag_if_supported("-msse4.2");
+        let target_feature = env::var("CARGO_CFG_TARGET_FEATURE").unwrap();
+        let target_features: Vec<_> = target_feature.split(",").collect();
+        if target_features.contains(&"sse2") {
+            config.flag_if_supported("-msse2");
+        }
+        if target_features.contains(&"sse4.1") {
+            config.flag_if_supported("-msse4.1");
+        }
+        if target_features.contains(&"sse4.2") {
+            config.flag_if_supported("-msse4.2");
+            config.define("HAVE_SSE42", Some("1"));
+        }
 
         if !target.contains("android") {
-            config.define("HAVE_PCLMUL", Some("1"));
-            config.flag_if_supported("-mpclmul");
+            if target_features.contains(&"pclmulqdq") {
+                config.define("HAVE_PCLMUL", Some("1"));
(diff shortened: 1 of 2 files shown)


stanislav-tkach (approved):

librocksdb-sys/build.rs
```diff
@@ -103,14 +103,24 @@ fn build_rocksdb() {
         // This is needed to enable hardware CRC32C. Technically, SSE 4.2 is
         // only available since Intel Nehalem (about 2010) and AMD Bulldozer
         // (about 2011).
-        config.define("HAVE_SSE42", Some("1"));
-        config.flag_if_supported("-msse2");
-        config.flag_if_supported("-msse4.1");
-        config.flag_if_supported("-msse4.2");
+        let target_feature = env::var("CARGO_CFG_TARGET_FEATURE").unwrap();
+        let target_features: Vec<_> = target_feature.split(",").collect();
+        if target_features.contains(&"sse2") {
```
akrylysov, line 108: Does it change the behavior to not enable sse2 etc by default? If that's the case, CHANGELOG.md should be updated accordingly so developers are aware about the performance degradation once they upgrade the rocksdb crate.

mbargull: I believe SSE2 is in the set of default features for x86-64, at least it is for me if I run
```sh

Comment (akrylysov) [reply librocksdb-sys/build.rs:108]: Maybe something like "RocksDB is not compiled with SSE4 instructions anymore unless the corresponding features are enabled in rustc"?

File: librocksdb-sys/build.rs [hunk 2]
```diff
@@ -103,14 +103,24 @@ fn build_rocksdb() {
         // This is needed to enable hardware CRC32C. Technically, SSE 4.2 is
         // only available since Intel Nehalem (about 2010) and AMD Bulldozer
         // (about 2011).
-        config.define("HAVE_SSE42", Some("1"));
-        config.flag_if_supported("-msse2");
-        config.flag_if_supported("-msse4.1");
-        config.flag_if_supported("-msse4.2");
+        let target_feature = env::var("CARGO_CFG_TARGET_FEATURE").unwrap();
+        let target_features: Vec<_> = target_feature.split(",").collect();
```
aleksuss, line 107:
```suggestion
        let target_features: Vec<_> = target_feature.split(',').collect();
```
Of course, it's not critical here. But I think the better way to use `char` rather than `&str` as a pattern when you use one character for splitting.

aleksuss (approved):
Overall LGTM 👍🏻

PR #531, opened 2021-06-16
Upgrade librocksdb-sys to 6.20.3

akrylysov:
Upgrades librocksdb-sys to 6.20.3.

### DIFF
--- a/librocksdb-sys/build.rs
+++ b/librocksdb-sys/build.rs
@@ -175,6 +175,8 @@ fn build_rocksdb() {
         lib_sources.push("port/win/win_thread.cc");
     }
 
+    config.define("ROCKSDB_SUPPORT_THREAD_LOCAL", None);
+
     if target.contains("msvc") {
         config.flag("-EHsc");
     } else {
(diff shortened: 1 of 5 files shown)


librocksdb-sys/build.rs
```diff
@@ -175,6 +175,8 @@ fn build_rocksdb() {
         lib_sources.push("port/win/win_thread.cc");
     }
 
+    config.define("ROCKSDB_SUPPORT_THREAD_LOCAL", None);
```
akrylysov, line 178: More context https://github.com/facebook/rocksdb/pull/8117

ryoqun:
(I want this to be included for the next release (v0.17.0) along side #509, especially because of this bugfix in 6.20.2:

https://github.com/facebook/rocksdb/blob/v6.20.3/HISTORY.md#bug-fixes-1

> Fixed the false-positive alert when recovering from the WAL file. Avoid reporting "SST file is ahead of WAL" on a newly created empty column family, if the previous WAL file is corrupted.

aleksuss (approved):

PR #539, opened 2021-07-22
Release 0.17.0

aleksuss:


stanislav-tkach (approved):
I'm concerned about the "reference to packed field is unaligned" warnings. As far as I can see, such warnings are only triggered in tests, but that is still worth investigating.

PR #544, opened 2021-07-27
add feature flag for runtime type information, and metadata for locating library and sources

jgraettinger:
We have an advanced use case with [Estuary Flow](github.com/estuary/flow) where we're interoperating Go and Rust within a single process, and wish to use this crate to actually build RocksDB, while linking it using the `go` toolchain.

We've gotten this working by having the `librocksdb-sys` crate publish metadata which dependent crates can utilize to locate the RocksDB include sources and the libraries built by this crate.

Our use case also requires the RTTI compile-time flag of RocksDB, which is disabled by default. I've added this as a feature flag.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -24,6 +24,7 @@ lz4 = ["librocksdb-sys/lz4"]
 zstd = ["librocksdb-sys/zstd"]
 zlib = ["librocksdb-sys/zlib"]
 bzip2 = ["librocksdb-sys/bzip2"]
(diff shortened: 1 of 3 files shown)


stanislav-tkach (approved):

jgraettinger:
I'm unclear on the approval workflow, and clicked the "Update from master branch" in the Github UI. Unsure if I was supposed to do that 😁

aleksuss (approved):

PR #545, opened 2021-07-27
add Env::from_raw constructor

jgraettinger:
Support usages where the caller already has an environment built by an
alternative means, which the Env of this crate should take ownership of.


stanislav-tkach (approved):

src/db_options.rs
```diff
@@ -127,6 +127,13 @@ impl Env {
         }
     }
 
+    /// Returns a new environment which wraps and takes ownership of the provided
+    /// raw environment.
+    // Marked unsafe because ownership is transferred to the returned Env.
+    pub unsafe fn from_raw(env: *mut ffi::rocksdb_env_t) -> Self {
```
aleksuss, line 133: Couple of questions.
1. Why ?
2. What is gonna be a caller of this method ?
3. Where are tests for this method ?
4. And we shouldn't forget about such [case](https://play.rust-lang.org/?version=stable&mode=debug&edition=2018&gist=e5feb2a24535644c715cadf18651fd93) ?

jgraettinger: 1) The rocks environment provides mechanisms to hook into lower-level subsystems of Rocks, such as file IO. For example the [rocks gazette consumer store](https://gazette.readthedocs.io/en/latest/consumers-concepts.html#stores) leverages this to replicate databases states as they're being written. You're deep into FFI land to do this type of thing, of course, but once you've _built_ an instrumented FFI environment it's nice to hand it to this crate for usage.

2) [Example call site](https://github.com/estuary/flow/blob/master/crates/derive/src/derive_api.rs#L76-L80). You have to already have an FFI *env pointer of course. This PR merely plumbs through a means to hand it to an Env of this crate, and as best I can tell matches the standard conventions for exposing and naming an unsafe operation of this kind in the standard library and elsewhere.

3) I'm not sure what it would cover, that the compiler doesn't check already. If there were a paired `into_raw()` function then a test case might exercise round-tripping through that raw pointer, but I don't get the feeling an `into_raw` is desired and I'm not suggesting it. (FWIW there's of course plenty of integrated, end-to-end coverage in our repo [that uses this API](https://github.com/estuary/flow)).

4) That's why this function is unsafe, and specifically notes it's unsafe because ownership is transferred, similarly to how `Vec::from_raw_parts` and other functions of this kind are documented.

stanislav-tkach (approved):
I think it would be better to extend the "why it is unsafe" section with a better explanation how this function should be used safely. I would prefer to follow the std approach in describing such things. `Vec::set_len` description can be used as the example.

jgraettinger:
> I think it would be better to extend the "why it is unsafe" section

After a short multi-year delay :grin: I've extended this section with more discussion.

jgraettinger:
Bump @stanislav-tkach ? Thanks

stanislav-tkach:
Thanks for the update, but I have already approved the pull request. 🙃 

@zaidoon1 What do you think?

aleksuss:
@jgraettinger could you please add any test or example of usage? But I don't like to expose **unsafe** public methods.

jgraettinger:
> @jgraettinger could you please add any test or example of usage? But I don't like to expose unsafe public methods.

🤔 I don't know how we'd realistically do that without _dramatically_ blowing up the scope of this change. For example, it would require new FFI bridge code to get a raw Env _out_ from this crate -- which would be silly, considering the motivating use case is having an Env that's wired up with C++ handlers that the C API doesn't even have access too.

Open to ideas.

Here's an actual call site in a real application: https://github.com/estuary/flow/blob/master/crates/runtime/src/rocksdb.rs#L236-L243

An example docstring in this context could just be:
```
let mut opts = rocksdb::Options::default();

// Re-hydrate the provided memory address into rocksdb::Env wrapping
// an owned *mut librocksdb_sys::rocksdb_env_t.
let env = unsafe {
    rocksdb::Env::from_raw(rocksdb_env_memptr as *mut librocksdb_sys::rocksdb_env_t)
};
opts.set_env(&env);
```

Is this what you had in mind?

>  But I don't like to expose unsafe public methods.

I mean, this _is_ a crate that's wrapping a highly customize-able, performance critical C++ library. It comes with the territory.

zaidoon1:
LGTM. I consider this for advanced users/special use cases so we can just assume the users know what they are doing/are willing to take the risk of this unsafe binding.

zaidoon1:
once builds pass, I'll merge

zaidoon1 (approved):

PR #551, opened 2021-08-25
Support UTF-8 file paths on Windows

### LINKED ISSUE #85: [Windows] Unicode path issue

### LINKED ISSUE #512: Unicode path names issue on Windows reappeared
It seems like the `ROCKSDB_WINDOWS_UTF8_FILENAMES` flag was removed in https://github.com/rust-rocksdb/rust-rocksdb/pull/493. This flag fixed #85 and allowed paths with non-ANSI characters. Would it be possible to re-add it, or was there a specific reason it was removed?

rajivshah3:
This re-adds support for UTF-8 file paths on Windows, which was removed in #493 

Fixes #512 
Fixes #85 (already closed)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,7 @@
 # Changelog
 
 ## [Unreleased]
+* Re-add support for UTF-8 file paths on Windows (rajivshah3)
 
 ## 0.17.0 (2021-07-22)
 
--- a/librocksdb-sys/build.rs
+++ b/librocksdb-sys/build.rs
@@ -152,6 +152,7 @@ fn build_rocksdb() {
         config.define("_MBCS", None);
         config.define("WIN64", None);
         config.define("NOMINMAX", None);
+        config.define("WITH_WINDOWS_UTF8_FILENAMES", "ON");
 
         if &target == "x86_64-pc-windows-gnu" {
             // Tell MinGW to create localtime_r wrapper of localtime_s function.


stanislav-tkach (approved):
As far as I can see, previously the `ROCKSDB_WINDOWS_UTF8_FILENAMES` option was set and now you have added the different (`WITH_WINDOWS_UTF8_FILENAMES`) option. Are the any differences between them?

rajivshah3:
@stanislav-tkach sorry, I forgot to clarify that. I think they're basically the same — setting `WITH_WINDOWS_UTF8_FILENAMES` to `ON` adds the `-DROCKSDB_WINDOWS_UTF8_FILENAMES` flag (https://github.com/facebook/rocksdb/blob/8608d75d85f8e1b3b64b73a4fb6d19baec61ba5c/CMakeLists.txt#L69-L72). I can change this if you'd like

aleksuss (approved):

rajivshah3:
Thanks @stanislav-tkach @aleksuss ! Would it be possible to make a new release with these changes? We'd like to use this fix and a few others in our projects

PR #556, opened 2021-09-20
Derive `serde::{Serialize, Deserialize}` for configuration enums

thibault-martinez:
This is useful to be able to use them to/from configuration files. At the moment we have to bypass by replicating the enums like https://github.com/iotaledger/bee/blob/coordicide/bee-storage/bee-storage-rocksdb/src/compression.rs.
This is under a feature flag so opt-in and no added dependency by default.

Also remove some warnings in the tests.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -26,10 +26,12 @@ zstd = ["librocksdb-sys/zstd"]
 zlib = ["librocksdb-sys/zlib"]
 bzip2 = ["librocksdb-sys/bzip2"]
 multi-threaded-cf = []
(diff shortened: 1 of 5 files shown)


stanislav-tkach (approved):
Sorry for the really late reply and thanks for the pull request! I think it is a nice improvement and hiding it under the feature is also a good idea.

thibault-martinez:
Hi @stanislav-tkach thanks for the review. I just fixed the CI.

aleksuss (approved):

Issue #560, opened 2021-10-19
Add more BackupEngine APIs


goldenleaves:
1. The current BackupEngine in src/backup.rs can only restore from the latest (`fn restore_from_latest_backup()`) and miss a method to restore from a specified backup id. As we can see, the rocksdb ffi `rocksdb_backup_engine_restore_db_from_backup` is provided in librocksdb-sys (rocksdb/include/rocksdb/c.h), can we add this method impl to BackupEngine?

2. Currenty we can only open a backup engine with default option, similarly rocksdb ffi provides `rocksdb_backup_engine_open_opts` and other setting funcs ( e.g. `rocksdb_backupable_db_options_set_max_background_operations`) to create a backup engine in a customized way, can we impl these for BackupEngine in rust-rocksdb?

3. Currently rocksdb C++ APIs can support deleting a specific backup with its backup id and backup gc, etc., but these are missing in rocksdb FFIs, so I wonder if we can report this issue to the rocksdb community to urge them to increase these FFIs, so as to further improve the function of BackupEngine in rust-rocksdb.

PR #561, opened 2021-10-20
Support restoring from a specified backup

GoldenLeaves:
The current BackupEngine in src/backup.rs can only restore from the latest (`fn restore_from_latest_backup()`) and miss a method to restore from a specified backup id. As we can see, the rocksdb ffi `rocksdb_backup_engine_restore_db_from_backup` is provided in librocksdb-sys.

This pull request makes it to support restoring from a specified backup.

Related to issue #560

### DIFF
--- a/src/backup.rs
+++ b/src/backup.rs
@@ -170,6 +170,55 @@ impl BackupEngine {
         Ok(())
     }
 
+    /// Restore from a specified backup
+    ///
+    /// # Arguments
+    ///
+    /// * `db_dir` - A path to the database directory
+    /// * `wal_dir` - A path to the wal directory
+    /// * `opts` - Restore options
+    /// * `backup_id` - The backup id to restore
+    pub fn restore_from_backup<D: AsRef<Path>, W: AsRef<Path>>(
+        &mut self,
+        db_dir: D,
+        wal_dir: W,
+        opts: &RestoreOptions,
+        backup_id: u32,
+    ) -> Result<(), Error> {
+        let db_dir = db_dir.as_ref();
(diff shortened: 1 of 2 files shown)


src/backup.rs
```diff
@@ -170,6 +170,55 @@ impl BackupEngine {
...
     }
 
+    /// Restore from a specified backup
+    ///
+    /// # Arguments
+    ///
+    /// * `db_dir` - A path to the database directory
+    /// * `wal_dir` - A path to the wal directory
+    /// * `opts` - Restore options
+    /// * `backup_id` - The backup id to restore
```
stanislav-tkach, line 180: Personally I don’t think that this description of the arguments is helpful. The names of the arguments are pretty self-explanatory. Perhaps it makes sense to extend the method documentation, but I don’t have a concrete suggestion.

tests/test_backup.rs
```diff
@@ -26,38 +26,72 @@ use util::DBPath;
 fn backup_restore() {
```
stanislav-tkach, line 26: I would rather leave the existing test as it is and add a new one, so that we have two tests: one for the `restore_from_latest_backup` method and another one for `restore_from_backup`.

stanislav-tkach (approved):
Thanks, only a few nitpicks.

tests/test_backup.rs
GoldenLeaves: oka~

src/backup.rs
GoldenLeaves: replace it with a comment

aleksuss (approved):

Issue #529, opened 2021-06-15
When will rust-rocksdb support transaction?


YjyJeff:
I need to use transaction in rocksdb, but I find this crate does not support it yet. When will the transaction be supported?

tv42:
It seems this can be closed now. https://docs.rs/rocksdb/latest/rocksdb/struct.Transaction.html

Issue #144, opened 2017-10-19
TxnDB : Transactions via rocksdb_transaction_t


rrichardson:
Transaction functionality is exposed in rocksdb via a parallel implementation of the `rocksdb_t` called `rocksdb_transaction_t`  .  Now that we have bindgen (#143) we get access to it for "free".  

First, for brevity, I propose a new acronym that is the RocksDB parallel of CRUD, I call it 
PIGMED , that's Put Iterate MErge Delete.  Not catchy, I know.  

The question is how to best expose the Transaction functionality while : 
1. Not duplicating the entire DB impl  for TxnDB
2. Maintaining type safety. 

There are quite a few new methods for TxnDb that would need to be wrapped, some of them might make sense as guards or even closures.  

I don't think there is a way to unify the PIGMED operations of DB and TxnDB while retaining type safety or even ergonomics.   But there isn't actually a ton of PIGMED functionality in DB.  

I propose we move all of the utility/environment functionality, that is anything that doesn't require a `rocksdb_t` into a separate utility module or modules (e.g. `rocksdb::cf::create`  for `rocksdb_create_cf`) 

We then slim down DB to just the PIGMED operations. 

My hope is that we would be able to leverage the same Iterator functionality.  There is a `rocksdb_transaction_create_iterator` but the `iterator_t` is the same. 

Thoughts?

prasannavl:
Looks like this has been there for a while. Just thought I'll drop a quick to see if there's any update on this? I'm holding off using rocksdb for some rust services due to transactions.

iSynaptic:
I reread @rrichardson  earlier comments and I think this issue is referring to the "pessimistic" version of transactional RocksDB, whereas #250 is adding the "optimistic" version.  I think we should be able to land both, leveraging much of the work in #250.  

I'm currently working on some internal restructuring in #268 that will make it easier to implement both `TransactionalDB` and `OptimisticTransactionalDB` without a lot of code duplication.

iSynaptic:
There are some missing methods in C API to make `TransactionalDB` roughly equivalent `DB` (see facebook/rocksdb#4999). That needs to be resolved before we can land this feature.

iSynaptic:
@hnakamur submitted facebook/rocksdb#5077 to add the missing APIs. Once they become available in a release, I'll work on adding support for it here in the Rust binding.

prasannavl:
Nice. Thanks for the all the great work! :)

[...]

PR #565, opened 2021-11-09
Support RocksDB transaction.

yiyuanliu:
We are building a project which uses RocksDB transaction. However, `rust-rocksdb` doesn't support transaction. PR #250 has not been updated in a long time and does not appear to be maintained.

This PR wants to provide support for transactions with minimal impact on existing code. I have finished wrapping RocksDB TransactionDB, and the basic functionality is tested. The next step is to provide support for Optimistic TransactionDB.

Related to: #529 #144


src/write_batch.rs
```diff
@@ -224,6 +224,8 @@ impl WriteBatch {
     /// Removes the database entries in the range ["begin_key", "end_key"), i.e.,
     /// including "begin_key" and excluding "end_key". It is not an error if no
     /// keys exist in the range ["begin_key", "end_key").
+    ///
+    /// `TxnDB` doesn't support `delete_range`.
```
ryoqun, line 228: what happens `TxnDB` actually used `delete_range`? the librocksdb emits an runtime error unconditionally?

ryoqun: if possible, `WriteBatch<D = DB>` and `WriteBatch<TxnDB>` might be more friendly?

src/txn/txn_db.rs
```diff
@@ -0,0 +1,729 @@
...
+    ReadOptions, SnapshotWithThreadMode, Txn, TxnDBOptions, TxnOptions, WriteBatch, WriteOptions,
+    DB, DEFAULT_COLUMN_FAMILY_NAME,
+};
+use libc::{c_char, c_int, size_t};
+
+/// RocksDB TransactionDB.
+///
+/// Please read the official [guide](https://github.com/facebook/rocksdb/wiki/Transactions)
+/// to learn more about RocksDB TransactionDB.
+pub struct TxnDB {
```
ryoqun, line 39: I'm not strongly opposed to the abbreviated `TxnDB`, but if possible, i like `TransactionDB` (not `Transactiondb` nor `TransactionDb`), especially if the upstream library isn't using `txndb`.

@aleksuss what do you think about this?

src/txn/txn_db.rs
```diff
@@ -0,0 +1,729 @@
...
+    pub(crate) inner: *mut ffi::rocksdb_transactiondb_t,
+    cfs: RwLock<BTreeMap<String, Arc<UnboundColumnFamily>>>,
```
ryoqun, line 41: i know this is a bit of stretch for you, but could you try to follow the `SingleThreaded` and `MultiThreaded` mechanism for transaction db as well? so that `cfs` doesn't necessarily need to be wrapped inside `RwLock`?

ryoqun (commented):
nice work after initial quick glance! it's been some time since i missed this functionality. thanks for contributing.

src/txn/txn_db.rs
aleksuss: @ryoqun 100% agree. `TransactionDB` would be better.

yiyuanliu: Ok, I will rename `Txn` to `Transaction`, and `TxnDB` to `TransactionDB`.

src/txn/txn_db.rs
yiyuanliu: I don't really understand why thread mode is needed, is it for performance reasons?

If we follow the mechanism in `DBWithThreadMode`, we will have typename such as `TransactionDBWithThreadMode<MultiThreaded>` and `OptimisticTransactionDBWithThreadMode<MultiThreaded>`. That looks ugly.

src/write_batch.rs
yiyuanliu: RocksDB will return `InvalidArguments` if `delete_range` is used on TransactionDB.

`WriteBatch<D = DB>` seems a good solution.

src/txn/txn_db.rs
mappum: Not necessarily, the naming convention could be changed - I think the type should have still been called `DB` but should have been given a default generic parameter to prevent ugliness.

yiyuanliu: > Not necessarily, the naming convention could be changed - I think the type should have still been called `DB` but should have been given a default generic parameter to prevent ugliness.

This has done in commit 5c7e0d9.

src/write_batch.rs
```diff
@@ -77,7 +89,7 @@ unsafe extern "C" fn writebatch_delete_callback(state: *mut c_void, k: *const c_
     leaked_cb.delete(key.to_vec().into_boxed_slice());
 }
 
-impl WriteBatch {
+impl<const TRANSACTION: bool> WriteBatchWithTransaction<TRANSACTION> {
```
ryoqun, line 92: const generics! i wonder it's okay to bump MSRV?

ryoqun:
wow great progress! i'll review this in depth later

src/write_batch.rs
```diff
@@ -16,25 +16,37 @@ use crate::{ffi, AsColumnFamilyRef};
...
 /// {
 ///     let db = DB::open_default(path).unwrap();
-///     let mut batch = WriteBatch::default();
+///     let mut batch = WriteBatchWithTransaction::<false>::default();
 ///     batch.put(b"my key", b"my value");
 ///     batch.put(b"key2", b"value2");
 ///     batch.put(b"key3", b"value3");
+///
+///     // DeleteRange is supported when use without transaction
+///     batch.delete_range(b"key2", b"key3");
```
ryoqun, line 40: cross ref: https://github.com/facebook/rocksdb/pull/8260#pullrequestreview-788541995

src/transactions/transaction_db.rs
```diff
@@ -0,0 +1,778 @@
...
+                self.inner,
+                readopts.inner,
+                key.as_ref().as_ptr() as *const c_char,
+                key.as_ref().len() as size_t,
+                &mut val_len,
+            ));
+            if val_ptr.is_null() {
+                Ok(None)
+            } else {
+                let val = Vec::from_raw_parts(val_ptr as *mut u8, val_len, val_len);
```
yiyuanliu, line 394: I realized that I misunderstood the use of `Vec::from_raw_parts`. It's not safe to build Vec from memory allocated in C.

We may need to copy data or build a new type such as `CVec` to wrap C memory.

vldm: Previously there was `DBVector` implementation, which was removed in:
https://github.com/rust-rocksdb/rust-rocksdb/pull/345/files#diff-7d7d3a6ce5c2a3e8b39ce4f49b78a71365c8e06c0d9ddb78bc2dca68bfeb4a5bL2028

I think it should be reverted, because not all of rocksdb functions use pinnable slice. (especially transaction related)

But if this type would be reverted, make sure that in drop it would use
`rocksdb_free` instead of `libc::free(self.base as *mut c_void);`

vldm (commented):
Hi, thank you for this pull-request, i have tried to use it in my project, and i can left some nitpicks:

I understand that this is C binding legacy:
But maybe we can reduce duplication?
- For example we can implement top-leveltype: `Database<DBMode>, where DBMode=Regular|OptimisticTransaction|Transaction`, and DBMode will be just a collection of c methods, that should be used in one or another place.
- Or at least hide all impls behind macro, to make sure that implementation is only differ in c-method prefix.

If you need any help with this pull-request i would like to help.

src/transactions/transaction.rs
```diff
@@ -0,0 +1,626 @@
...
+            Ok(())
+        }
+    }
+
+    /// Merge value with existing value of key, and also do conflict checking on the key.
+    ///
+    /// See [`merge_cf`] for details.
+    ///
+    /// [`merge_cf`]: Self::merge_cf
+    pub fn merge<K: AsRef<[u8]>, V: AsRef<[u8]>>(&self, key: &K, value: &V) -> Result<(), Error> {
```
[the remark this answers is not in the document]
yiyuanliu: This's a mistake, fixed in 5db5b97.

src/transactions/transaction_db.rs
yiyuanliu: It's better to support `get_pinned` in rocksdb C API, see facebook/rocksdb#9252.

yiyuanliu:
> Hi, thank you for this pull-request, i have tried to use it in my project, and i can left some nitpicks:
> 
> I understand that this is C binding legacy: But maybe we can reduce duplication?
> 
> * For example we can implement top-leveltype: `Database<DBMode>, where DBMode=Regular|OptimisticTransaction|Transaction`, and DBMode will be just a collection of c methods, that should be used in one or another place.
> * Or at least hide all impls behind macro, to make sure that implementation is only differ in c-method prefix.
> 
> If you need any help with this pull-request i would like to help.

I have already add new type `DBCommon<T: ThreadMode, I: DBInner>`. Currently, `DB` and `OptimisticTransactionDB` are implemented through `DBCommon`.

But it seems hard to put TransactionDB into `Database<DBMode>`. The main reason for this is rocksdb C API only provide a limit support for `TransactionDB`. There are many missing functions in `TransactionDB` compared to a normal DB.

vldm:
@yiyuanliu
> There are many missing functions in TransactionDB compared to a normal DB.
Sorry maybe i am wrong but i see that TransactionDB/OptimisticTransactionDB is inherited from regular db. And most of methods are `virtual`.
Why we cant call them from base class? Keeping only transaction related functionality in some kind of impl block?
```rust
impl DBCommon<_, TransactionDB>
{
  fn transaction(&self) -> Transaction;
}
```

yiyuanliu:
> @yiyuanliu
> 
> > There are many missing functions in TransactionDB compared to a normal DB.
> 
> Sorry maybe i am wrong but i see that TransactionDB/OptimisticTransactionDB is inherited from regular db. And most of methods are `virtual`. Why we cant call them from base class? Keeping only transaction related functionality in some kind of impl block?
> 
> ```rust
> impl DBCommon<_, TransactionDB>
> {
>   fn transaction(&self) -> Transaction;
> }
> ```

It seems feasible, although I haven't verified it.

src/transactions/options.rs
```diff
@@ -0,0 +1,283 @@
...
+
+impl Drop for TransactionDBOptions {
+    fn drop(&mut self) {
+        unsafe {
+            ffi::rocksdb_transactiondb_options_destroy(self.inner);
+        }
+    }
+}
+
+pub struct OptimisticTransactionOptions {
```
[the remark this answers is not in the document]
yiyuanliu: Fixed in [9dcf6ef](https://github.com/rust-rocksdb/rust-rocksdb/pull/565/commits/9dcf6efb434360746304350e7dfef03d6fddcf57)

Mo0nbase:
Looks like there were just some log retention issues with the checks is anyone looking to review this atm? Can't wait for this to be merged into main.

aleksuss:
Resolve conflicts please.

librocksdb-sys/rocksdb-c.cc
```diff
@@ -0,0 +1,292 @@
+#include "rocksdb-c.h"
```
aleksuss, line 1: I think it's a bad idea to add this code. I suppose you should create a PR to https://github.com/facebook/rocksdb/ first. What if I'd like to link with system `librocksdb` rather than compile from the sources?

yiyuanliu: > I think it's a bad idea to add this code. I suppose you should create a PR to https://github.com/facebook/rocksdb/ first. What if I'd like to link with system `librocksdb` rather than compile from the sources?

I have created a PR to rocksdb but get no response after a long time (see https://github.com/facebook/rocksdb/pull/9252). So I chose to implement these in rust-rocksdb. And I guess `rust-rocksdb` doesn't support link with system `librocksdb` currently.

aleksuss: On macOS, if I build from master I can link with system `librocksdb` without any problem.
```
 % brew info rocksdb                                                                                                                                                                                                                          
rocksdb: stable 7.0.3 (bottled), HEAD
```

yiyuanliu: > On macOS, if I build from master I can link with system `librocksdb` without any problem.
> 
> ```
>  % brew info rocksdb                                                                                                                                                                                                                          
> rocksdb: stable 7.0.3 (bottled), HEAD
> ```

In this case, maybe we can just compile the C API part of rocksdb and link with system `librocksdb`? But considering that `librocksdb` may use a different C++ compiler, I don't know if this can guarantee compatibility.

gcxfd: > I think it's a bad idea to add this code. I suppose you should create a PR to https://github.com/facebook/rocksdb/ first. What if I'd like to link with system `librocksdb` rather than compile from the sources?

https://github.com/facebook/rocksdb/commit/b71466e9821579592c3324e6b44ad529f2a4f052

gcxfd: > I think it's a bad idea to add this code. I suppose you should create a PR to https://github.com/facebook/rocksdb/ first. What if I'd like to link with system `librocksdb` rather than compile from the sources?

see 
https://github.com/facebook/rocksdb/commit/b71466e9821579592c3324e6b44ad529f2a4f052 , the code seems merged into rocksdb

Cloud33: Any progress? Need to wait for the rocksdb release that contains this update?

gcxfd: RocksDB 7.3.1 [Latest](https://github.com/facebook/rocksdb/releases/latest)
released

yiyuanliu: > RocksDB 7.3.1 [Latest](https://github.com/facebook/rocksdb/releases/latest) released

Unfortunately, this release version doesn't contain the required changes, it added some bug fixes compared to 7.3.0. May have to wait 7.4.

f4184f: Any luck with 7.4.3 ?

src/transactions/transaction.rs
```diff
@@ -0,0 +1,881 @@
...
+        let readopts = ReadOptions::default();
+        self.iterator_opt(mode, readopts)
+    }
+
+    pub fn iterator_opt<'a: 'b, 'b>(
+        &'a self,
+        mode: IteratorMode,
+        readopts: ReadOptions,
+    ) -> DBIteratorWithThreadMode<'b, Self> {
+        DBIteratorWithThreadMode::new(self, readopts, mode)
```
aleksuss, line 745: It's wrong. [Transaction](https://github.com/facebook/rocksdb/blob/main/include/rocksdb/c.h#L2579) and [TransactionDB](https://github.com/facebook/rocksdb/blob/main/include/rocksdb/c.h#L2588) have its own implementations for returning iterators.

aleksuss (changes requested):

aleksuss:
@yiyuanliu do you plan to finish this PR?

yiyuanliu:
> @yiyuanliu do you plan to finish this PR?

I will finish it, but it will take some time. Due to a job change, I've left the project I was working on when I built this PR. I'm sorry I haven't been able to find the time to work on this PR for a while.

src/transactions/transaction.rs
yiyuanliu: Transaction also implements `DBAccess` and call [rocksdb_transaction_create_iterator](https://github.com/yiyuanliu/rust-rocksdb/blob/transaction/src/transactions/transaction.rs#L48) in `create_iterator`.

librocksdb-sys/rocksdb-c.cc
yiyuanliu: updated

aleksuss (approved):

src/write_batch.rs
aleksuss: Yeah. I think it's ok to define MSRV 1.60.0. There is PR #649 which requires MSRV as well.

src/transactions/options.rs
```diff
@@ -0,0 +1,297 @@
...
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+//
+
+use crate::ffi;
+use core::panic;
```
stanislav-tkach, line 17: Why is the `core::panic` import needed?

stanislav-tkach (approved):
Looking good to me, though I would admit, that I can spend more time checking the pull request if I have more time.

src/lib.rs
```diff
@@ -143,6 +172,28 @@ impl Error {
...
+            "Result incomplete" => ErrorKind::Incomplete,
+            "Shutdown in progress" => ErrorKind::ShutdownInProgress,
+            "Operation timed out" => ErrorKind::TimedOut,
+            "Operation aborted" => ErrorKind::Aborted,
+            "Resource busy" => ErrorKind::Busy,
+            "Operation expired" => ErrorKind::Expired,
+            "Operation failed. Try again." => ErrorKind::TryAgain,
+            "Compaction too large" => ErrorKind::CompactionTooLarge,
+            "Column family dropped" => ErrorKind::ColumnFamilyDropped,
+            _ => ErrorKind::Unknown,
```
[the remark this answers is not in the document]
yiyuanliu: For any errors, the user can always get the error message.

Error message is stored in `Error`. Users can parse error messages by calling `error.kind()` . Regardless of the type of error, the user can still get the original error message using `error.to_string()`.

src/transactions/options.rs
aleksuss: I'll remove it later but for now, I merge it as is.

PR #567, opened 2021-11-16
Open DB with read-only access and with column family descriptors

nikurt:
This pull request is motivated by https://github.com/near/nearcore/issues/5194
In short, I want to ensure that a debug-only tool accesses RocksDB strictly read-only.
It has happened already that a newer version of the debug-only tool made changes to the DB. Unfortunately, that meant that the original binary was no longer able to use that DB, and the DB had to be recovered from a backup.

### DIFF
--- a/src/db.rs
+++ b/src/db.rs
@@ -327,6 +327,7 @@ impl<T: ThreadMode> DBWithThreadMode<T> {
             },
         )
     }
+
     /// Opens a database for read only with the given database options and column family names.
     pub fn open_cf_for_read_only<P, I, N>(
         opts: &Options,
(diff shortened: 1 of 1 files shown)


stanislav-tkach (commented):
Perhaps it would be better to expose the `AccessType` value as a parameter?

nikurt:
Thank you for a quick response.

Made an alternative pull request: https://github.com/rust-rocksdb/rust-rocksdb/pull/569
It basically exposes `AccessType` and `open_cf_descriptors_internal` (renamed) in the public interface.

aleksuss:
@nikurt is the PR could be closed ?

PR #569, opened 2021-11-17
Open DB with read-only access and with column family descriptors

nikurt:
Expose `AccessType` in the public interface to avoid creating `open_cf_descriptors` functions for each use case.

This pull request is motivated by near/nearcore#5194
In short, I want to ensure that a debug-only tool accesses RocksDB strictly read-only.
It has happened already that a newer version of the debug-only tool made changes to the DB. Unfortunately, that meant that the original binary was no longer able to use that DB, and the DB had to be recovered from a backup.


stanislav-tkach (approved):

src/db.rs
```diff
@@ -208,7 +208,7 @@ unsafe impl<T: ThreadMode> Send for DBWithThreadMode<T> {}
 unsafe impl<T: ThreadMode> Sync for DBWithThreadMode<T> {}
 
 // Specifies whether open DB for read only.
-enum AccessType<'a> {
+pub enum AccessType<'a> {
```
aleksuss, line 211: Does it really need to be public ?

aleksuss:
Why doesn't `open_cf_for_read_only` method fit for your case  ?

nikurt:
`open_cf_for_read_only` doesn't let me define a merge operator for a column family.
Reads after `open_cf_for_read_only` result in `Error { message: "Invalid argument: merge_operator is not properly initialized." }`

aleksuss:
Ok. What about to change a signature of the `open_cf_for_read_only` for something like this: 
```rust
pub fn open_cf_for_read_only<P, I, N>(
        db_opts: &Options,
        path: P,
        cfs: I,
        cfs_opts: Option<&Options>,
        error_if_log_file_exist: bool,
    ) -> Result<Self, Error>
    where
        P: AsRef<Path>,
        I: IntoIterator<Item = N>,
        N: AsRef<str>,
    {
        let cfs = cfs
            .into_iter()
            .map(|name| ColumnFamilyDescriptor::new(name.as_ref(), cfs_opt.unwrap_or(Options::default())));

        Self::open_cf_descriptors_access_type(
            opts,
            path,
            cfs,
            &AccessType::ReadOnly {
                error_if_log_file_exist,
            },
        )
    }
```

I think it will cover your case and will be useful for future usage. Also, it allows us not to make two entities public. 
cc @stanislav-tkach  ?

nikurt:
I'm in favor of not making two entities public, but that suggestion doesn't provide a way to set different options for different column families. In my use case there are not many different cf options, but this seems to be an unnecessary restriction.

CF options in my use case: https://github.com/near/nearcore/blob/a9c07344b542173d353b5c4db43cf82e07db1a5f/core/store/src/db.rs#L687

aleksuss:
Ok. What about such version: 
```rust
pub fn open_cf_for_read_only<P, I, N>(
        db_opts: &Options,
        path: P,
        cfs: I,
        cfs_opts: Option<&Options>,
        error_if_log_file_exist: bool,
    ) -> Result<Self, Error>
    where
        P: AsRef<Path>,
        I: IntoIterator<Item = (N, Options)>,
        N: AsRef<str>,
    {
        let cfs = cfs
            .into_iter()
            .map(|(name, cf_opts)| ColumnFamilyDescriptor::new(name.as_ref(), cfs_opts));

        Self::open_cf_descriptors_access_type(
            opts,
            path,
            cfs,
            &AccessType::ReadOnly {
                error_if_log_file_exist,
            },
        )
    }
```
???

nikurt:
Can you please clarify, do you suggest to add another function, or to change the existing `open_cf_for_read_only` function?

nikurt:
If the suggestion is to add a new function which accepts `(name, options)`, then it's basically the same as https://github.com/rust-rocksdb/rust-rocksdb/pull/567/files , which passes `ColumnFamilyDescriptors` created from `(name, options)`.

nikurt:
Added a function as you suggested in this PR, please take a look.

aleksuss:
Yeah. It looks better as for me. 
But in the future db public API should be reworked. Because absence of a possibility to create database with custom cf settings looks weird.

aleksuss (approved):

PR #584, opened 2022-01-11
Replace mem::transmute with narrower conversions

niklasf:
Using more specific conversions is preferable, where possible.

### DIFF
--- a/src/db_options.rs
+++ b/src/db_options.rs
@@ -13,7 +13,6 @@
 // limitations under the License.
 
 use std::ffi::{CStr, CString};
-use std::mem;
 use std::path::Path;
 use std::sync::Arc;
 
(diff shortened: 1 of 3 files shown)


src/db_options.rs
```diff
@@ -1202,7 +1201,7 @@ impl Options {
 
         unsafe {
             let cf = ffi::rocksdb_compactionfilter_create(
-                mem::transmute(cb),
+                Box::into_raw(cb) as *mut c_void,
```
aleksuss, line 1204:
```suggestion
                Box::into_raw(cb).cast::<c_void>(),
```

Isn't it safer it such way ?

niklasf: Didn't know this one. It's safer, because it can't accidentally go from `const *` to `mut *`. Will update the PR accordingly.

aleksuss (approved):

stanislav-tkach (approved):

PR #586, opened 2022-01-14
Make `CacheWrapper` and `EnvWrapper` Send and Sync

aleksuss:
Suggestions by clippy

### DIFF
--- a/src/db_options.rs
+++ b/src/db_options.rs
@@ -382,9 +382,7 @@ unsafe impl Send for BlockBasedOptions {}
 unsafe impl Send for CuckooTableOptions {}
 unsafe impl Send for ReadOptions {}
 unsafe impl Send for IngestExternalFileOptions {}
-unsafe impl Send for Cache {}
 unsafe impl Send for CacheWrapper {}
-unsafe impl Send for Env {}
 unsafe impl Send for EnvWrapper {}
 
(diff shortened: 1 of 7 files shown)


src/db_options.rs
```diff
@@ -394,7 +396,9 @@ unsafe impl Sync for CuckooTableOptions {}
 unsafe impl Sync for ReadOptions {}
 unsafe impl Sync for IngestExternalFileOptions {}
 unsafe impl Sync for Cache {}
+unsafe impl Sync for CacheWrapper {}
 unsafe impl Sync for Env {}
```
niklasf, line 400: Looks like `Env` will now automatically be `Send + Sync` (https://doc.rust-lang.org/std/sync/struct.Arc.html#impl-Send). Same for `Cache`.

src/ffi_util.rs
```diff
@@ -28,9 +28,9 @@ pub(crate) unsafe fn raw_data(ptr: *const c_char, size: usize) -> Option<Vec<u8>
     if ptr.is_null() {
         None
     } else {
-        let mut dst = Vec::with_capacity(size);
+        let mut dst = vec![0; size];
```
niklasf, line 31: After #583 this change would no longer be required to make clippy happy.

stanislav-tkach (approved):

PR #588, opened 2022-01-19
Fix theoretical UB while transmuting Arc

niklasf:
As far as I can tell, there is no guarantee that `Arc<T>` can be soundly transmuted to `Arc<U>`, even if `T` can be soundly transmuted to `U`.

Looking at the implementation of `Arc`, it is effectively a single field struct, wrapping a private `NonNull<ArcInner<T>>` where `ArcInner<T>` is `#[repr(C)]`, so the transmute is not UB as of now, and of course it's unlikely that this would change. But it's still not good to rely on such an implementation detail for soundness.


stanislav-tkach (approved):
Makes sense, thank you!

aleksuss (approved):

PR #589, opened 2022-01-20
Add constants for all supported properties

steviez:
I want to query several properties and noticed there aren't equivalent constants as there are in the RocksDB repo.

This was a copy/paste with some regex'ing to make it adhere to Rust syntax. I figured including the comments inline is nice, but I could also see argument of not including them and instead referring reader to the block starting at:
https://github.com/facebook/rocksdb/blob/08809f5e6cd9cc4bc3958dd4d59457ae78c76660/include/rocksdb/db.h#L428

### DIFF
--- a/src/properties.rs
+++ b/src/properties.rs
@@ -1,241 +1,217 @@
-// Copyright ???
-//
-// Licensed under the Apache License, Version 2.0 (the "License");
-// you may not use this file except in compliance with the License.
-// You may obtain a copy of the License at
-//
-// http://www.apache.org/licenses/LICENSE-2.0
-//
-// Unless required by applicable law or agreed to in writing, software
-// distributed under the License is distributed on an "AS IS" BASIS,
-// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
-// See the License for the specific language governing permissions and
-// limitations under the License.
+//! Properties
+//!
+//! Full list of valid properties and descriptions pulled from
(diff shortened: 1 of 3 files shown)


src/properties.rs
```diff
@@ -0,0 +1,241 @@
...
+
+/// Full list of valid properties and descriptions pulled from
+/// [here](https://github.com/facebook/rocksdb/blob/08809f5e6cd9cc4bc3958dd4d59457ae78c76660/include/rocksdb/db.h#L428-L634).
+
+// concat!() takes string literals only so use a macro instead of another constant
+macro_rules! ROCKSDB_PREFIX {
+    () => {
+        "rocksdb."
+    };
+}
```
stanislav-tkach, line 23: I suppose it would be better to define a slightly different macro:
```rust
macro_rules! property {
    ($suffics: literal) => {
        concat!("rocksdb.", $suffics)
    };
}
```
This way declarations would be slightly less verbose:
```rust
pub const NUM_FILES_AT_LEVEL_PREFIX: &str = property!("num-files-at-level");
```

stanislav-tkach (commented):
Thanks for the pull request! I think having comments are fine, though I would prefer doc-comments instead of the regular ones.

src/properties.rs
steviez: Ooh nice, much cleaner than what I had

stanislav-tkach (approved):

aleksuss (approved):

PR #590, opened 2022-01-21
Support configuring bottom-most compression level

mina86:
Weirdly, RocksDB C interface does not have a way to just set the
`enabled` flag of the bottommost compression options so I’ve decided
to replicate that interface in Rust as well.  As far as I understand,
this unfortunately means that to set bottommost compression one needs
to set the type and then also use one of the other methods for
configuring the compression.

Issue: https://github.com/rust-rocksdb/rust-rocksdb/issues/260
Issue: https://github.com/near/nearcore/issues/6119


stanislav-tkach (approved):

mina86:
I guess all green means it can be merged? ;)

PR #593, opened 2022-01-25
Update RocksDB to v6.29.0

niklasf:
For bugfixes and new features. The following required further changes in this library:

* From 6.29.0:

  > Take compression level_values as const pointer 

* From 6.24.0:

  > Generalized bits_per_key parameters in C API from int to double for greater configurability. Although this is a compatible change for existing C source code, anything depending on C API signatures, such as foreign function interfaces, will need to be updated.

### DIFF
--- a/librocksdb-sys/rocksdb_lib_sources.txt
+++ b/librocksdb-sys/rocksdb_lib_sources.txt
@@ -1,5 +1,7 @@
 cache/cache.cc
 cache/cache_entry_roles.cc
+cache/cache_key.cc
+cache/cache_reservation_manager.cc
 cache/clock_cache.cc
 cache/lru_cache.cc
 cache/sharded_cache.cc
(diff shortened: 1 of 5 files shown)


librocksdb-sys/rocksdb_lib_sources.txt
```diff
@@ -136,6 +142,12 @@ options/options.cc
 options/options_helper.cc
 options/options_parser.cc
 port/port_posix.cc
+port/win/env_default.cc
+port/win/env_win.cc
+port/win/io_win.cc
+port/win/port_win.cc
+port/win/win_logger.cc
+port/win/win_thread.cc
```
aleksuss, line 150: These files isn't needed for unix like OS. They're added manually in build.rs.

niklasf: Ah! Removed.

stanislav-tkach (approved):

aleksuss:
Overall looks good. But my concern is that there is no release 6.29.0 yet. The last one is [6.28.2](https://github.com/facebook/rocksdb/releases).

niklasf:
Oh, I just assumed they stopped tagging on GitHub when they put [`2021-12-17` as the release date](https://github.com/facebook/rocksdb/blob/main/HISTORY.md#6280-2021-12-17) and no `v6.28.x` tag was in sight. But today the corresponding release/tag got published. I guess the same will happen with `v6.29.x`, so it might be better to wait, then.

Could also PR `v6.28.2` in the meantime, although I personally want one of the fixes from `v6.29.0`.

akrylysov:
I made a similar PR and saw this one a minute later. https://github.com/rust-rocksdb/rust-rocksdb/pull/600 upgrades to the latest release - 6.28.2.

niklasf:
Closing in favor of the immediately actionable PR for now.

PR #601, opened 2022-02-03
Release 0.18.0

aleksuss:


akrylysov:
There is https://github.com/rust-rocksdb/rust-rocksdb/issues/602 that should be fixed before the release.

stanislav-tkach (approved):
I'm still not entirely convinced about downgrading the `librocksdb-sys` version, but we have already discussed that. Otherwise looking good. :upside_down_face:

PR #616, opened 2022-03-03
Small clean-ups (including fix for UB)

niklasf:
Most of these are trivial. I'll highlight some important bits with inline comments.


src/slice_transform.rs
```diff
@@ -108,9 +108,8 @@ pub unsafe extern "C" fn in_domain_callback(
     raw_cb: *mut c_void,
     raw_key: *const c_char,
     key_len: size_t,
-) -> u8 {
+) -> c_uchar {
     let cb = &mut *(raw_cb as *mut TransformCallback);
     let key = slice::from_raw_parts(raw_key as *const u8, key_len as usize);
-    cb.in_domain_fn
-        .map_or(0xff, |in_domain| in_domain(key) as u8)
```
niklasf, line 115: Returning `0xff` here was somewhat questionable (as a `bool` is expected in C++ land), but not a problem as far as I can tell after reading the current implementation in RocksDB.

src/db.rs
```diff
@@ -637,25 +638,27 @@ impl<T: ThreadMode> DBWithThreadMode<T> {
                     cfhandles.as_mut_ptr(),
                 )),
                 AccessType::Secondary { secondary_path } => {
+                    let secondary_cpath = to_cpath(secondary_path)?;
                     ffi_try!(ffi::rocksdb_open_as_secondary_column_families(
                         opts.inner,
                         cpath.as_ptr() as *const _,
-                        to_cpath(secondary_path)?.as_ptr() as *const _,
+                        secondary_cpath.as_ptr() as *const _,
```
niklasf, line 645: I believe this was undefined behavior, because the temporary owned `CString` would have been dropped before the pointer was used in the called function.

src/db.rs
```diff
@@ -637,25 +638,27 @@ impl<T: ThreadMode> DBWithThreadMode<T> {
...
+                    let ttls_v = vec![ttl.as_secs() as c_int; cfs_v.len()];
                     ffi_try!(ffi::rocksdb_open_column_families_with_ttl(
                         opts.inner,
                         cpath.as_ptr(),
                         cfs_v.len() as c_int,
                         cfnames.as_ptr(),
                         cfopts.as_ptr(),
                         cfhandles.as_mut_ptr(),
-                        &(ttl.as_secs() as c_int) as *const _,
+                        ttls_v.as_ptr(),
```
niklasf, line 661: I believe this was undefined behavior because the temporary `c_int` would be out of scope before the pointer is used in the called function.

Aditionally there would be an out-of-bounds read if `cfs_v.len() > 1`, because the implementation expects a pointer to `cfs_v.len()` consecutive entries.

niklasf:
Oh, looks like the pointers to temporaries are fine, actually:

> The [drop scope](https://doc.rust-lang.org/stable/reference/destructors.html#drop-scopes) of the temporary is usually the end of the enclosing statement.

https://doc.rust-lang.org/stable/reference/expressions.html?highlight=Tempo#temporaries

That leaves only the out of bounds read as UB, with all other changes being cosmetic.

src/db.rs
stanislav-tkach: Perhaps I'm missing something, but where can out-of-bound read occur? `cfnames`? But it is an array, so everything should be OK. Am I missing something?

stanislav-tkach (approved):

src/db.rs
niklasf: I mean the `ttls` array. Previously we were passing a pointer to a single integer value, but `n` integer values are expected (https://github.com/facebook/rocksdb/blob/f20b674796ffd7ca32471705876fc651b8e246db/db/c.cc#L827-L828), one for each column family.

niklasf: I mean the `ttls` array. Previously we were passing a pointer to a single integer value, but `n` integer values are expected (https://github.com/facebook/rocksdb/blob/f20b674796ffd7ca32471705876fc651b8e246db/db/c.cc#L827-L828), one for each column family.

niklasf:
Rebased since #627 got merged first.

niklasf:
Prepared rustsec advisory https://github.com/rustsec/advisory-db/pull/1237.

aleksuss (approved):

PR #620, opened 2022-03-14
Use more target features to build librocksdb-sys

niklasf:
These offer slight speedups mostly in the Ribbon filter implementation.

* https://github.com/facebook/rocksdb/blob/a6a179859e612856f9de8b7667dd0a2cfddd286a/build_tools/build_detect_platform#L711-L718
* https://github.com/facebook/rocksdb/blob/a6a179859e612856f9de8b7667dd0a2cfddd286a/build_tools/build_detect_platform#L805


dignifiedquire:
nice, would be great to get these in. if you rebase on master ci should be happy

niklasf:
Will do, merge conflict resolved.

PR #621, opened 2022-03-17
librocksdb-sys: Add support for building with io_uring on Linux.

parazyd:


librocksdb-sys/build.rs
```diff
@@ -198,6 +198,11 @@ fn build_rocksdb() {
         config.define("WITH_JEMALLOC", "ON");
     }
 
+    if target.contains("linux") && cfg!(feature = "io-uring") {
+        pkg_config::probe_library("liburing").unwrap();
```
aleksuss, line 202: What will be if there is no `liburing` library on a host?

```suggestion
        if pkg_config::probe_library("liburing").is_ok() {
            config.define("ROCKSDB_IOURING_PRESENT", Some("1"));
        }
```

aleksuss, line 202: What will be if there is no `liburing` library on a host?

```suggestion
        if pkg_config::probe_library("liburing").is_ok() {
            config.define("ROCKSDB_IOURING_PRESENT", Some("1"));
        }
```

aleksuss:
It would be good to add a new job in a CI for checking this feature.

parazyd:
@aleksuss I thought it's better not to silently fall through if the feature is explicitly enabled but the library is unavailable. Would you like me to chane this?

aleksuss:
Then you probably should add a message to inform a user: 
```rust
pkg_config::probe_library("liburing").except("The io_uring feature was provided but the library was not found");
```
or something...

parazyd:
Updated. Thanks.

librocksdb-sys/build.rs
```diff
@@ -218,6 +218,12 @@ fn build_rocksdb() {
         config.define("WITH_JEMALLOC", "ON");
     }
 
+    if target.contains("linux") && cfg!(feature = "io-uring") {
```
aleksuss, line 221:
```suggestion
    #[cfg(feature = "io-uring")]
    if target.contains("linux") && cfg!(feature = "io-uring") {
```

librocksdb-sys/build.rs
```diff
@@ -218,6 +218,12 @@ fn build_rocksdb() {
...
+        pkg_config::probe_library("liburing")
+            .except("The io-uring feature was requested but the library is not available");
```
aleksuss, line 223:
```suggestion
            .expect("The io-uring feature was requested but the library is not available");
```

aleksuss (changes requested):
Fix the typo in method `expect` please.

stanislav-tkach (approved):

aleksuss (approved):

PR #622, opened 2022-03-24
Fix memory leak when reading properties and avoid memory allocation

mina86:
First of all, when property_value* method fails to parse returned
value as a string (possibly because the value has non UTF-8 value) it
returns an error.  When that happens, `libc::free` line is never
reached and thus the value is never freed.  Fix that by eliminating
the early return.

Second of all, while at it, refactor the property_value* methods by
introducing property_value_impl generic method which is parameterised
with a function that parses `&str` property to appropriate type.  This
means that now when calling property_value_int* the `&str` value is
not first converted to a `String` to then be parsed as `u64`; instead,
`&str` is parsed directly as `u64` which saves the `String` memory
allocation.

As an aside, based on use cases there may still be another unnecessary
allocation, e.g. `println!("{}", db.property_value("property"))`
results in unnecessary `String` allocation.  This could be addressed
e.g. by having a `pub struct PropertyValue(*mut c_char, usize)` type
which wraps offers necessary conversion methods.  Such change is of
course beyound the scope of this commit.


stanislav-tkach (approved):

mina86:
Looks like Clippy is complaining about unrelated code.

mina86:
Friendly ping.

aleksuss:
We should merge #565 first

aleksuss (approved):

PR #626, opened 2022-04-21
Upgrade to rocksdb v7

dignifiedquire:
This updates all sources files & the sys package so it compiles. Does not add any new features


stanislav-tkach (approved):

librocksdb-sys/build.rs
```diff
@@ -184,12 +184,20 @@ fn build_rocksdb() {
...
+        let mut win_sources = vec![
+            "port/win/env_default.cc",
+            "port/win/port_win.cc",
+            "port/win/xpress_win.cc",
+            "port/win/io_win.cc",
+            "port/win/win_thread.cc",
+            "port/win/env_win.cc",
+            "port/win/win_logger.cc",
+        ];
+        lib_sources.append(&mut win_sources);
```
aleksuss, line 196: I think the creation of the additional vector is redundant here. You could write something like this:
```suggestion
        lib_sources.extend([
            "port/win/env_default.cc",
            "port/win/port_win.cc",
            "port/win/xpress_win.cc",
            "port/win/io_win.cc",
            "port/win/win_thread.cc",
            "port/win/env_win.cc",
            "port/win/win_logger.cc",
        ]);
```

aleksuss (commented):
Overall looks good.

Issue #609, opened 2022-02-15
librocksdb_sys::bz_internal_error conflicts with bzip2-sys::bz_internal_error


nikurt:
I'm getting the following link error due to the duplicate symbol when linking a binary with link-time optimizations:
```
   Compiling neard v0.0.0 (/home/nikurt/nearcore-rdb-3/neard)
warning: Linking globals named 'bz_internal_error': symbol multiply defined!

error: failed to load bc of "librocksdb_sys-33c86cdce183a558.librocksdb_sys.8e9e6b3e-cgu.0.rcgu.o":

warning: `neard` (bin "neard") generated 1 warning
error: could not compile `neard` due to previous error; 1 warning emitted
make: *** [Makefile:34: neard-release] Error 101
```

Symbols are defined here:
* `librocksdb-sys`: https://github.com/rust-rocksdb/rust-rocksdb/blob/180e65e1b00f6b032e1513d94a4f0a4de7cafba1/librocksdb-sys/src/lib.rs#L32
* `bzip2-sys`: https://github.com/alexcrichton/bzip2-rs/blob/016e18155ef7c05983ea244cae1344c5b68defd8/bzip2-sys/lib.rs#L73

A workaround is to disable feature `bzip2` of crate `rocksdb`.

Link-time optimizations defined in my [`Makefile`](https://github.com/near/nearcore/blob/5568df666a0f5025c724027814069050acb194c6/Makefile#L2) as follows:
```
CARGO_PROFILE_RELEASE_LTO = fat
```

dnrusakov:
Is it possible to release a new version of the `rocksdb` crate in order to accommodate this change? The current version available on crates.io is 0.18.0 and doesn't have all recent changes including this one. cc @aleksuss

PR #555, opened 2021-09-14
Use *-sys crates where possible

Dr-Emann:
Depend on `bzip2-sys`, `libz-sys`, and `zstd-sys`, rather than bundling them ourselves. This should reduce the burden of keeping those libraries up to date: someone else has already figured out how to do the build system, etc. Additionally, this avoids issues with multiple versions of the same functions when `rocksdb-sys` links against say `zstd`, and another crate pulls in `zstd-sys`, which also links against zstd.

This does not use `snappy-sys` or `lz4-sys` because neither currently exposes the include directory (via "cargo:include=..", which we can use as "DEP_XXX_INCLUDE") to dependencies when building a vendored version of the library, which we need.

Additionally, add tests for creating a DB with each compression type (checking that each compression type works/doesn't work as expected based on the features. Also put the zstd specific test behind a cfg, so tests pass with `--no-default-features`

Fixes #514 
Fixes #535
Fixes Nemo157/async-compression#127

PR #628, opened 2022-04-28
fix: 'bz_internal_error': symbol multiply defined!

nanpuyue:
Fix: https://github.com/rust-rocksdb/rust-rocksdb/issues/609

The `bz_internal_error` was introduced when  [bzip2 support was added](https://github.com/rust-rocksdb/rust-rocksdb/pull/209), the bzip2-sys crate was not used at that time.
This function should be removed after [using bzip2-sys crate](https://github.com/rust-rocksdb/rust-rocksdb/pull/555), otherwise it will conflict with the definition in bzip2-sys when compiled with lto enabled.

### DIFF
--- a/librocksdb-sys/src/lib.rs
+++ b/librocksdb-sys/src/lib.rs
@@ -26,9 +26,3 @@ extern crate libz_sys;
 extern crate zstd_sys;
 
 include!(concat!(env!("OUT_DIR"), "/bindings.rs"));
-
-#[cfg(feature = "bzip2")]
-#[no_mangle]
-pub fn bz_internal_error(errcode: libc::c_int) {
-    panic!("bz internal error: {}", errcode);
-}


stanislav-tkach (approved):

nanpuyue:
A test failed, but it's ok before rebase.
I have no idea,  have a retry? @stanislav-tkach

stanislav-tkach:
The test has passed after restart...

dnrusakov:
Is it possible to release a new version of the rocksdb crate in order to accommodate this change? The current version available on crates.io is 0.18.0 and it doesn't have all recent changes including this one. cc @aleksuss @stanislav-tkach

joleeee:
It's now 4 months since the original issue, and 2 months since it was fixed, but no new crate release. Like @dnrusakov said, it would be great if a new release could be made

joleeee:
A temporary workaround apart from using the git version is to disable default-features
```toml
rocksdb = { version = "0.18.0", default-features = false, features = [ ... ] }
```

Issue #558, opened 2021-10-09
iOS support


pascalbros:
For a personal project I just added a "raw" support for iOS as stated [here](https://github.com/pascalbros/rust-rocksdb/blob/3152d248708ee91239b6e1fe2e6dc9e69fdf4735/librocksdb-sys/build.rs#L120).
I think it can help, I just added some copy-paste code from darwin and it works like a charm on aarch64-apple-ios.

Cheers!

stanislav-tkach:
Sorry for the late reply, but the link gives me the 404 error.

pascalbros:
[Here](https://github.com/iotaledger/rust-rocksdb/blob/3152d248708ee91239b6e1fe2e6dc9e69fdf4735/librocksdb-sys/build.rs#L120)

PR #637, opened 2022-05-23
Add (experimental) iOS support

rajivshah3:
This PR fixes compilation errors when targeting `aarch64-apple-ios` (based on work by @pascalbros in #558). The minimum iOS version is 11.0 because of support for thread-local storage as well as the following error:

```
warning: rocksdb/db/repair.cc:198:17: error: aligned allocation function of type 'void *(std::size_t, std::align_val_t)' is only available on iOS 11 or newer
warning:       db_impl = new DBImpl(db_options_, dbname_);
warning:                 ^
warning: rocksdb/db/repair.cc:198:17: note: if you supply your own aligned allocation functions, use -faligned-allocation to silence this diagnostic
```

IMO this is still "experimental" because I'm not sure if the full RocksDB is intended to be built for iOS. The docs say that [RocksDBLite](https://github.com/facebook/rocksdb/blob/main/ROCKSDB_LITE.md), which includes less features, should be used on mobile. When building RocksDB for iOS, flags for RocksDBLite are enabled [by default](https://github.com/facebook/rocksdb/blob/56ce3aef3377e90cf1aae24189c6e218f01f17db/build_tools/build_detect_platform#L157).


rajivshah3:
I'm not sure why Windows is failing 🤔

stanislav-tkach:
> I'm not sure why Windows is failing 🤔

It looks disturbing, but shouldn't be related to your changes.

aleksuss:
Closed in favour #653

PR #641, opened 2022-06-06
Allow passing &CStr arguments

mina86:
Methods such as property_value take `&str` name as argument.  However,
since they interact with C functions, such string arguments need to be
converted into a `&CStr`.  Because the argument is `&str` this must be
done via `CString::new` which needs to allocate memory and verify that
the value does not have embedded NUL bytes.  Even if the caller
already has a `&CStr` there is no way to pass it directly to the
function and a needless allocation happens.

This commit introduces a `CStrLike` trait which can be used in place
of `&str` arguments.  A function which accepts `impl CStrLike` as an
argument can be called either with `&str` or `&CStr` which allows it
to take the C string directly if it’s passed.

Note that this is strictly speaking backwards-incompatible change.
While call sites such as `db.property_value("foo")` won’t be affected,
call sites which use long deref chains or need type inference will
stop compiling.

Alternatives to this are to:
1. do nothing,
2. introduce `_cstr` variants of functions which need `&CStr`
   internally or
3. change argument type to `&CStr` where functions do the conversion.

The first option forces users to incur the allocation and verification
cost each time method is called.  The second bloats the API by
introducing near-duplicates of existing methods and may lead to
further exponential explosion in the future.  The third one removes
convenience of using string literals from users who might not care
about performance as much.

I think the third option is actually the cleanest but wished to
maintain some backwards compatibility so went with this commit.

### DIFF
--- a/src/db.rs
+++ b/src/db.rs
@@ -1083,8 +1083,10 @@ impl<T: ThreadMode> DBWithThreadMode<T> {
         name: impl CStrLike,
         opts: &Options,
     ) -> Result<*mut ffi::rocksdb_column_family_handle_t, Error> {
-        let cf_name = name.bake().map_err(|_| {
-            Error::new("Failed to convert path to CString when creating cf".to_owned())
+        let cf_name = name.bake().map_err(|err| {
+            Error::new(format!(
+                "Failed to convert path to CString when creating cf: {err}"
+            ))
         })?;
         Ok(unsafe {
             ffi_try!(ffi::rocksdb_create_column_family(
--- a/src/db.rs
+++ b/src/db.rs
@@ -19,7 +19,7 @@ use crate::{
     column_family::UnboundColumnFamily,
     db_options::OptionsMustOutliveDB,
     ffi,
-    ffi_util::{from_cstr, opt_bytes_to_ptr, raw_data, to_cpath},
+    ffi_util::{from_cstr, opt_bytes_to_ptr, raw_data, to_cpath, CStrLike},
     ColumnFamily, ColumnFamilyDescriptor, CompactOptions, DBIteratorWithThreadMode,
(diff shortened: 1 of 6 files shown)


src/db.rs
```diff
@@ -1080,16 +1080,12 @@ impl<T: ThreadMode> DBWithThreadMode<T> {
...
         opts: &Options,
     ) -> Result<*mut ffi::rocksdb_column_family_handle_t, Error> {
-        let cf_name = if let Ok(c) = CString::new(name.as_bytes()) {
-            c
-        } else {
-            return Err(Error::new(
-                "Failed to convert path to CString when creating cf".to_owned(),
-            ));
-        };
+        let cf_name = name.bake().map_err(|_| {
```
stanislav-tkach, line 1086: Lets preserve the internal error like in other places.

stanislav-tkach (approved):
Thanks for the pull request! I like the idea.

aleksuss:
Does it make sense to add `From<&str> for CString` and `From<&String> for CString` trait implementations?

mina86:
> Does it make sense to add `From<&str> for CString` and `From<&String> for CString` trait implementations?

From, &str and CString are all built-in types so we wouldn’t be able to add that implementation.  Furthermore, conversion from `str` to `CString` can fail so that is not an option.  There could be a TryFrom<&str> for CString which essentially is the CString::new constructor.

Regardless, I’m actually not sure what you’re suggesting.  Part of the issue is that we don’t always want to convert argument into a new owned CString.  If user passes us &CStr we just want to do that directly.  This is why CStrLike has the Baked associated type which differs depending on whether we need to perform conversion (in which case we bake to CString) or we don’t (in which case bake is mostly a no-op returning &CStr).

PR #643, opened 2022-06-15
Convert properties to &CStr

mina86:
Now that property_value et al accept &CStr arguments (via the use of
CStrLike trait), change properties to use &CStr type.  This way, if
someone calls `db.property_value(properties::STATS)` (which is the
most obvious and ergonomic way to use the property definitions) they
don’t have to pay the price of allocating temporary CString.

Furthermore, replace ‘at level’ property name prefixes with functions
which take level as an argument and return a property name.  This is
a simpler interface than providing a &CStr prefix and expecting user
to append the level index since user would have to deal with &CStr not
being Display.

Note that this is a breaking change.  While uses which don’t care
about the type of properties and simply passed them as arguments to
DB::property_value method will work, code that assumed &str type will
break as well as uses of the prefix constants which are now gone.

### DIFF
--- a/src/properties.rs
+++ b/src/properties.rs
@@ -3,36 +3,46 @@
 //! Full list of valid properties and descriptions pulled from
 //! [here](https:///github.com/facebook/rocksdb/blob/08809f5e6cd9cc4bc3958dd4d59457ae78c76660/include/rocksdb/db.h#L428-L634).
 
+use std::ffi::{CStr, CString};
+
 macro_rules! property {
     ($suffix: literal) => {
-        concat!("rocksdb.", $suffix)
+        // SAFETY: We’re appending terminating NUL byte and this macro is always
+        // called with values without interior NUL bytes.
+        unsafe {
+            CStr::from_bytes_with_nul_unchecked(concat!("rocksdb.", $suffix, "\0").as_bytes())
+        }
     };
 }
 
 /// "rocksdb.num-files-at-level<N>" - returns string containing the number
 /// of files at level <N>, where <N> is an ASCII representation of a
 /// level number (e.g., "0").
-pub const NUM_FILES_AT_LEVEL_PREFIX: &str = property!("num-files-at-level");
+pub fn num_files_at_level(level: usize) -> CString {
+    unsafe { level_property("num-files-at-level", level) }
+}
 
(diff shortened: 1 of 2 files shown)


src/properties.rs
```diff
@@ -42,176 +52,200 @@ pub const CFSTATS: &str = property!("CFSTATS");
...
+///
+/// `name` is the infix of the property name (e.g. `"num-files-at-level"`) and
+/// `level` is level to get statistics of.  The property name is constructed as
+/// `"rocksdb.<name><level>"`.
+///
+/// Expects `name` not to contain any interior NUL bytes.
+unsafe fn level_property(name: &str, level: usize) -> CString {
+    let bytes = format!("rocksdb.{}{}\0", name, level).into_bytes();
+    // SAFETY: We’re appending terminating NUL and we’re expecting user not to
+    // pass us a string with interior NUL bytes.
```
stanislav-tkach, line 240: I didn't get a comment about a user. This function is private and is only called from the library code, so this is indeed safe.

I think it would be better to place this (or a modified) comment into the function description as it is usually done in the standard library.

stanislav-tkach (approved):
Thanks!

src/properties.rs
mina86: User is anyone who calls the function.  In this case all the users are internal.  I’ve updated the comment to mention that.

The docstring of the function actually has that requirement as well.  It’s the last line.

The confusing thing here is that the `CString::from_vec_with_nul_unchecked` call should be wrapped inside of an unsafe block (and then this `SAFETY` comment would relate to that) but that currently results in ‘unnecessary `unsafe` block’ warning since [unsafe_block_in_unsafe_fn](https://rust-lang.github.io/rfcs/2585-unsafe-block-in-unsafe-fn.html) isn’t a thing yet.

stanislav-tkach: > The docstring of the function actually has that requirement as well. It’s the last line.

Yes, but in the standard library it is described in a separate section. For example here: https://doc.rust-lang.org/src/core/str/mod.rs.html#331
I think that safety invariants are more clear this way.

Anyway, I have approved your pull request, thanks for making this library better!

aleksuss (approved):

PR #645, opened 2022-06-21
Upgrade to rocksdb v7.3.1

yhchiang-sol:
This PR upgrades to rocksdb v7.3.1.
v7.3.1 release note (https://github.com/facebook/rocksdb/releases/tag/v7.3.1).


yhchiang-sol:
Manually replacing the `build_version.cc` under `librocksdb-sys` solved the above error.

src/db_iterator.rs
```diff
@@ -529,6 +529,9 @@ impl Iterator for DBWALIterator {
     type Item = (u64, WriteBatch);
 
     fn next(&mut self) -> Option<(u64, WriteBatch)> {
+        if !self.valid() {
```
yhchiang-sol, line 532: Adding this check as the test will crash without this check.

Also, according to [include/rocksdb/iterator.h](https://github.com/yhchiang-sol/rocksdb/blob/main/include/rocksdb/iterator.h#L65-L68), valid is required before calling next.

librocksdb-sys/build_version.cc
```diff
@@ -3,13 +3,33 @@
 #include <memory>
 
 #include "rocksdb/version.h"
+#include "rocksdb/utilities/object_registry.h"
```
yhchiang-sol, line 6: The build_verion.cc is generated by building rocksdb.  The file is generated from build_version.cc.in.

Since https://github.com/facebook/rocksdb/pull/7949 updates build_version.cc.in, the resulting build_verion.cc will also reflect the change.

yhchiang-sol: Without updating our build_version.cc, it will result in the following link error, which symbol is introduced in https://github.com/facebook/rocksdb/pull/7949:

```
  = note: Undefined symbols for architecture arm64:
            "rocksdb::ObjectRegistry::builtins_", referenced from:
                rocksdb::ObjectRegistry::ObjectRegistry(std::__1::shared_ptr<rocksdb::ObjectLibrary> const&) in liblibrocksdb_sys-d6e9cb3766a4628b.rlib(object_registry.o)
          ld: symbol(s) not found for architecture arm64
          clang: error: linker command failed with exit code 1 (use -v to see invocation)
```

src/db_iterator.rs
yhchiang-sol, line 532: Because of the GetUpdatesSince() fix from facebook/rocksdb#9459,
the resulting iterator from `DB::get_updates_since()` will have its status Invalid
on empty result.

tests/test_db.rs
```diff
@@ -408,8 +408,8 @@ fn test_get_updates_since_empty() {
     let path = DBPath::new("_rust_rocksdb_test_get_updates_since_empty");
     let db = DB::open_default(&path).unwrap();
     // get_updates_since() on an empty database
-    let mut iter = db.get_updates_since(0).unwrap();
-    assert!(iter.next().is_none());
+    let iter = db.get_updates_since(0).unwrap();
+    assert!(!iter.valid());
```
yhchiang-sol, line 412: Because of the GetUpdatesSince() fix from https://github.com/facebook/rocksdb/pull/9459,
the resulting iterator from DB::get_updates_since() will have its status Invalid on empty result.

Based on the [rocksdb code comment](https://github.com/facebook/rocksdb/blob/main/include/rocksdb/iterator.h#L65-L68), the iterator status is required to be `valid` when calling next().  As a result, the previous `assert!(iter.next().is_none())` will now crash instead.

The PR currently fixes the test, but please feel free to let me know if you prefer to check the iterator is valid inside the `next()` call instead.

stanislav-tkach: Yes, I think it makes sense to check an iterator for validity inside the `next` call. It is a safe api after all, so we don't want crashes here.

yhchiang-sol: Sounds good!  I've updated the PR!

src/db_iterator.rs
```diff
@@ -529,6 +529,10 @@ impl Iterator for DBWALIterator {
...
```
yhchiang-sol, line 532: `self.valid()` check is added before invoking the ffi.

yhchiang-sol:
@stanislav-tkach Updated the PR to include a validity check inside the next call.  Looks like the test workflow needs maintainers' approval once again after the PR is updated.  Can you help approve the flow?  Thank you!

yhchiang-sol:
Hello @aleksuss, hope everything is going well on your side.  This PR upgrades rocksdb to 7.3.1 and need your review.  Can you help review the PR when you're more available?  Thank you!

yhchiang-sol:
Hello @aleksuss, our Solana needs this PR and https://github.com/rust-rocksdb/rust-rocksdb/pull/633 in order to use the optimized version of multi-get to speed up our read requests.  Both PRs have been waiting for your review for around a week.  Do you mind reviewing this PR that upgrades rocksdb to v7.3.1 and https://github.com/rust-rocksdb/rust-rocksdb/pull/633 that adds the optimized version of multi-get api?  Thank you so much!

yhchiang-sol:
Hello @stanislav-tkach.  While the PR has one approval, it looks like I am not able to merge it.

If the PR looks good to you, do you mind helping me merge the PR?  Thank you!

aleksuss (approved):

yhchiang-sol:
Thank you so much, @aleksuss!

PR #647, opened 2022-06-28
Add `DBRawIteratorWithThreadMode::item` method returning (key, value)

mina86:
The method is morally equivalent to calling `key` and `value` methods
separately however it reduces number of times validity of the iterator
needs to be checked.  Since that check goes through ffi layer and
calls a virtual method it doesn’t optimise well.  Therefore, reducing
number of times the `valid` method is called optimises the code a wee
bit.


stanislav-tkach (approved):
👍

aleksuss:
Overall looks good. Resolve conflict please.

aleksuss (approved):

PR #648, opened 2022-06-28
Change iterators to return Result

mina86:
Currently, Iterator implementation for DBIteratorWithThreadMode
produce (key, value) pairs on iteration.  When it stops (i.e. next
method returns None) the user is expected to call status method to see
if the iteration finished normally or due to an error.  This is, dare
I say, not very idiomatic.

Consider for example std::io::BufRead::lines or std::fs::read_dir
methods.  They return iterators whose Items are std::io::Result<_>
objects.  On error the error is returned as an item from the iterator
rather than finishing the iterator and providing separate way for the
user to check whether there was an error.

Change Iterator implementations for DBIteratorWithThreadMode such that
it mimic this behaviour.  Rather than returning a value and expecting
user to test status once iterator finishes, the iterators now return
Result<_, Error>.  With that also remove valid and status methods
(since the information is already returned via iteration and it makes
the code slightly simpler).

Similarly, for the same reasons, change Item of DBWALIterator to be
Result<_, Error> as well.

This is obviously an API breaking change.  Anyone using iterators will
have to change their code to handle the errors.

### DIFF
--- a/tests/test_iterator.rs
+++ b/tests/test_iterator.rs
@@ -17,130 +17,61 @@ mod util;
 use pretty_assertions::assert_eq;
 
 use rocksdb::{Direction, IteratorMode, MemtableFactory, Options, DB};
-use util::DBPath;
-
-fn cba(input: &[u8]) -> Box<[u8]> {
-    input.to_vec().into_boxed_slice()
-}
+use util::{assert_iter, pair, DBPath};
 
 #[test]
 #[allow(clippy::cognitive_complexity)]
 fn test_iterator() {
     let n = DBPath::new("_rust_rocksdb_iterator_test");
     {
-        let k1: Box<[u8]> = b"k1".to_vec().into_boxed_slice();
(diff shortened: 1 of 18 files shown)


mina86:
SG.  I’ll try to finish this up by the end of the week.

tests/test_iterator.rs
```diff
@@ -73,29 +73,39 @@ fn test_iterator() {
...
-            iterator1.next();
-            assert!(iterator1.valid());
-            iterator1.next();
-            assert!(!iterator1.valid());
+            iterator1.next().unwrap().unwrap();
+            assert_eq!(None, iterator1.next());
+            assert_eq!(None, iterator1.next());
+        }
+        {
+            // Chock that set_mode resets the iterator
```
stanislav-tkach, line 103: Nitpick:
```suggestion
            // Check that set_mode resets the iterator
```

stanislav-tkach (approved):

aleksuss:
> Friendly ping.

Looks good. But I will merge it after #565

PR #650, opened 2022-06-28
Use NonNull in DBRawIteratorWithThreadMode

mina86:
The inner pointer in DBRawIteratorWithThreadMode is never null so mark
it as such.  With this change, Rust can optimise layout of the types
such that Option<DBRawIteratorWithThreadMode> is the same size as
DBRawIteratorWithThreadMode.  Previously it would be 8 bytes larger
(72 vs 64 bytes).

Ideally rocksdb_create_iterator would return NonNull but since it
doesn’t we need to go ahead and check for null explicitly.

### DIFF
--- a/src/db_iterator.rs
+++ b/src/db_iterator.rs
@@ -70,7 +70,7 @@ pub type DBRawIterator<'a> = DBRawIteratorWithThreadMode<'a, DB>;
 /// let _ = DB::destroy(&Options::default(), path);
 /// ```
 pub struct DBRawIteratorWithThreadMode<'a, D: DBAccess> {
-    inner: *mut ffi::rocksdb_iterator_t,
+    inner: std::ptr::NonNull<ffi::rocksdb_iterator_t>,
 
     /// When iterate_upper_bound is set, the inner C iterator keeps a pointer to the upper bound
     /// inside `_readopts`. Storing this makes sure the upper bound is always alive when the
@@ -82,12 +82,11 @@ pub struct DBRawIteratorWithThreadMode<'a, D: DBAccess> {
 
 impl<'a, D: DBAccess> DBRawIteratorWithThreadMode<'a, D> {
     pub(crate) fn new(db: &D, readopts: ReadOptions) -> Self {
-        unsafe {
-            Self {
-                inner: ffi::rocksdb_create_iterator(db.inner(), readopts.inner),
-                _readopts: readopts,
-                db: PhantomData,
-            }
+        let ptr = unsafe { ffi::rocksdb_create_iterator(db.inner(), readopts.inner) };
+        Self {
+            inner: std::ptr::NonNull::new(ptr).unwrap(),
+            _readopts: readopts,
+            db: PhantomData,
         }
     }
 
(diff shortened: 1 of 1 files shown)


stanislav-tkach (approved):
👍

src/db_iterator.rs
```diff
@@ -82,12 +82,11 @@ pub struct DBRawIteratorWithThreadMode<'a, D: DBAccess> {
...
     pub(crate) fn new(db: &D, readopts: ReadOptions) -> Self {
-        unsafe {
-            Self {
-                inner: ffi::rocksdb_create_iterator(db.inner(), readopts.inner),
-                _readopts: readopts,
-                db: PhantomData,
-            }
+        let ptr = unsafe { ffi::rocksdb_create_iterator(db.inner(), readopts.inner) };
+        Self {
+            inner: std::ptr::NonNull::new(ptr).unwrap(),
```
aleksuss, line 87: Is really this `unwrap` unsafe?

aleksuss: As I understand, would panic if the ptr is null here. I like the current version is much better. Because bellow `ffi::rocksdb_iter_valid(self.inner)` function checks for null pointer `self.inner` and if `self.inner` is null then iterator just is invalid? No panic.

aleksuss (changes requested):
See comment for code.

src/db_iterator.rs
mina86: My thinking here is that rocksdb_create_iterator and rocksdb_create_iterator_cf always return non-null, e.g.:

```c++
rocksdb_iterator_t* rocksdb_create_iterator(
    rocksdb_t* db,
    const rocksdb_readoptions_t* options) {
  rocksdb_iterator_t* result = new rocksdb_iterator_t;
  result->rep = db->rep->NewIterator(options->rep);
  return result;
}
```

and this merely takes advantage of this.  In practice there’s currently no way the unwrap to panic.

aleksuss: Potentially, `new` could return a `null` pointer. But in such case, the problem would be in the `rocksdb` library rather than in rust code. Definitely, you should leave some comments about this.

mina86: > new could return a null pointer.

Ah, I see the code is compiled with `-fno-exceptions`.  So yes, this should be separately fixed in the C wrapper functions and at this point the unwrap would behave like memory allocation error always does in Rust.  By the way:

> Because bellow ffi::rocksdb_iter_valid(self.inner) function checks for null pointer self.inner

This is not true:

```rust
unsigned char rocksdb_iter_valid(const rocksdb_iterator_t* iter) {
  return iter->rep->Valid();
}
```

None of the C wrapper functions check if the pointer is not-null.

aleksuss: Ok. You've convinced me 😄

aleksuss (approved):

src/db_iterator.rs
aleksuss: Thanks for your investigation 👍🏻

PR #652, opened 2022-06-29
Refactor iterator-related tests

mina86:
Firstly, introduce helper assert_iter and assert_iter_reversed
functions to reduce code repetition.  This will be especially useful
in upcoming commit which changes the Item type of the iterator.

Secondly, keep keys and values declared at the start of some of the
tests as slices without boxing them.  Doing the latter introduces
conveluted syntax such as `&*k1`; keeping them as slices simplifies
the syntax and also allows them to be converte into constsants.

Lastly, replace `cba` helper function with `pair` which construct
a pair of boxed slices.  This removes some repetition since after
the previous point `cba` was only ever used in a pair.

### DIFF
--- a/tests/util/mod.rs
+++ b/tests/util/mod.rs
@@ -38,3 +38,26 @@ impl AsRef<Path> for &DBPath {
         &self.path
     }
 }
+
+#[allow(dead_code)]
+pub fn pair(left: &[u8], right: &[u8]) -> (Box<[u8]>, Box<[u8]>) {
+    (Box::from(left), Box::from(right))
+}
+
+// Use macro rather than functions so that on failure we get line number where
+// the macro was instantiated rather than line number inside of a function.
+#[allow(unused_macros)]
+macro_rules! assert_iter {
+    ($iter:expr, $want:expr) => {
+        let got = $iter.collect::<Vec<_>>();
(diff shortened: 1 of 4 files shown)


tests/util/mod.rs
```diff
@@ -38,3 +38,26 @@ impl AsRef<Path> for &DBPath {
...
+
+#[allow(dead_code)]
+pub fn pair(left: &[u8], right: &[u8]) -> (Box<[u8]>, Box<[u8]>) {
+    (Box::from(left), Box::from(right))
+}
+
+// Use macro rather than functions so that on failure we get line number where
+// the macro was instantiated rather than line number inside of a function.
+#[allow(unused_macros)]
+macro_rules! assert_iter {
```
stanislav-tkach, line 50: Can this macro be replaced with a function with the `#[track_caller]` attribute?

stanislav-tkach (approved):

tests/util/mod.rs
mina86: Perfect. Done.

mina86:
Friendly ping.

mina86:
Sorry, didn’t notice I’d broken clippy. Fixed now.

aleksuss (approved):

PR #653, opened 2022-06-30
feat: enable IOS building

dignifiedquire:
Tested on my local device, assumes building on macos, which I believe is a requirement anyway.

Based on #637


dignifiedquire:
clippy failures seem unrelated

stanislav-tkach (approved):

aleksuss (approved):

PR #656, opened 2022-07-16
Introduce ReadOptions::set_iterate_range and PrefixRange

mina86:
Firstly, introduce ReadOptions::set_iterate_range method.  It accepts
range as an argument and in this way in an idiomatic way covers use
cases of set_iterate_lower_bound and set_iterate_upper_bound methods.
It also allows bounds to be cleared after they were initially set.

Secondly, and more crucially, introduce PrefixRange which can be used
as an argument to set_iterate_range.  When used, it sets the lower and
upper bounds so that they cover keys with given prefix only.  This
allows for easy implementation of manual prefix iteration.

### DIFF
--- a/src/db_options.rs
+++ b/src/db_options.rs
@@ -3302,38 +3302,83 @@ impl ReadOptions {
         }
     }
 
+    /// Sets the lower bound for an iterator.
+    pub fn set_iterate_lower_bound<K: Into<Vec<u8>>>(&mut self, key: K) {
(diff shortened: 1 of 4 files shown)


mina86:
This could deprecate set_iterate_lower_bound and set_iterate_upper_bound methods.

stanislav-tkach (approved):
Thank you for the pull request, especially for the effort with documentation!

aleksuss (approved):

Issue #660, opened 2022-07-20
Segfault in DBRawIteratorWithThreadMode::next()


niklasf:
Usage (not reduced to minimal example, yet): https://github.com/lichess-org/lila-openingexplorer/blob/02f29bd3d34ff0c3ed20c0985348e9f6535469a1/src/db.rs#L327-L338

Crashes deterministically.

Backtrace:

```
Thread 13 "tokio-runtime-w" received signal SIGSEGV, Segmentation fault.
[Switching to Thread 0x7ffff1dff640 (LWP 380709)]
__memcmp_avx2_movbe_rtm () at ../sysdeps/x86_64/multiarch/memcmp-avx2-movbe.S:412
412		vmovdqu	(%rsi), %ymm2
(gdb) bt
#0  __memcmp_avx2_movbe_rtm () at ../sysdeps/x86_64/multiarch/memcmp-avx2-movbe.S:412
#1  0x0000555556687a28 in rocksdb::Slice::compare (this=0x7ffff1deaa70, b=...) at rocksdb/include/rocksdb/slice.h:243
#2  0x0000555556969348 in rocksdb::(anonymous namespace)::BytewiseComparatorImpl::CompareWithoutTimestamp (this=0x7ffff71b7aa0, a=..., b=...)
    at rocksdb/util/comparator.cc:140
#3  0x00005555565825ac in rocksdb::UserComparatorWrapper::CompareWithoutTimestamp (this=0x7ffff1048488, a=..., a_has_ts=false, b=..., b_has_ts=false)
    at rocksdb/util/user_comparator_wrapper.h:73
#4  0x000055555668aa4a in rocksdb::DBIter::FindNextUserEntryInternal (this=0x7ffff1048440, skipping_saved_key=true, prefix=0x7ffff1deabf0)
    at rocksdb/db/db_iter.cc:278
#5  0x000055555668a897 in rocksdb::DBIter::FindNextUserEntry (this=0x7ffff1048440, skipping_saved_key=true, prefix=0x7ffff1deabf0)
    at rocksdb/db/db_iter.cc:225
#6  0x000055555668a4a4 in rocksdb::DBIter::Next (this=0x7ffff1048440) at rocksdb/db/db_iter.cc:159
#7  0x00005555569e4168 in rocksdb::ArenaWrappedDBIter::Next (this=0x7ffff1048400) at rocksdb/db/arena_wrapped_db_iter.h:67
[...]

PR #661, opened 2022-07-20
Revert "Don’t hold onto ReadOptions.inner when iterating"

mina86:
Turns out there’s a two-step reference chain for lower and upper bound.
The C++ ReadOptions refer slices inside of the rocksdb_readoptions_t
C wrapper which refer to Rust vectors.   Holding onto the vectors is
insufficient for all the references to be valid.

This reverts commit 947e4b63034fff2151f52c399a5d85925ea80911.

Fixes: https://github.com/rust-rocksdb/rust-rocksdb/issues/660


niklasf:
Confirmed that this fixes #660 as expected. Thanks for the super quick response.

stanislav-tkach (approved):

stanislav-tkach:
I had some doubts about that, but then I thought: the tests passed, so everything should be OK... 🙃

aleksuss (approved):

PR #668, opened 2022-08-04
Make possible to pass a list of cherry-picks/patches to rocksdb library through env variables

EdvardD:


librocksdb-sys/build.rs
```diff
@@ -357,9 +355,40 @@ fn update_submodules() {
...
 }
 
+fn update_submodules() {
+    let program = "git";
+    let dir = "../";
+    let args = ["submodule", "update", "--init"];
+    run_command(program, dir, &args);
+}
+
+fn cherry_pick_commits(commits: &Vec<&str>) {
```
akhi3030, line 365:
```suggestion
fn cherry_pick_commits(commits: &[&str]) {
```

librocksdb-sys/build.rs
```diff
@@ -357,9 +355,40 @@ fn update_submodules() {
...
+    for commit in commits {
+        let program = "git";
+        let dir = "rocksdb/";
+        let args = ["cherry-pick", &commit];
+        run_command(program, dir, &args);
+    }
+}
+
+fn apply_patches(patches: &Vec<&str>) {
```
akhi3030, line 374:
```suggestion
fn apply_patches(patches: &[&str]) {
```

librocksdb-sys/build.rs
```diff
@@ -357,9 +355,40 @@ fn update_submodules() {
...
+    for patch in patches {
+        let program = "git";
+        let dir = "rocksdb/";
+        let args = ["apply", &patch];
```
akhi3030, line 378:
```suggestion
        let args = ["apply", patch];
```

I haven't tried running it but I don't think the additional reference is needed.

librocksdb-sys/build.rs
```diff
@@ -357,9 +355,40 @@ fn update_submodules() {
...
+    let dir = "../";
+    let args = ["submodule", "update", "--init"];
+    run_command(program, dir, &args);
+}
+
+fn cherry_pick_commits(commits: &Vec<&str>) {
+    for commit in commits {
+        let program = "git";
+        let dir = "rocksdb/";
+        let args = ["cherry-pick", &commit];
```
akhi3030, line 369:
```suggestion
        let args = ["cherry-pick", commit];
```

librocksdb-sys/build.rs
EdvardD: I don't know to to get a slice from a split function

librocksdb-sys/build.rs
EdvardD: I don't know to to get a slice from a split function, it won't work just like this

akhi3030: `split` returns an iterator.  When you use `collect`, you need to sometimes provide type hints if the compiler is not able to infer the actual types.  Something like the following works.

```
fn foo(arg: &[&str]) {
    println!("{:?}", arg);
}


fn main() {
    let arg = "hello world";
    foo(&arg.split(" ").collect::<Vec<&str>>());
}
```

akhi3030: In general, slices are preferred over `Vec<T>` because slices are more generic i.e. more things can be converted into them so it makes your function able more inclusive.

niklasf:
Wouldn't it be easier to maintain an actual fork instead of a collection of patches? You probably have some specific patches in mind. Don't any of those require corresponding changes to the `rocksdb` crate?

akhi3030:
Thanks for the feedback @niklasf!  Edvard has now left the org, so we need to find someone else to pick up their open tasks.  We are having an active discussion on how to best handle this.  I think we can close this PR for now.  If we need to do further work here, we will get back to you folks.

aleksuss:
I close this PR because I think this feature is needless.

PR #677, opened 2022-08-23
make `create_new_backup_flush` generic

minshao:
This allows `create_new_backup` for `OptimisticTransactionDB<T: ThreadMode>`.


stanislav-tkach (approved):

MillionIntegrals:
I just found this issue right now when trying to use `BackupEngine` with `TransactionDB`. Is it in any way supported right now? 

In the C++ API we have this:
```
db = txn_db->GetBaseDB();
```

Which always can be used to get the underlying DB object.  Maybe that's all that we need here?

aleksuss (approved):

PR #682, opened 2022-08-30
Fix int128 compatibility check(Issue #681)

### LINKED ISSUE #681: Unsupport type '__uint128_t' when I used arm-linux-gnueabihf
When I cross-compile it by the compiler "arm-linux-gnueabihf", it reported that '__uint128_t' was not declared in this scope

```
The following warnings were emitted during compilation:

warning: In file included from rocksdb/util/hash.h:26:0,
warning:                  from rocksdb/cache/cache_key.cc:13:
warning: rocksdb/util/fastrange.h: 在静态成员函数‘static Range rocksdb::detail::FastRangeGenericImpl<long long unsigned int, Range>::Fn(uint64_t, Range)’中:
warning: rocksdb/util/fastrange.h:62:5: 错误： ‘__uint128_t’在此作用域中尚未声明
warning:      __uint128_t wide = __uint128_t{range} * hash;
warning:      ^~~~~~~~~~~
warning: rocksdb/util/fastrange.h:62:5: 附注： suggested alternative: ‘__uint32_t’
warning:      __uint128_t wide = __uint128_t{range} * hash;
warning:      ^~~~~~~~~~~
warning:      __uint32_t
warning: rocksdb/util/fastrange.h:62:45: 错误： 一元 ‘*’参数类型无效(有‘uint64_t {aka long long unsigned int}’)
warning:      __uint128_t wide = __uint128_t{range} * hash;
warning:                                              ^~~~
warning: rocksdb/util/fastrange.h:63:31: 错误： ‘wide’在此作用域中尚未声明
warning:      return static_cast<Range>(wide >> 64);
warning:                                ^~~~
warning: rocksdb/util/fastrange.h:63:31: 附注： suggested alternative: ‘fwide’
warning:      return static_cast<Range>(wide >> 64);
warning:                                ^~~~
warning:                                fwide

error: failed to run custom build command for `librocksdb-sys v0.8.0+7.4.4
```

#620 Add `config.define("HAVE_UINT128_EXTENSION", Some("1"));` to the [build.rs](https://github.com/rust-rocksdb/rust-rocksdb/blob/master/librocksdb-sys/build.rs#L236). It can cause problem.

In fact, not all compilers support the type of int128, so the [facebook/rocksdb#7338](https://github.com/facebook/rocksdb/pull/7338)  uses makefile to check int128 compatibility. The type of int128 is not related to target, but related to compiler。

So I think [ `config.define("HAVE_UINT128_EXTENSION", Some("1"));` ](https://github.com/rust-rocksdb/rust-rocksdb/blob/master/librocksdb-sys/build.rs#L236) should be deleted.
@niklasf

Dirreke:
Fix #681 

delete 'config.define("HAVE_UINT128_EXTENSION", Some("1"));' in [build.rs](https://github.com/rust-rocksdb/rust-rocksdb/blob/master/librocksdb-sys/build.rs#L236), which can cause Issue #681 

Not all compilers support the type of int128, so the https://github.com/facebook/rocksdb/pull/7338 uses makefile to check int128 compatibility. The type of int128 is not related to target, but related to compiler。

### DIFF
--- a/librocksdb-sys/build.rs
+++ b/librocksdb-sys/build.rs
@@ -233,7 +233,6 @@ fn build_rocksdb() {
     } else {
         config.flag(&cxx_standard());
         // matches the flags in CMakeLists.txt from rocksdb
-        config.define("HAVE_UINT128_EXTENSION", Some("1"));
         config.flag("-Wsign-compare");
         config.flag("-Wshadow");
         config.flag("-Wno-unused-parameter");


stanislav-tkach:
I'm not sure about simply deleting it. It seems better to enable it conditionally.

niklasf:
Definitely agree if there's a good way to feature-detect it. If not, deleting it should be fine. The gains (if any) are pretty minor.

Dirreke:
CARGO_CFG_TARGET_POINTER_WIDTH shows the bits of target cpu including  32bits or 64 bits. However, it cannt show whether target support the type of 'int128'. The type of '128bits' depends on the compiler, not target arch. Maybe we need to use other methods.



---- Replied Message ----
| From | Julien ***@***.***> |
| Date | 12/10/2022 01:12 |
| To | ***@***.***> |
| Cc | ***@***.***>***@***.***> |
| Subject | Re: [rust-rocksdb/rust-rocksdb] Fix int128 compatibility check(Issue #681) (PR #682) |

Definitely agree if there's a good way to feature-detect it

I think the CARGO_CFG_TARGET_POINTER_WIDTH environment variable could be used in build.rs :
https://doc.rust-lang.org/cargo/reference/environment-variables.html#environment-variables-cargo-sets-for-build-scripts

niklasf:
I added this as part of #620 because it was easy to add, without testing if there are actually any advantages to this particular flag. But it turns out there is a clear disadvantage to adding it unconditionally, so I think we should just remove it, i.e., merge this PR.

*If* someone wants to bring it back later, they can figure out a way to feature-detect this.

Dirreke:
Thanks!

PR #690, opened 2022-09-28
add WriteBatch::data

w41ter:


aleksuss:
Could you provide an example of the usage of the method, please? Also, public methods require doc-comments and it would be good to add some tests.

w41ter:
@aleksuss Hi, I have added a test case to describe the usage of `data()` .

aleksuss (commented):
I suppose the new functionality allows us to add `impl Clone for WriteBatch` implementation? 
cc @stanislav-tkach

stanislav-tkach (approved):

aleksuss (approved):

PR #694, opened 2022-10-11
make sure build.rs is re-run, when env vars change

drahnr:
Currently a change in linkage mode env vars will not trigger a re-run of the `build.rs`-binary, and hence compilations fail until the that binary is deleted or other triggers are manually triggered.

To my understanding, it's a mere lag of 3 lines telling cargo to re-run the build script on env var change.

Also adds the option to _NOT_ link statically if i.e. `ROCKSDB_STATIC=false` is provided

### DIFF
--- a/librocksdb-sys/build.rs
+++ b/librocksdb-sys/build.rs
@@ -300,11 +300,9 @@ fn try_to_find_and_link_lib(lib_name: &str) -> bool {
 
     if let Ok(lib_dir) = env::var(&format!("{}_LIB_DIR", lib_name)) {
         println!("cargo:rustc-link-search=native={}", lib_dir);
-        let mode = match env::var_os(&format!("{}_STATIC", lib_name))
-            .as_ref()
-        {
+        let mode = match env::var_os(&format!("{}_STATIC", lib_name)) {
             Some(_) => "static",
-            _ => "dylib",
+            None => "dylib",
         };
         println!("cargo:rustc-link-lib={}={}", mode, lib_name.to_lowercase());
         return true;
--- a/librocksdb-sys/build.rs
+++ b/librocksdb-sys/build.rs
@@ -299,7 +299,10 @@ fn try_to_find_and_link_lib(lib_name: &str) -> bool {
     println!("cargo:rerun-if-env-changed={}_STATIC", lib_name);
     if let Ok(lib_dir) = env::var(&format!("{}_LIB_DIR", lib_name)) {
         println!("cargo:rustc-link-search=native={}", lib_dir);
-        let mode = match env::var(&format!("{}_STATIC", lib_name)).as_ref().ok().map(|s| s.as_str()) {
+        let mode = match env::var_os(&format!("{}_STATIC", lib_name))
+            .as_ref()
+            .and_then(|s| s.to_str())
+        {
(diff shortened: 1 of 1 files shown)


librocksdb-sys/build.rs
```diff
@@ -288,17 +288,20 @@ fn build_snappy() {
...
 
+    println!("cargo:rerun-if-env-changed={}_LIB_DIR", lib_name);
+    println!("cargo:rerun-if-env-changed={}_STATIC", lib_name);
     if let Ok(lib_dir) = env::var(&format!("{}_LIB_DIR", lib_name)) {
         println!("cargo:rustc-link-search=native={}", lib_dir);
-        let mode = match env::var_os(&format!("{}_STATIC", lib_name)) {
-            Some(_) => "static",
-            None => "dylib",
+        let mode = match env::var(&format!("{}_STATIC", lib_name)).as_ref().ok().map(|s| s.as_str()) {
+            Some("") | Some("true") | Some("1") | Some("static") => "static",
```
aleksuss, line 303: 1. Does it really make sense to check all these variants? What if I set `ROCKSDB_STATIC=yes`? 
2. Why do you use `env::var`with `.ok()` rather than `env::var_os` ? 
```rust
let mode = match env::var_os(&format!("{}_STATIC", lib_name))
            .as_ref()
            .and_then(|s| s.to_str()) {
...
```
?

drahnr: Will change to `var_os`, I wasn't aware of the internal implementation of https://doc.rust-lang.org/src/std/env.rs.html#236 `fn var`.
Will also add `yes` as an enabling value.

The idea to accept only specific names or empty, is to allow the inverse usage as well: `ROCKSDB_STATIC=false` is something that is intuitive, when `ROCKSDB_STATIC=1` enables `static` linkage.

aleksuss:
> Also adds the option to NOT link statically if i.e. ROCKSDB_STATIC=false is provided

If you don't want to link statically just unset env variable. So. I'd leave logic of working with env variable `ROCKSDB_STATIC` as is.
cc @stanislav-tkach

drahnr:
I did a brief survey, and the dependencies are less coherent than I expect, tbh it's a mess.

https://github.com/rust-lang/libz-sys/blob/main/build.rs#L25
https://github.com/alexcrichton/xz2-rs/blob/master/lzma-sys/build.rs#L12
https://github.com/alexcrichton/bzip2-rs/blob/master/bzip2-sys/build.rs#L15-L24
https://github.com/10XGenomics/lz4-rs/blob/master/lz4-sys/build.rs#L43
https://github.com/gyscos/zstd-rs/blob/master/zstd-safe/zstd-sys/build.rs#L22-L25

Hence, I don't care about the solution, it'd be nice to have the duality that at least _some_ references above exhibit in regards to value `1` or add a feature that ensures static compilation and linkage of all deps. The latter is out of scope, the former is debatable.

Note that with the current change, the current behavior of having an `ROCKSB_STATIC=""` resulting in `static`-linkage is preserved.

stanislav-tkach:
I think that changing the `ROCKSDB_STATIC` environment variable behavior is technically a breaking change.

aleksuss (approved):

stanislav-tkach (approved):

PR #698, opened 2022-10-19
Support Write Buffer Manager

zhuwenshen:
Support Write Buffer Manager


aleksuss:
No. It's a bad practice to add custom source files here. E.g. in the case of the usage system rocksdb library. Create the PR to [facebook/rockcdb](https://github.com/facebook/rocksdb) first, please.

PR #700, opened 2022-10-28
Support RocksDB 7.x BackupEngineOptions

exabytes18:
The main objective of this PR is to support `rocksdb_backup_engine_options_t` (introduced in https://github.com/facebook/rocksdb/commit/78aee6fedc62ca6048e500969912d1e1f0553490) so that we can call the associated `rocksdb_backup_engine_options_set_max_background_operations` function and achieve higher parallelism for backup and restore operations (notably, the current parallelism is 1, meaning that rocksdb will only copy/checksum/hardlink/whatever 1 file at a time, which is insufficient for rocksdb to full utilize hardware on even modest machines).

One side-effect of this PR is that the signature for BackupEngine::open changes from `open<P: AsRef<Path>>(opts: &BackupEngineOptions, path: P)` to `open(opts: &BackupEngineOptions, env: &Env)`. This may cause some consumer code to fail to compile, but it aligns better with the RocksDB-proper `BackupEngine::Open()` signature, so I believe it's a worthwhile change.

Finally, I'm not well-versed in ffi or the related patterns (like `_outlive`), so please carefully review those bits in the PR.

Thanks!
Matt

### DIFF
--- a/src/backup.rs
+++ b/src/backup.rs
@@ -13,9 +13,11 @@
 // limitations under the License.
 //
 
+use crate::env::Env;
 use crate::{db::DBInner, ffi, ffi_util::to_cpath, DBCommon, Error, ThreadMode};
 
-use libc::{c_int, c_uchar};
+use libc::c_uchar;
+use std::ffi::CString;
 use std::path::Path;
 
 /// Represents information of a backup including timestamp of the backup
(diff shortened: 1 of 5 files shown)


exabytes18:
@aleksuss, @stanislav-tkach, could you help to review this PR?

exabytes18:
@aleksuss, I've fixed a clippy warning. Could you approve the workflow run again?

exabytes18:
@aleksuss, could you take a look at the PR when you next have a chance?

stanislav-tkach (approved):

exabytes18:
@aleksuss, @stanislav-tkach, could you help to merge this?

src/backup.rs
```diff
@@ -217,27 +224,40 @@ impl BackupEngine {
 }
 
 impl BackupEngineOptions {
-    //
-}
+    pub fn new<P: AsRef<Path>>(backup_dir: P) -> Result<Self, Error> {
```
aleksuss, line 227: Please, add some description to the public methods.

exabytes18: Done. Please have another look.

aleksuss (approved):

PR #721, opened 2022-12-17
Add drop_cf function to TransactionDB

bothra90:
Since `TransactionDB` is a subclass of `StackableDB` which is subclass of `DB`, there's no reason for the `drop_cf` function should be callable directly from TransactionDB.

The core of this change is a cast in `transaction_db.rs`:
```rust
self.inner as *mut ffi::rocksdb_t,
```

The PR also includes a test, but that's mostly copy-pasted from a similar test on `DB`, with the addition of testing for error when opening a previously deleted column family to make sure that the drop was actually successful.

### DIFF
--- a/src/transactions/transaction_db.rs
+++ b/src/transactions/transaction_db.rs
@@ -935,7 +935,7 @@ impl<T: ThreadMode> TransactionDB<T> {
     fn drop_column_family<C>(
         &self,
         cf_inner: *mut ffi::rocksdb_column_family_handle_t,
-        cf: C,
+        _cf: C,
     ) -> Result<(), Error> {
         unsafe {
             // first mark the column family as dropped
@@ -944,9 +944,8 @@ impl<T: ThreadMode> TransactionDB<T> {
                 cf_inner
             ));
         }
-        // then finally reclaim any resources (mem, files) by destroying the only single column
-        // family handle by drop()-ing it
-        drop(cf);
+        // Since `_cf` is dropped here, the column family handle is destroyed
+        // and any resources (mem, files) are reclaimed.
         Ok(())
     }
 }
@@ -971,7 +970,7 @@ impl TransactionDB<SingleThreaded> {
         if let Some(cf) = self.cfs.cfs.remove(name) {
             self.drop_column_family(cf.inner, cf)
         } else {
-            Err(Error::new(format!("Invalid column family: {}", name)))
+            Err(Error::new(format!("Invalid column family: {name}")))
         }
     }
 }
@@ -1004,7 +1003,7 @@ impl TransactionDB<MultiThreaded> {
         if let Some(cf) = self.cfs.cfs.write().unwrap().remove(name) {
             self.drop_column_family(cf.inner, cf)
         } else {
-            Err(Error::new(format!("Invalid column family: {}", name)))
+            Err(Error::new(format!("Invalid column family: {name}")))
         }
     }
 }
--- a/src/transactions/transaction_db.rs
+++ b/src/transactions/transaction_db.rs
@@ -931,6 +931,24 @@ impl<T: ThreadMode> TransactionDB<T> {
     pub fn snapshot(&self) -> SnapshotWithThreadMode<Self> {
         SnapshotWithThreadMode::<Self>::new(self)
     }
+
+    fn drop_column_family<C>(
+        &self,
+        cf_inner: *mut ffi::rocksdb_column_family_handle_t,
+        cf: C,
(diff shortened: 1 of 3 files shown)


bothra90:
@aleksuss: Could you take a look, please?

src/transactions/transaction_db.rs
```diff
@@ -947,6 +965,15 @@ impl TransactionDB<SingleThreaded> {
     pub fn cf_handle(&self, name: &str) -> Option<&ColumnFamily> {
         self.cfs.cfs.get(name)
     }
+
+    /// Drops the column family with the given name
+    pub fn drop_cf(&mut self, name: &str) -> Result<(), Error> {
+        if let Some(cf) = self.cfs.cfs.remove(name) {
+            self.drop_column_family(cf.inner, cf)
+        } else {
+            Err(Error::new(format!("Invalid column family: {}", name)))
```
aleksuss, line 974: It makes clippy happy
```suggestion
            Err(Error::new(format!("Invalid column family: {name}")))
```

tests/test_column_family.rs
```diff
@@ -112,6 +116,132 @@ fn test_column_family() {
...
+    {
+        let opts = Options::default();
+        let vec = DB::list_cf(&opts, &n);
+        match vec {
+            Ok(vec) => assert_eq!(vec, vec![DEFAULT_COLUMN_FAMILY_NAME, "cf1"]),
+            Err(e) => panic!("failed to drop column family: {}", e),
+        }
+    }
+
+    // TODO should be able to use writebatch ops with a cf
```
aleksuss, line 194: Remove it if you don't have plans to do it.

src/transactions/transaction_db.rs
```diff
@@ -970,6 +997,16 @@ impl TransactionDB<MultiThreaded> {
...
             .map(UnboundColumnFamily::bound_column_family)
     }
+
+    /// Drops the column family with the given name by internally locking the inner column
+    /// family map. This avoids needing `&mut self` reference
+    pub fn drop_cf(&self, name: &str) -> Result<(), Error> {
+        if let Some(cf) = self.cfs.cfs.write().unwrap().remove(name) {
+            self.drop_column_family(cf.inner, cf)
+        } else {
+            Err(Error::new(format!("Invalid column family: {}", name)))
```
aleksuss, line 1007: And here.

aleksuss (commented):
Overall looks good, with a couple of nitpicks

src/transactions/transaction_db.rs
```diff
@@ -931,6 +931,24 @@ impl<T: ThreadMode> TransactionDB<T> {
...
+        unsafe {
+            // first mark the column family as dropped
+            ffi_try!(ffi::rocksdb_drop_column_family(
+                self.inner as *mut ffi::rocksdb_t,
+                cf_inner
+            ));
+        }
+        // then finally reclaim any resources (mem, files) by destroying the only single column
+        // family handle by drop()-ing it
+        drop(cf);
```
stanislav-tkach, line 949: Is this only for readability? The handle will be dropped anyway. Personally I would leave the comment and remove the `drop` function call, but it is up to you.

stanislav-tkach (approved):

src/transactions/transaction_db.rs
bothra90: done

src/transactions/transaction_db.rs
```diff
@@ -970,6 +997,16 @@ impl TransactionDB<MultiThreaded> {
             .cloned()
             .map(UnboundColumnFamily::bound_column_family)
     }
+
+    /// Drops the column family with the given name by internally locking the inner column
+    /// family map. This avoids needing `&mut self` reference
+    pub fn drop_cf(&self, name: &str) -> Result<(), Error> {
+        if let Some(cf) = self.cfs.cfs.write().unwrap().remove(name) {
+            self.drop_column_family(cf.inner, cf)
```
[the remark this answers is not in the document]
bothra90: good point! I think this will (correctly) return an error if the user drops a column family from one thread while continuing to use it from another. Overall seems safe.

bothra90:
@stanislav-tkach: I've updated the PR with a fix for the clippy error.

stanislav-tkach (approved):

aleksuss (approved):

bothra90:
Thanks for the review, @aleksuss. Since I don't have write access to the repo, you'll need to merge the PR.

stanislav-tkach (approved):

Issue #568, opened 2021-11-17
Load RocksDB option file


Congyuwang:
https://github.com/facebook/rocksdb/blob/main/include/rocksdb/utilities/options_util.h#L73

Is there any way to load DB options from the RocksDB option file?

This allows users to open existing rocksDB without configuring the compatible options manually.

akrylysov:
rust-rocksdb uses [C RocksDB API](https://github.com/facebook/rocksdb/blob/dc5de45af8fc1bcc90edea5cd9a5695bd85a8baf/include/rocksdb/c.h). It doesn't seem the C API provides a way to load options from a file. To add this feature to rust-rocksdb you'll need to extend the C RocksDB API first.

akrylysov:
I took another look at the C API and noticed there is [rocksdb_get_options_from_string](https://github.com/facebook/rocksdb/blob/2a67d475f1a3fe24b4baae6e590cecabca099464/db/c.cc#L4791), it might be close to what you are looking for.

Congyuwang:
Yes close to. It uses such a format as string input to get options: [StringToMap](https://github.com/facebook/rocksdb/blob/dc5de45af8fc1bcc90edea5cd9a5695bd85a8baf/options/options_helper.cc#L560)
```
  // Example:
  //   opts_str = "write_buffer_size=1024;max_write_buffer_number=2;"
  //              "nested_opt={opt1=1;opt2=2};max_bytes_for_level_base=100"
```
in the implementation of [GetDBOptionsFromString](https://github.com/facebook/rocksdb/blob/dc5de45af8fc1bcc90edea5cd9a5695bd85a8baf/options/options_helper.cc#L748).

whereas the automatically dumped `OPTIONS-0000x` file looks like:
```
[DBOptions]
  max_background_flushes=-1
  compaction_readahead_size=0
  strict_bytes_per_sync=false
  wal_bytes_per_sync=0
  max_open_files=-1
...
[TableOptions/PlainTable "default"]
  store_index_in_file=false
  full_scan_mode=false
  encoding_type=kPlain
  hash_table_ratio=0.750000
```
The more convenient [LoadLatestOptions](https://github.com/facebook/rocksdb/blob/main/include/rocksdb/utilities/options_util.h#L73) however is not exposed as C API.

Congyuwang:
https://github.com/facebook/rocksdb/issues/9233

Congyuwang:
[...]

PR #724, opened 2023-01-04
Add Options::load_latest() method to load latest options from RockDB path

Congyuwang:
Load latest options from RocksDB path.

close PR #701 and issue #568.

### DIFF
--- a/src/db_options.rs
+++ b/src/db_options.rs
@@ -14,6 +14,8 @@
 
 use std::ffi::CStr;
 use std::path::Path;
+use std::ptr::null_mut;
+use std::slice;
 use std::sync::Arc;
 
 use libc::{self, c_char, c_double, c_int, c_uchar, c_uint, c_void, size_t};
@@ -24,12 +26,12 @@ use crate::{
     comparator::{self, ComparatorCallback, CompareFn},
     db::DBAccess,
     ffi,
-    ffi_util::{to_cpath, CStrLike},
+    ffi_util::{from_cstr, to_cpath, CStrLike},
     merge_operator::{
         self, full_merge_callback, partial_merge_callback, MergeFn, MergeOperatorCallback,
     },
     slice_transform::SliceTransform,
-    Error, SnapshotWithThreadMode,
+    ColumnFamilyDescriptor, Error, SnapshotWithThreadMode,
 };
 
 fn new_cache(capacity: size_t) -> *mut ffi::rocksdb_cache_t {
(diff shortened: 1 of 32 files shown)


Congyuwang:
This allows loading column families directly from an unknown RocksDB directory, w/o having to manually configure an Options object. Anyone take a look?

tests/test_rocksdb_options.rs
```diff
@@ -16,9 +16,32 @@ mod util;
...
+        Env::new().unwrap(),
+        true,
+        Cache::new_lru_cache(1024 * 8).unwrap(),
+    )
+    .unwrap();
+    assert!(cfs.iter().any(|cf| cf.name() == "default"));
+    assert!(cfs.iter().any(|cf| cf.name() == "cf0"));
+    assert!(cfs.iter().any(|cf| cf.name() == "cf1"));
+}
+
```
Congyuwang, line 44: This is a simple test added for testing the load_latest api.

src/db_options.rs
```diff
@@ -875,6 +877,76 @@ pub enum LogLevel {
...
+        let column_descriptors = column_family_names_iter
+            .zip(column_family_options_iter)
+            .map(|(name, options)| ColumnFamilyDescriptor { name, options })
+            .collect::<Vec<_>>();
+        // free pointers
+        slice::from_raw_parts(column_family_names, num_column_families)
+            .iter()
+            .for_each(|ptr| libc::free(*ptr as *mut c_void));
+        libc::free(column_family_names as *mut c_void);
+        libc::free(column_family_options as *mut c_void);
```
stanislav-tkach, line 946: I think that `rocksdb_free` should be used here.

Congyuwang: https://github.com/facebook/rocksdb/blob/0a2d3b663a962cf86af5ad82c80baf1cd71f7766/db/c.cc#L6231
I think both are ok.

Congyuwang: I think here we are just dealing with c strings and two arrays. Options pointers are not destroyed since they are still used in returned values. So, libc::free should work. Don't know how rocksdb_free behaves on arrays and strings though.

aleksuss (approved):

src/db_options.rs
aleksuss: I'd also prefer to use `rocksdb_free` here.

Congyuwang:
- Changed `libc::free()` to `ffi::rocksdb_free()`
- Added some description to `ColumnFamilyDescriptor::name()`

stanislav-tkach (approved):

aleksuss (approved):

PR #740, opened 2023-02-10
Allow 1.60 Rust compiler by minor code reorg

### LINKED ISSUE #739: MSRV is not correct for 0.20 version. 0.20 uses let_else that introduced in Rust 1.65 only
Current MSRV in Readme is 1.60
Compiling rust-rocksdb with it result in an error:
```
   Compiling rocksdb v0.20.0
error[E0658]: `let...else` statements are unstable
   --> /root/.cargo/registry/src/github.com-1ecc6299db9ec823/rocksdb-0.20.0/src/transactions/transaction_db.rs:362:9
    |
362 | /         let Ok(cf_name) = CString::new(name.as_bytes()) else {
363 | |             return Err(Error::new(
364 | |                 "Failed to convert path to CString when creating cf".to_owned(),
365 | |             ));
366 | |         };
    | |__________^
    |
For more information about this error, try `rustc --explain E0658`.
error: could not compile `rocksdb` due to previous error
```

Either MSRV should be updated or code should be changed to allow 1.60 compiler.

YuraKotov:
Fixes #739

### DIFF
--- a/src/transactions/transaction_db.rs
+++ b/src/transactions/transaction_db.rs
@@ -359,19 +359,19 @@ impl<T: ThreadMode> TransactionDB<T> {
         name: &str,
         opts: &Options,
     ) -> Result<*mut ffi::rocksdb_column_family_handle_t, Error> {
-        let Ok(cf_name) = CString::new(name.as_bytes()) else {
-            return Err(Error::new(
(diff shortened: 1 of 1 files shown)


aleksuss:
Closed in favour #741

PR #741, opened 2023-02-10
Check in CI compatibility with MSRV

aleksuss:


stanislav-tkach (approved):
Should [`rust-version`](https://doc.rust-lang.org/cargo/reference/manifest.html#the-rust-version-field) be added to the config?

aleksuss:
I thought about adding `rust-toolchain` file with the rust version, but this file is annoying me sometime 😄

PR #751, opened 2023-02-26
Actually enable io-uring

niklasf:
I just tried and failed to actually use io-uring with RocksDB. Turns out, in addition to the compiler flag, users also have to define the `RocksDbIOUringEnable` symbol.

Does nothing if `ROCKSDB_IOURING_PRESENT` is not defined (when the `io-uring` feature flag is off).

https://github.com/facebook/rocksdb/blob/3c9eed688ec912e0c5526a6c1ef23948b50c59b3/env/fs_posix.cc#L76

### DIFF
--- a/librocksdb-sys/build_version.cc
+++ b/librocksdb-sys/build_version.cc
@@ -23,6 +23,10 @@ static const std::string rocksdb_build_date = "rocksdb_build_date:2023-02-19 21:
 
 std::unordered_map<std::string, ROCKSDB_NAMESPACE::RegistrarFunc> ROCKSDB_NAMESPACE::ObjectRegistry::builtins_ = {};
 
+extern "C" bool RocksDbIOUringEnable() {
+  return true;
+}
+
 namespace ROCKSDB_NAMESPACE {
 static void AddProperty(std::unordered_map<std::string, std::string> *props, const std::string& name) {
   size_t colon = name.find(":");
--- a/src/db_options.rs
+++ b/src/db_options.rs
@@ -3441,6 +3441,11 @@ impl ReadOptions {
         }
     }
 
+    /// Asynchronously prefetch some data.
+    ///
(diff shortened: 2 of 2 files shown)


aleksuss (approved):

stanislav-tkach (approved):

zaidoon1:
I've been wanting to test io-uring but don't we need to add the async_io to the read options (it was exposed via the c api in https://github.com/facebook/rocksdb/commit/ff04fb154bd74fd0681baa83b478095207e2719d)?

niklasf:
For most [interesting things](https://rocksdb.org/blog/2022/10/07/asynchronous-io-in-rocksdb.html) that will come with 8.x, yes. There's a little bit of io uring usage already, for multi-read.

aleksuss:
@niklasf resolve the conflict, please.

niklasf:
Rebased. Now that we have 8.x, I also added `set_async_io()` that @zaidoon1 mentioned.

zaidoon1:
> Rebased. Now that we have 8.x, I also added `set_async_io()` that @zaidoon1 mentioned.

thanks! I wonder if we should also add `optimize_multiget_for_io`. Per https://rocksdb.org/blog/2022/10/07/asynchronous-io-in-rocksdb.html:

`
A new flag in ReadOptions, async_io, controls the usage of async IO. This flag, when set, enables async IO in Iterators and MultiGet. For MultiGet, an additional ReadOptions flag, optimize_multiget_for_io (defaults to true), controls how aggressively to use async IO. If the flag is not set, files in the same level are read in parallel but not different levels. If the flag is set, the level restriction is removed and as many files as possible are read in parallel, regardless of level. The latter might have a higher CPU cost depending on the workload.
`

This can be a follow up of course.

niklasf:
Just checked and was disappointed that there is no C API yet, but actually the default is `true` in any case (though it only takes effect if `async_io` is enabled).

PR #756, opened 2023-03-03
Allow specifying checksum type

romanz:
Following https://github.com/facebook/rocksdb/pull/10778, this change will allow creating backward-compatible RocksDB tables (that can be read using older versions of RocksDB, without XXH3 support).


stanislav-tkach (approved):

aleksuss (approved):

romanz:
Thanks!

PR #759, opened 2023-03-22
Retrieve Value from `KeyMayExist` if value found in Cache or Memory

Congyuwang:
Add a method `key_may_exist_cf_opt_value()` that also returns value if the value is found compared to `key_may_exist_cf_opt()`.

### DIFF
--- a/src/db.rs
+++ b/src/db.rs
@@ -1272,6 +1272,48 @@ impl<T: ThreadMode, D: DBInner> DBCommon<T, D> {
         }
     }
 
+    // If the key definitely does not exist in the database, then this method
+    // returns false, else true. If the caller wants to obtain value when the key
+    // is found in memory, a bool for 'value_found' must be passed. 'value_found'
+    // will be true on return if value has been set properly.
+    // This check is potentially lighter-weight than invoking DB::Get(). One way
+    // to make this lighter weight is to avoid doing any IOs.
+    pub fn key_may_exist_cf_opt_value<K: AsRef<[u8]>>(
+        &self,
+        cf: &impl AsColumnFamilyRef,
+        key: K,
+        readopts: &ReadOptions,
+    ) -> (bool, Option<Box<[u8]>>) {
+        let key = key.as_ref();
+        let mut val: *mut c_char = ptr::null_mut();
+        let mut val_len: usize = 0;
+        let mut value_found: c_uchar = 0;
+        let may_exists = 0
+            != unsafe {
+                ffi::rocksdb_key_may_exist_cf(
+                    self.inner.inner(),
+                    readopts.inner,
+                    cf.inner(),
+                    key.as_ptr() as *const c_char,
+                    key.len() as size_t,
+                    &mut val,         /*value*/
+                    &mut val_len,     /*val_len*/
+                    ptr::null(),      /*timestamp*/
+                    0,                /*timestamp_len*/
+                    &mut value_found, /*value_found*/
+                )
+            };
+        // The value is only allocated (using malloc) and returned if it is found and
+        // value_found isn't NULL. In that case the user is responsible for freeing it.
+        if value_found == 0 {
+            (may_exists, None)
+        } else {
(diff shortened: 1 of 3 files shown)


src/db.rs
```diff
@@ -1272,6 +1272,48 @@ impl<T: ThreadMode, D: DBInner> DBCommon<T, D> {
...
+                    &mut val,         /*value*/
+                    &mut val_len,     /*val_len*/
+                    ptr::null(),      /*timestamp*/
+                    0,                /*timestamp_len*/
+                    &mut value_found, /*value_found*/
+                )
+            };
+        // The value is only allocated (using malloc) and returned if it is found and
+        // value_found isn't NULL. In that case the user is responsible for freeing it.
+        if may_exists && value_found != 0 {
```
Congyuwang, line 1308: According to the implementation [DBImpl::KeyMayExist](https://github.com/facebook/rocksdb/blob/b92bc04ab000d4e4d64605ca8123f41940d4a5e7/db/db_impl/db_impl.cc#L3322-L3346), value_found is only falsified if key-may-exist but can't fetch value. So, when key-may-not-exist, value_found is still true. Thus we need `may_exists && value_found != 0` to return `Some(value)`

src/db.rs
```diff
@@ -1272,6 +1272,48 @@ impl<T: ThreadMode, D: DBInner> DBCommon<T, D> {
...
+            let value =
+                unsafe { Box::from_raw(slice::from_raw_parts_mut(val as *mut u8, val_len)) };
```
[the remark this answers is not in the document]
niklasf: Looks like a good solution. I guess the only real alternative would be creating an owned copy, which isn't great.

src/ffi_util.rs
```diff
@@ -195,6 +195,25 @@ impl<'a> CStrLike for &'a CString {
     }
 }
 
+pub struct CSlice {
+    pub data: *const c_char,
```
niklasf, line 199: The members can not be public, because they could be replaced with invalid pointers / lengths in safe code (creating bad pointer is safe, dereferencing is not), while our unsafe code relies on their validity.

niklasf (changes requested):

src/ffi_util.rs
Congyuwang: Yes. I've fixed that. (What a mistake.)

src/ffi_util.rs
```diff
@@ -198,8 +198,14 @@ impl<'a> CStrLike for &'a CString {
...
 /// Do not derive `Clone` for this because it will cause double-free.
 pub struct CSlice {
-    pub data: *const c_char,
-    pub len: size_t,
+    data: *const c_char,
+    len: size_t,
+}
+
+impl CSlice {
+    pub(crate) fn new(data: *const c_char, len: size_t) -> Self {
```
niklasf, line 206: All good now, as far as library users are concerned, which is ultimately all that matters. Could still make this unsafe, so that even internally constructing CSlice's requires appropriate care.

niklasf (approved):
Looking good!

src/ffi_util.rs
Congyuwang: I've added some more comment on the constructor, and also marked it as unsafe, which gives more warning to internal uses.

niklasf (approved):

Congyuwang:
Can some reviewer merge this?

src/db.rs
```diff
@@ -1272,6 +1273,49 @@ impl<T: ThreadMode, D: DBInner> DBCommon<T, D> {
         }
     }
 
+    // If the key definitely does not exist in the database, then this method
```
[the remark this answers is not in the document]
Congyuwang: Hi! This comment was originally copied from rocksdb header file. I've modified it a bit to suit this particular method and converted it into a doc.

stanislav-tkach (approved):

aleksuss (approved):

PR #760, opened 2023-03-22
Enable librocksdb-sys to be built by rustc_codegen_cranelift

ZePedroResende:
When attempting to build `librocks-sys` using `rustc_codegen_cranelift`, you encounter a panic error caused by an `unwrap` function in the `build.rs` file. The error occurs because the `unwrap` function is attempting to access the `CARGO_CFG_TARGET_FEATURE` environment variable, which is not available in the cranelift platform.

This PR adds a condition to check that the env variable actually exists instead of unwrapping it.

### DIFF
--- a/librocksdb-sys/build.rs
+++ b/librocksdb-sys/build.rs
@@ -101,12 +101,15 @@ fn build_rocksdb() {
         .filter(|file| !matches!(*file, "util/build_version.cc"))
         .collect::<Vec<&'static str>>();
 
-    if target.contains("x86_64") {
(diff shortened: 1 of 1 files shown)


stanislav-tkach (approved):

aleksuss (approved):

ZePedroResende:
Hey guys , the CI failed on the MacOS tests , but it doesn't seem related to the changes introduced by this PR. 

Can you take a look at it or give me some idea on how to fix it ?

PR #761, opened 2023-03-27
Update to RocksDB 8.0.0

niklasf:
Minimal update, without adding support for any of the new features.

* Compressed block cache removed.
* RocksDB Lite removed.

https://github.com/facebook/rocksdb/releases/tag/v8.0.0


aleksuss (approved):
@niklasf, thank you for the contribution. Nice work 🚀

PR #765, opened 2023-04-07
support for comparators as closures

pegesund:
I have updated the code for supporting comparators as closures.

Reason for this is that moving from functions to closures makes it possible to add more information into the comparator function. Let us for example say one is creating a sql-base and then would like to add a compound index, based on the create index statement. The sort order in the comparator should then depend on the directions of the sorting fields, for example first sort on a float and then an int.

The old solution did not support a way to add meta-info like this into the compare function in a dynamical way and this is currently solved by allowing the comparator to be a closure. By using the move-functionality with closures we can add extra context to the compare function.

The solution is backwards compatible, as it still supports functions as these can be casted to closures in Rust.

I have also added a test-file for comparators where I test backwards compatiblity, and the new solution with support for closures.

Feel free to drop a line to me on petter.egesund@gmail.com if you would like to discuss.

### DIFF
--- a/src/comparator.rs
+++ b/src/comparator.rs
@@ -18,11 +18,11 @@ use std::cmp::Ordering;
 use std::ffi::CString;
 use std::slice;
 
-pub type CompareFn = fn(&[u8], &[u8]) -> Ordering;
+pub type CompareFn = dyn Fn(&[u8], &[u8]) -> Ordering;
 
 pub struct ComparatorCallback {
     pub name: CString,
-    pub f: CompareFn,
+    pub f: Box<CompareFn>,
 }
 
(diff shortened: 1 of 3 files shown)


stanislav-tkach (approved):

Congyuwang:
Don't know if this will incur performance overhead.

pegesund:
Hi @Congyuwang 

I do not think there should be a performance overhead, it will still only be a single function call after the rust compiler has done its job. It will however be a speed gain when having compound comparators. In my old solution I had to keep types before every value in my bytearray so be able to acomplish this - this took much disk space, cpu and memory.

Issue #331, opened 2019-09-10
Wrap the GetUpdatesSince() function


nlfiedler:
This function relies on the sequence number feature of RocksDB, and returns an iterator that yields every update made since that particular sequence. Since I saw how to add a function to rust-rocksdb earlier, I may submit a PR for this when the time comes (I'm not in a rush, just wanted to get this filed for now). Thanks for bringing RocksDB to Rust, it's been awesome.

vitvakatu:
Great, feel free to submit a PR!

nlfiedler:
Tried looking at how this could be added. The function in question takes a new kind of options (`rocksdb_wal_readoptions_t`) and returns a new kind of iterator (`rocksdb_wal_iterator_t`). From what I can tell, the options have no implementation of any sort, and are optional given the implementation of `rocksdb_get_updates_since()`. The iterator does not appear to have a factory function, and its API is so different I can't even tell how to get the keys and values from it.

vitvakatu:
You probably should look at the version 6.1.2 of the rocksb, cause we have a PR that will probably be merged soon (#328). By the way, this PR compiles on Linux and Mac, so you can try it yourself to look through generated code.

You can use `cargo doc --open` to get a full list of functions and types generated by the binding in `librocksdb-sys` crate. You can also open a file `target/debug/build/librocksdb-sys-*/out/bindings.rs` to read through generated code.

Regarding `GetUpdatesSince`: looks like in 6.1.2 read_options contains only a single boolean: https://github.com/facebook/rocksdb/blob/2b38e2dd6602a17a2010308580fd5d8c91dea650/include/rocksdb/transaction_log.h#L113

A returned iterator can be used with a functions listed in the same file: https://github.com/facebook/rocksdb/blob/2b38e2dd6602a17a2010308580fd5d8c91dea650/include/rocksdb/transaction_log.h

To get values from it you need to call GetBatch() (called `rocksdb_wal_iter_get_batch` in bindings).

vitvakatu:
Perhaps you can create an issue in facebook/rocksdb repo and ask how to solve your problem in a ideomatic way?

nlfiedler:
[...]

PR #771, opened 2023-04-18
Fix bug in `DBWALIterator` that would miss updates

Zagitta:
I found a bug in the `DBWALIterator` that would cause it to miss updates in certain scenarios.
One such scenario is calling `latest_sequence_number()` on a newly created DB which returns `0`, then writing a couple of things and calling `get_updates_since(0)`. This would cause the first batch returned by the iterator to be for sequence 2 and not 1 as expected.
The test case I added demonstrates this and fails without my fix.
I also suspect this would cause issues in scenarios where a given sequence number is in the past but no longer exists which causes RocksDB to return an iterator that starts after the sequence number based on the following documentation:
>Sets iter to an iterator that is positioned at a write-batch whose sequence number range [start_seq, end_seq] covers seq_number. If no such write-batch exists, then iter is positioned at the next write-batch whose start_seq > seq_number.

I'm not 100% sure if that scenario is possible but it's a bug nevertheless.

I've fixed the issue by checking if each returned batch's sequence number matches the one provided to `get_changes_since` and only if so, skip it. This should be safe as sequence numbers are monotonically increasing. 

I hope this can be merged and released ASAP as I need it for my project 😄 

Thank you.

### DIFF
--- a/src/db.rs
+++ b/src/db.rs
@@ -1845,7 +1845,10 @@ impl<T: ThreadMode, D: DBInner> DBCommon<T, D> {
                 seq_number,
                 opts
             ));
-            Ok(DBWALIterator { inner: iter })
+            Ok(DBWALIterator {
+                inner: iter,
+                start_seq_number: seq_number,
+            })
         }
     }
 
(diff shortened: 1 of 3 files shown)


Zagitta:
Rather unfortunately this seems to have been noticed way back in 2019 already: https://github.com/rust-rocksdb/rust-rocksdb/issues/331#issuecomment-918753782 by @tancehao but never fixed or followed up on.
I guess @nlfiedler might also be interested to know about this data loss 😅

stanislav-tkach (approved):
Looking good to me. 👍

zaidoon1:
I would love it if this fix lands 🙏

aleksuss (approved):

PR #779, opened 2023-05-12
db_options: Expose remaining PlainTableFactoryOptions

glittershark:
This is intended to go along with
https://github.com/facebook/rocksdb/pull/11442.

### DIFF
--- a/src/db_options.rs
+++ b/src/db_options.rs
@@ -2446,6 +2446,10 @@ impl Options {
                 options.bloom_bits_per_key,
                 options.hash_table_ratio,
                 options.index_sparseness,
+                options.huge_page_tlb_size,
+                options.encoding_type as _,
+                options.full_scan_mode as _,
+                options.store_index_in_file as _,
             );
         }
     }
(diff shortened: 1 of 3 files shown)


src/db_options.rs
```diff
@@ -3578,6 +3582,27 @@ pub enum ChecksumType {
     XXH3 = 4, // Supported since RocksDB 6.27
 }
 
+/// Used by [`PlainTableFactoryOptions`]
+#[repr(u8)]
+#[derive(Debug, Clone, Copy, PartialEq, Eq, Default)]
```
aleksuss, line 3587: The rust version should be bumped. 1.60.0 doesn't support the `Default` trait for enums.

glittershark: That, or I could hand-impl the trait so we don't need an MSRV bump

src/db_options.rs
```diff
@@ -3578,6 +3582,32 @@ pub enum ChecksumType {
     XXH3 = 4, // Supported since RocksDB 6.27
 }
 
+/// Used by [`PlainTableFactoryOptions`]
+#[repr(u8)]
+#[derive(Debug, Clone, Copy, PartialEq, Eq)]
+pub enum EncodingType {
```
aleksuss, line 3588: I guess the `EncodingType` is a too common name. I would rename it to `KeyEncodingType`.

aleksuss:
Done in #815

PR #780, opened 2023-05-15
Wrap props names into a PropName type offering free conversion to str

mina86:
Introduce properties::PropName and properties::PropertyName which wrap CStr and CString types respectively with additional requirement that the value is valid UTF-8.  This means that the types offer zero-cost conversion into string.

Use it with property names in properties module such that the properties can be used with property_value (which was the case before) and also Displayed (which previously required UTF-8 verification or an unsafe block and now can be done safely with zero additional cost).

This is API breaking change (since type of objects in properties module change) but the breakage should be easy to deal with (especially since the new types implement CStrLike and thus most common usage of the property names will continue to work).

Admittedly this is a substantial amount of code for a relatively insignificant feature.


src/prop_name.rs
```diff
@@ -0,0 +1,318 @@
+use crate::ffi_util::CStrLike;
+
+use std::ffi::{CStr, CString};
+
+/// A borrowed name of a RocksDB property.
+///
+/// The value is guaranteed to be a NUL-terminated UTF-8 string.  This means it
```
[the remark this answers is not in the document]
mina86: I always use NULL to mean pointer and NUL to mean byte but sure I can change it if you prefer.

src/prop_name.rs
```diff
@@ -0,0 +1,318 @@
...
+            "input was not nul-terminated"
+        );
+        // Check all other bytes are non-NUL ASCII bytes.
+        while idx > 0 {
+            idx -= 1;
+            assert!(
+                bytes[idx].is_ascii() && bytes[idx] != 0,
+                "input contained interior nul or non-ASCII byte"
+            );
+        }
```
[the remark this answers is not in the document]
mina86: Unfortunately `iter` isn’t const hence why the somewhat awkward indexing.

src/prop_name.rs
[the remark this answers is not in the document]
mina86: So [`std::ffi::CStr`](https://doc.rust-lang.org/std/ffi/struct.CStr.html) calls them ‘nul bytes’ (lower case, single l) so I ended up going with that.

src/prop_name.rs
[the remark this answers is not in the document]
mina86: Improved by using `split_last` but otherwise manual indexing must stay due to const fn.

src/prop_name.rs
```diff
@@ -0,0 +1,312 @@
...
+    pub(crate) const fn new_unwrap(value: &str) -> &Self {
+        let bytes = if let Some((&0, bytes)) = value.as_bytes().split_last() {
+            bytes
+        } else {
+            panic!("input was not nul-terminated");
+        };
+
+        let mut idx = 0;
+        while idx < bytes.len() {
+            assert!(bytes[idx] != 0, "input contained interior nul byte");
```
aleksuss, line 28:
```suggestion
            assert_ne!(bytes[idx], 0, "input contained interior nul byte");
```

mina86: Alas:

    error: `Arguments::<'a>::new_v1` is not yet stable as a const fn
    error[E0015]: cannot call non-const fn `assert_failed::<u8, u8>` in constant functions

aleksuss: Yeah. I see. See comment below.

src/prop_name.rs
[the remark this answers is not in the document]
aleksuss: Yes. I find the usage. Ok then.

src/prop_name.rs
```diff
@@ -0,0 +1,312 @@
...
+    /// interior nul bytes.
+    pub(crate) const fn new_unwrap(value: &str) -> &Self {
+        let bytes = if let Some((&0, bytes)) = value.as_bytes().split_last() {
+            bytes
+        } else {
+            panic!("input was not nul-terminated");
+        };
+
+        let mut idx = 0;
+        while idx < bytes.len() {
```
aleksuss, line 27: Here should be `bytes.len() - 1`? We don't need to check the last element because it should be `\0`?

src/prop_name.rs
aleksuss: It seems this assert doesn't work as well.

aleksuss: Should be something like this: 
```rust
       let bytes = value.as_bytes();

        if !(bytes.len() > 1 && bytes[bytes.len() - 1] == 0) {
            panic!("input was not nul-terminated");
        }

        let mut idx = 0;
        while idx < bytes.len() - 1 {
            if bytes[idx] == 0 {
                panic!("input contained interior nul byte");
            }

            idx += 1;
        }
```

src/prop_name.rs
mina86: `bytes` comes from `split_last` so it already has the terminating byte excluded.

src/prop_name.rs
mina86: It should work fine.  Not sure when, but `assert` without formatting is const:

    $ c +1.60 t --lib -- prop_name
        Finished test [unoptimized] target(s) in 0.02s
         Running unittests (target/debug/deps/rocksdb-86da67792af28854)

    running 3 tests
    test prop_name::sanity_checks ... ok
    test prop_name::test_non_nul_terminated - should panic ... ok
    test prop_name::test_interior_nul - should panic ... ok

    test result: ok. 3 passed; 0 failed; 0 ignored; 0 measured; 10 filtered out; finished in 0.00s

1.60 is Rust version specified in Cargo.toml

src/prop_name.rs
```diff
@@ -0,0 +1,324 @@
+use crate::ffi_util::CStrLike;
+
+use std::ffi::{CStr, CString};
+
+/// A borrowed name of a RocksDB property.
+///
+/// The value is guaranteed to be a nul-terminated UTF-8 string.  This means it
```
aleksuss, line 7:
```suggestion
/// The value is guaranteed to be a nul-terminated UTF-8 string. This means it
```

src/prop_name.rs
```diff
@@ -0,0 +1,324 @@
...
+
+impl std::borrow::ToOwned for PropName {
+    type Owned = PropertyName;
+
+    #[inline]
+    fn to_owned(&self) -> Self::Owned {
+        PropertyName(self.0.to_owned())
+    }
+
+    // TODO: Uncomment once compiler version is updated to 1.63.
```
aleksuss, line 89: Let's bump rust version to uncomment this.

src/prop_name.rs
```diff
@@ -0,0 +1,324 @@
...
+/// An owned name of a RocksDB property.
+///
+/// The value is guaranteed to be a nul-terminated UTF-8 string.  This means it
+/// can be converted to [`CString`] and [`String`] at zero cost.
+#[derive(PartialEq, Eq, PartialOrd, Ord, Hash)]
+#[repr(transparent)]
+pub struct PropertyName(CString);
+
+impl PropertyName {
+    /// Creates a new object from valid nul-terminated UTF-8 string.  The string
```
aleksuss, line 161:
```suggestion
    /// Creates a new object from valid nul-terminated UTF-8 string. The string
```

src/prop_name.rs
```diff
@@ -0,0 +1,324 @@
...
+
+    #[inline]
+    fn into_c_string(self) -> Result<CString, Self::Error> {
+        Ok(self.0.to_owned())
+    }
+}
+
+/// An owned name of a RocksDB property.
+///
+/// The value is guaranteed to be a nul-terminated UTF-8 string.  This means it
```
aleksuss, line 154:
```suggestion
/// The value is guaranteed to be a nul-terminated UTF-8 string. This means it
```

src/prop_name.rs
```diff
@@ -0,0 +1,324 @@
...
+    }
+}
+
+impl std::ops::Deref for PropertyName {
+    type Target = PropName;
+
+    #[inline]
+    fn deref(&self) -> &Self::Target {
+        // SAFETY: 1. PropName and CStr have the same representation so casting
+        // is safe.  2. self.0 is guaranteed to be valid nul-terminated UTF-8
```
aleksuss, line 192:
```suggestion
        // is safe. 2. self.0 is guaranteed to be valid nul-terminated UTF-8
```

src/prop_name.rs
```diff
@@ -0,0 +1,324 @@
...
+    #[inline]
+    fn into_c_string(self) -> Result<CString, Self::Error> {
+        Ok(self.0.clone())
+    }
+}
+
+/// Constructs a property name for an ‘at level’ property.
+///
+/// `name` is the infix of the property name (e.g. `"num-files-at-level"`) and
+/// `level` is level to get statistics of.  The property name is constructed as
```
aleksuss, line 294:
```suggestion
/// `level` is level to get statistics of. The property name is constructed as
```

src/prop_name.rs
mina86: Done.

mina86:
Clippy failure in unrelated code:

> ```
> error: very complex type used. Consider factoring parts into `type` definitions
>   --> tests/test_comparator.rs:12:17
> 12 |     compare_fn: Box<dyn Fn(&[u8], &[u8]) -> Ordering>,
> ```

aleksuss (approved):

stanislav-tkach (approved):

mina86:
Friendly ping. ^_^

PR #789, opened 2023-06-07
Add support for write buffer manager

sjwiesman:
Adds support for write buffer manager. All the c/ c++ code should be upstreamed. It's here because these functions are missing from [rocksdb/c.h](https://github.com/facebook/rocksdb/blob/main/include/rocksdb/c.h)


sjwiesman:
Sorry. Meant to open against my fork for testing.

Issue #792, opened 2023-06-13
Expose rocksdb_flush_cfs API


lizhanhui:
API `rocksdb_flush_cfs` is very useful when [atomic-flush](https://github.com/facebook/rocksdb/wiki/Atomic-flush) is used as it will flush all column families specified in `cfs` up to the latest sequence number at the time when flush is requested.

PR #793, opened 2023-06-13
Expose flush_cfs_opt to flush multiple column families

lizhanhui:
which is critically important for atomic flush use case

Fix https://github.com/rust-rocksdb/rust-rocksdb/issues/792


stanislav-tkach (commented):
Can you add a simple test for this method?

lizhanhui:
> Can you add a simple test for this method?

Added

stanislav-tkach (approved):

aleksuss (approved):

Issue #799, opened 2023-07-07
Snappy is set as default compression if lib is found, even if feature is not enabled


DaveWK:
I am seeing this behavior in v0.21:

1. Install snappy-devel in os packages
2.  Compile without snappy compression feature enabled, IE:
```
[dependencies.rocksdb]
version = "0.21.0"
default-features = false
features = ["lz4"]
```
3. When instantiating a rocksdb store, it bails out with the error:
`RocksDb(Error { message: \"Not implemented: Unsupported compression method for this build: Snappy\" })'`

According to the docs:
https://docs.rs/rocksdb/latest/rocksdb/struct.Options.html#method.set_compression_type

`Default: DBCompressionType::Snappy (DBCompressionType::None if snappy feature is not enabled).`

This does not occur if the snappy-devel package is not installed. I would not expect the package/lib detection for snappy to take place if the feature was not enabled, or for it to set the default to a feature that is not enabled.

DaveWK:
I put up a PR for where I think the problem is coming from -- https://github.com/rust-rocksdb/rust-rocksdb/pull/800

PR #800, opened 2023-07-07
BugFix: Skip looking for snappy if disabled

DaveWK:
Old logic would link snappy when not configured, but the lib was found via try_to_find_and_link_lib(). Results in contradictory Default=Snappy, Snappy=disabled state.

Ref: https://github.com/rust-rocksdb/rust-rocksdb/issues/799

### DIFF
--- a/librocksdb-sys/build.rs
+++ b/librocksdb-sys/build.rs
@@ -358,10 +358,12 @@ fn main() {
             println!("cargo:rustc-link-lib=dylib=stdc++");
         }
     }
-    if cfg!(feature = "snappy") && !try_to_find_and_link_lib("SNAPPY") {
-        println!("cargo:rerun-if-changed=snappy/");
-        fail_on_empty_directory("snappy");
-        build_snappy();
+    if cfg!(feature = "snappy") {
+        if !try_to_find_and_link_lib("SNAPPY") {
+            println!("cargo:rerun-if-changed=snappy/");
+            fail_on_empty_directory("snappy");
+            build_snappy();
+        }
     }
 
     // Allow dependent crates to locate the sources and output directory of


librocksdb-sys/build.rs
```diff
@@ -358,10 +358,12 @@ fn main() {
             println!("cargo:rustc-link-lib=dylib=stdc++");
         }
     }
-    if cfg!(feature = "snappy") && !try_to_find_and_link_lib("SNAPPY") {
-        println!("cargo:rerun-if-changed=snappy/");
-        fail_on_empty_directory("snappy");
-        build_snappy();
+    if cfg!(feature = "snappy") {
+        if !try_to_find_and_link_lib("SNAPPY") {
```
steviez, line 362: Unless my weekend brain is missing something, your proposed change is equivalent to the existing code no? If the feature is not enabled, the compound statement should [short circuit](https://doc.rust-lang.org/reference/expressions/operator-expr.html#lazy-boolean-operators
)
> the right-hand operand is only evaluated when the left-hand operand does not already determine the result of the expression. That is, || only evaluates its right-hand operand when the left-hand operand evaluates to false, and && only when it evaluates to true.

aleksuss: @steviez It's about assigning evaluation to a boolean variable. In the case of `if` both left-hand and right-hand operands are [checked](https://play.rust-lang.org/?version=stable&mode=debug&edition=2021&gist=e8a77a5b1331f77bce6a7d182453431e).

cvijdea-bd: @aleksuss both operands are checked in the if only because you inverted the conditions: `if check_a() && !check_b()`  stops at `check_a()` if it returns false (`check_b` vs `!check_b()` is actually not relevant). 
Your example had `if !check_a() && !check_b()` which doesn't stop at `!check_a()` because that's true when `check_a()` is false.

In the same vein `if cfg!(feature = "snappy") && !try_to_find_and_link_lib("SNAPPY")` will never call `try_to_find_and_link_lib` when `cfg!(feature = "snappy")` is false - the problem described in the issue has to come from somewhere else, this PR is quite literally a no-op (and the failed Clippy lint even suggests reverting the change).

DaveWK: yeah you're right -- I wasn't sure if the macro expansion was doing something strange to the short circuit, but have locally tested some examples to rule out that being the case. My B -- closing this PR, since it does not fix the issue.

DaveWK:
Does not fix the underlying issue where somehow default = Snappy is being set despite being disabled. Closing

librocksdb-sys/build.rs
aleksuss: @cvijdea-bd Yeah, good catch. I was wondering why it works in that way but now I found out that the example I provided is wrong.

PR #808, opened 2023-08-02
Add `allow_ingest_behind` ffi call for DB Options

siyuan0322:
I found this option for DB is missing whereas the options of ingest exists in https://github.com/siyuan0322/rust-rocksdb/blob/f8906c39ac26d51b4f5873370c8e214ff86fa02c/src/db_options.rs#L3514

But this ingest option could only be enabled when DB is opened with `allow_ingest_behind` set to true.


stanislav-tkach (approved):

siyuan0322:
Hi guys, I fixed the rustfmt, but this error doesn't appear to be introduced by my PR.

```
error: package `linux-raw-sys v0.4.5` cannot be built because it requires rustc 1.63 or newer, while the currently active rustc version is 1.60.0
```

siyuan0322:
Regarding the failure in CI, I have submitted another PR #809 to bump up the version of rustc.

siyuan0322:
> I guess it's ok to bump the rust version in this PR. And bumping up to `1.63.0` is enough.

Bumped up rustc to 1.63.0

aleksuss:
Bump [here](https://github.com/rust-rocksdb/rust-rocksdb/blob/master/Cargo.toml#L6) and [here](https://github.com/rust-rocksdb/rust-rocksdb/blob/master/librocksdb-sys/Cargo.toml#L5) as well, please.

siyuan0322:
> Bump [here](https://github.com/rust-rocksdb/rust-rocksdb/blob/master/README.md?plain=1#L8), [here](https://github.com/rust-rocksdb/rust-rocksdb/blob/master/Cargo.toml#L6) and [here](https://github.com/rust-rocksdb/rust-rocksdb/blob/master/librocksdb-sys/Cargo.toml#L5) as well, please.

Done

siyuan0322:
Well, the CI keeps failing...
This commit fix the clippy, I guess this time it should be green now 😄

siyuan0322:
The test may fail due to no space left on device? 
```
cargo:warning=ar: /home/runner/work/rust-rocksdb/rust-rocksdb/target/tests/trybuild/x86_64-unknown-linux-gnu/debug/build/librocksdb-sys-b0837d8a760c5317/out/librocksdb.a: No space left on device
```

How about remove the target directory after the previous test suite?

aleksuss:
> How about remove the target directory after the previous test suite?

It's a case, but it will increase the time of the CI flow.

PR #810, opened 2023-08-05
feat: Support column_family_metadata, column_family_metadata_cf

ovr:
Hello!

Thanks

### DIFF
--- a/src/db.rs
+++ b/src/db.rs
@@ -2007,7 +2007,8 @@ impl<T: ThreadMode, D: DBInner> DBCommon<T, D> {
         }
     }
 
-    pub fn get_column_family_metadata(&self) -> Result<ColumnFamilyMetaData, Error> {
+    /// Obtains the LSM-tree meta data of the default column family of the DB
+    pub fn get_column_family_metadata(&self) -> ColumnFamilyMetaData {
         unsafe {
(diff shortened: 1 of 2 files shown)


src/db.rs
```diff
@@ -2007,6 +2007,45 @@ impl<T: ThreadMode, D: DBInner> DBCommon<T, D> {
         }
     }
 
+    pub fn get_column_family_metadata(&self) -> Result<ColumnFamilyMetaData, Error> {
```
aleksuss, line 2010: A couple of comments:
1. It would be good to add some description to the public methods.
2. What's the point of the `Result` here?
3. I guess it makes sense to check the pointer `ptr` for null.

aleksuss (changes requested):

src/db.rs
ovr: 1. Done. 
2. I was thinking that it's better for design because C-API can change at any time. Removed.
3. As I can see from the code, it cannot be null.

aleksuss (approved):

aleksuss:
@ovr rebase on the master please.

aleksuss:
@ovr fix the test please.

stanislav-tkach (approved):

Issue #778, opened 2023-05-12
Please add Options::set_wal_compression()


holtgrewe:
There already is a function `rocksdb_options_set_wal_compression` in `librocksdb-sys`. However, this is not exposed in the Rust bindings. Please expose this function.

holtgrewe:
Also, `kCompactionStyleNone` is not exposed in the `pub enum DBCompactionStyle`. That would be nice to have as well.

PR #817, opened 2023-09-14
feat: Expose set_wal_compression_type

ovr:
Hello!

Fix https://github.com/rust-rocksdb/rust-rocksdb/issues/778

Thanks

### DIFF
--- a/tests/test_rocksdb_options.rs
+++ b/tests/test_rocksdb_options.rs
@@ -171,6 +171,18 @@ fn set_compression_options_zstd_max_train_bytes() {
     }
 }
 
+#[test]
+#[cfg(feature = "zstd")]
+fn set_wal_compression_zstd() {
+    let path = DBPath::new("_set_wal_compression_zstd");
+    {
+        let mut opts = Options::default();
+        opts.create_if_missing(true);
+        opts.set_wal_compression_type(DBCompressionType::Zstd);
+        let _db = DB::open(&opts, &path).unwrap();
+    }
+}
+
 fn test_compression_type(ty: DBCompressionType) {
     let path = DBPath::new("_test_compression_type");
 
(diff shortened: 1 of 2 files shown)


aleksuss (approved):

tests/test_rocksdb_options.rs
```diff
@@ -171,6 +171,18 @@ fn set_compression_options_zstd_max_train_bytes() {
     }
 }
 
+#[test]
+#[cfg(feature = "zstd")]
```
stanislav-tkach, line 175: Shouldn't the method itself, as well as other similar methods, be hidden under the `zstd` feature?

stanislav-tkach (approved):

tests/test_rocksdb_options.rs
aleksuss: I guess it doesn't make sense because a compression type can be `None`, I believe. The question is, what would happen if we choose something different rather than `Zstd` or `None`? I guess it would be good to check it in a runtime:
```rust
pub fn set_wal_compression_type(&mut self, t: DBCompressionType) {
    match t {
        DBCompressionType::Zstd | DBCompressionType::None => unsafe {
            ffi::rocksdb_options_set_wal_compression(self.inner, t as c_int);
        },
        _ => unimplemented!(),
    }
}
```

aleksuss:
@ovr, please add a test or extend existing where you pass `DBCompressionType::Lz4` and, namely, unsupported compression type.

ovr:
I have rebased and updated it to reflect review comments.

Thanks

tests/test_rocksdb_options.rs
ovr: My IMHO: We should not throw unimplemented here because it is how RocksDB works. If something is not supported with compression, it will throw a message in the LOG file.

But it's better for DX, in the case when someone writes an unsupported compression type. 

Anyway, I've updated PR with `unimplemented`.

Ty

PR #830, opened 2023-10-28
use write buffer manager introduced in the C API

mohitreddy1996:


mohitreddy1996:
Apologies for this PR. Was supposed to be in a forked repo. Will eventually create a PR here :)

PR #837, opened 2023-11-25
Expose LRU cache options

athre0z:
This PR exposes the following two DB options related to BlobDB caching:

https://github.com/facebook/rocksdb/blob/f6fd4b9dbd15dba36f7e5ad23de407b5c26b1460/include/rocksdb/advanced_options.h#L1129-L1148


athre0z:
Added one more commit to also expose the advanced LRU cache constructor with custom options.

src/db_options.rs
```diff
@@ -57,6 +57,13 @@ impl Cache {
         Cache(Arc::new(CacheWrapper { inner }))
     }
 
+    /// Creates an LRU cache with custom options.
+    pub fn new_lru_cache_opts(opts: &LruCacheOptions) -> Cache {
+        let inner =
+            NonNull::new(unsafe { ffi::rocksdb_cache_create_lru_opts(opts.inner) }).unwrap();
```
stanislav-tkach, line 63: Am I missing something or is there an issue in the code below?
```rust
let options = LruCacheOptions::default();
let cache = Cache::new_lru_cache_opts(&options);
drop(options);
// Now the cache holds a pointer to the freed memory.
```

athre0z: The pointer to the options is not stored in the resulting cache object; it is only borrowed for the duration of construction. See the following call chain:

https://github.com/facebook/rocksdb/blob/f6fd4b9dbd15dba36f7e5ad23de407b5c26b1460/db/c.cc#L4825

https://github.com/facebook/rocksdb/blob/f6fd4b9dbd15dba36f7e5ad23de407b5c26b1460/include/rocksdb/cache.h#L286

https://github.com/facebook/rocksdb/blob/f6fd4b9dbd15dba36f7e5ad23de407b5c26b1460/cache/lru_cache.cc#L689

https://github.com/facebook/rocksdb/blob/f6fd4b9dbd15dba36f7e5ad23de407b5c26b1460/cache/lru_cache.cc#L651C43-L651C47

stanislav-tkach: Thanks for the clarification!

stanislav-tkach (approved):

zaidoon1:
This also LGTM, adding tests would be great

athre0z:
I can write some tests, but it might take a while until I get to it depending on how deep these tests should go. Simply adding tests that invoke the new methods to make sure it links & doesn't segfault is quickly done, but I might have to read up a bit deeper if you'd like tests that actually make sure that RocksDB is caching things as expected etc (like some other test cases seem to do). What's the expectation here?

zaidoon1:
> I can write some tests, but it might take a while until I get to it depending on how deep these tests should go: I'm unfortunately rather short on time right now. Simply adding tests that invoke the new methods to make sure it links & doesn't segfault is quickly done, but I might have to read up a bit deeper if you'd like tests that actually make sure that RocksDB is caching things as expected etc (like some other test cases seem to do). What's the expectation here?

a simple test case where you set the new options being exposed here, and put some kvs without segfaulting is good enough

zaidoon1:
LGTM, will merge once builds pass

zaidoon1 (approved):

PR #841, opened 2023-12-02
feat: Expose wait_for_compact

zaidoon1:


stanislav-tkach (approved):

src/db_options.rs
```diff
@@ -4009,6 +4009,78 @@ impl CompactOptions {
...
+    }
+
+    /// If true, close db after waiting is done. By the time Close() is
+    /// called here, there should be no background jobs in progress and no new
+    /// background jobs should be added. DB may not have been closed if Close()
+    /// returned Aborted status due to unreleased snapshots in the system. See
+    /// comments in DB::Close() for details.
+    ///
+    /// Default: false
+    pub fn set_close_db(&mut self, v: bool) {
```
cvijdea-bd, line 4066: I think method shouldn't be exposed, or alternatively should be marked unsafe? It allows the underlying database to be closed without dropping the DB object, and RocksDB is pretty light on guarantees about continuing to use a DB after it was closed.

https://github.com/facebook/rocksdb/blob/c96d9a0fbbfd2215e7ae88af8a13b94109a349b5/include/rocksdb/db.h#L345

> Regardless of the return status, the DB must be freed.

From the wiki:

> Q: Is it safe to close RocksDB while another thread is issuing read, write or manual compaction requests?
> 
> A: No. The users of RocksDB need to make sure all functions have finished before they close RocksDB. You can speed up the waiting by calling CancelAllBackgroundWork().

Looking at the implementation, there *is* a `closed_` flag in RocksDB but it is only checked in Close and in the destructor

https://github.com/facebook/rocksdb/blob/c96d9a0fbbfd2215e7ae88af8a13b94109a349b5/db/db_impl/db_impl.cc#L5136

cvijdea-bd: Another safe alternative would be remove this and add a `pub fn wait_for_compact_and_close(self, opts: &WaitForCompactOptions) -> Result<(), Error>` method that consumes the DB instance

zaidoon1: hmmm, ok for the time being, I'm going to remove set_close_db as I don't need it. We can revist if someone needs this option

aleksuss (approved):

pragmaxim:
Is it possible to access the `wait_for_compact` on `TransactionDB` instance?

PR #857, opened 2024-01-29
only enable jemalloc on non-windows targets

tzemanovic:


librocksdb-sys/build.rs
```diff
@@ -206,14 +206,14 @@ fn build_rocksdb() {
             "port/win/win_logger.cc",
         ]);
 
-        if cfg!(feature = "jemalloc") {
+        if cfg!(all(feature = "jemalloc", not(windows))) {
             lib_sources.push("port/win/win_jemalloc.cc");
```
aleksuss, line 210: I don't get it. Why do we need to add a source file related to the windows on `not(windows)` target?

tzemanovic: yeah good point, this is pointless

tzemanovic: I just removed this port [c195c31](https://github.com/rust-rocksdb/rust-rocksdb/pull/857/commits/c195c315281f31fe06108cde464d3de0c2620de5)

PR #865, opened 2024-03-08
Feat: Adds /MT for static crt on msvc targets

spector-9:
# Issue
Sometime other libraries in rust might be compiled on MSVC target with different CRT flag. There are two option for the flag, /MT and /MD. If one library uses /MD and other uses /MT then there will be issue in compilation of the project that have both those libraries as dependencies.

# Solution
The fix is pretty simple. I just added `crt_static = true` in the `build.rs` of librocksdb-sys which is behind a feature flag `mt_static`.


zaidoon1 (approved):

stanislav-tkach:
Perhaps it would make sense to describe this option either in the `Cargo.toml` file or in the readme?..

aleksuss (approved):

PR #867, opened 2024-03-14
Add portable feature for RocksDB build

### LINKED ISSUE #527: Important to have a portable feature
I came from Aleo's snarkOS, where we know two users have had issues with "illegal instructions". A detailed, package-by-package walkthrough shows that `rust-rocksdb` uses BMI2/AVX2 instruction, which will fail in the older machines. Since the release of snarkOS is supposed to run on many users' machines, it has become a necessity to have portability.

It is good for `rust-rocksdb` to have a portable feature, which introduces the `PORTABLE` feature of facebook/rocksdb, so that those instructions are not being used.

And this is also an inherently necessary practice of Rust programming --- note that by default, Rust's release mode only uses minimal instruction sets, and intentionally avoids the newer instruction sets, in an effort to maintain portability.

This is related to https://github.com/rust-rocksdb/rust-rocksdb/issues/363. Parity Tech wants to fix this issue, and in the end, they have to fork the rust-rocksdb layer by layer: https://crates.io/crates/parity-rocksdb-sys

sujayakar:
Add a feature for passing `PORTABLE=1` to the RocksDB build, letting downstream crates opt into a (potentially slower) portable build.

Fixes #527.


zaidoon1 (approved):

zaidoon1:
probably also good to update README about the existence of this feature/option?

aleksuss (approved):

sujayakar:
thanks for merging @zaidoon1 ! would you be able to a crates.io release too?

aleksuss:
@sujayakar you can use dependencies directly from the github in your Cargo.toml. 
@zaidoon1 I guess, release a new version after every commit it's overkill.

axnsan12:
> config.define("PORTABLE", Some("1"));

This doesn't do anything. There is no code in rocksdb that uses a preprocessor define named PORTABLE https://github.com/search?q=repo%3Afacebook%2Frocksdb+PORTABLE&type=code&p=2

axnsan12:
`PORTABLE` appears only as a [Makefile build script](https://github.com/facebook/rocksdb/blob/6ddfa5f06140c8d0726b561e16dc6894138bcfa0/build_tools/build_detect_platform#L631) / [CMake](https://github.com/facebook/rocksdb/blob/6ddfa5f06140c8d0726b561e16dc6894138bcfa0/CMakeLists.txt#L275) parameter, neither of which is used by librocksdb-sys in `build.rs` - it just calls the C++ compiler directly using the `cc` crate.

Also, in fact setting PORTABLE to 1 in cmake also has the effect of explicitly not doing anything - they default to passing `-march=native` which is disabled by `PORTABLE=1`.

For librocksdb-sys, there is no `-march` option passed so the portability will depend on the defaults for the C+ compiler specified by the build environment. There are also architecture feature flags that are enabled according to cargo features: https://github.com/rust-rocksdb/rust-rocksdb/blob/0d8804a5d668989fc450205e990204bf72427650/librocksdb-sys/build.rs#L108

sujayakar:
ah, thanks for the explanation @axnsan12 and for reverting @zaidoon1! I should have really tested with an x64 machine and seen what codegen was emitting.

I'll double check what our system compiler is doing in our setup.

PR #754, opened 2023-03-03
Fix 32-bit ARM build

EyeOfPython:
Building this crate on `arm-linux-gnueabihf` is fine, HOWEVER, opening a rocksdb database using the built code fails:

On our CI, we get a "Value too large for defined data type" in our functional tests, see [build output](https://build.bitcoinabc.org/repository/download/BitcoinABC_BitcoinAbcStaging/515561:id/artifacts.tar.gz!/functional/test_runner_%E2%82%BF%E2%82%B5_%F0%9F%8F%83_20230302_193244/chronik_block_214/node0/regtest/debug.log) (you can login as guest if necessary).

After a bit of digging, it turns out this refers to a [missing environment variable specifically for 32-bit systems that want to open files using 64-bit offsets](https://stackoverflow.com/a/23372153).

Therefore, this PR [fixes our build](https://build.bitcoinabc.org/viewLog.html?buildId=515991&buildTypeId=BitcoinABC_BitcoinAbcStaging&tab=buildResultsDiv&branch_BitcoinAbcDiffs=refs%2Ftags%2Fphabricator%2Fdiff%2F38261) by adding those two env variables when CARGO_CFG_TARGET_POINTER_WIDTH is not 64.

PR #869, opened 2024-03-20
fix: android build in 32-bit devices

LucasXu0:
The `_FILE_OFFSET_BITS` macro was defined with a default value of 64, which broke the Android 32-bit target build. The error is shown below.

```
  cargo:warning=/Users/lucas.xu/Library/Android/sdk/ndk/26.2.11394342/toolchains/llvm/prebuilt/darwin-x86_64/bin/../sysroot/usr/include/c++/v1/fstream:973:20: error: use of undeclared identifier 'ftello'; did you mean 'ftell'?
  cargo:warning=    pos_type __r = ftello(__file_);
  cargo:warning=                   ^
  cargo:warning=/Users/lucas.xu/Library/Android/sdk/ndk/26.2.11394342/toolchains/llvm/prebuilt/darwin-x86_64/bin/../sysroot/usr/include/c++/v1/cstdio:151:9: note: 'ftell' declared here
  cargo:warning=using ::ftell _LIBCPP_USING_IF_EXISTS;
```


LucasXu0:
Screenshot 2024-03-20 at 10 47 20

LucasXu0:
@zaidoon1 Could you take a look at this PR, and please let me know if there is any change I can do.

librocksdb-sys/build.rs
```diff
@@ -224,11 +230,6 @@ fn build_rocksdb() {
         config.define("ROCKSDB_IOURING_PRESENT", Some("1"));
     }
 
-    if env::var("CARGO_CFG_TARGET_POINTER_WIDTH").unwrap() != "64" {
```
zaidoon1, line 227: so removing this changes the behaviour for non android targets. I can't comment on the change for the android targets, but if that's what is failing, let's continue to have this apply for non andorid builds

librocksdb-sys/build.rs
```diff
@@ -154,6 +154,12 @@ fn build_rocksdb() {
         config.define("OS_ANDROID", None);
         config.define("ROCKSDB_PLATFORM_POSIX", None);
         config.define("ROCKSDB_LIB_IO_POSIX", None);
+
+        if &target == "armv7-linux-androideabi" {
+            config.define("_FILE_OFFSET_BITS", Some("32"));
+        } else if &target == "aarch64-linux-android" {
+            config.define("_FILE_OFFSET_BITS", Some("64"));
```
zaidoon1, line 161: why do we not set `config.define("_LARGEFILE64_SOURCE", Some("1"));` here?

librocksdb-sys/build.rs
```diff
@@ -154,6 +154,12 @@ fn build_rocksdb() {
         config.define("OS_ANDROID", None);
         config.define("ROCKSDB_PLATFORM_POSIX", None);
         config.define("ROCKSDB_LIB_IO_POSIX", None);
+
+        if &target == "armv7-linux-androideabi" {
+            config.define("_FILE_OFFSET_BITS", Some("32"));
```
zaidoon1, line 159: it looks like the real fix is here? so can we keep this if, remove the else if `aarch64-linux-android` and on line 227 do:

```
if &target != "armv7-xxxx' && env::var("CARGO_CFG_TARGET_POINTER_WIDTH").unwrap() != "64" {
}
```

LucasXu0: Yep. We can only keep the change for the "armv7-linux-androideabi" target.

librocksdb-sys/build.rs
LucasXu0: I'm not sure if the code here is correct. Because if "CARGO_CFG_TARGET_POINTER_WIDTH" is NOT equal to 64, then it will define "_FILE_OFFSET_BITS" to 64.

I can keep these lines and filter the Android target here if you think it's necessary.

LucasXu0:
```
    if &target != "armv7-linux-androideabi"
        && env::var("CARGO_CFG_TARGET_POINTER_WIDTH").unwrap() != "64"
    {
        config.define("_FILE_OFFSET_BITS", Some("64"));
        config.define("_LARGEFILE64_SOURCE", Some("1"));
    }
```

zaidoon1: > I'm not sure if the code here is correct. Because if "CARGO_CFG_TARGET_POINTER_WIDTH" is NOT equal to 64, then it will define "_FILE_OFFSET_BITS" to 64.
> 
> I can keep these lines and filter the Android target here if you think it's necessary.

I looked at https://github.com/rust-rocksdb/rust-rocksdb/pull/754 which added this (and it also apparently broke other builds but that was not addressed?). It's not clear to me why after detecting the pointer width is not 64, we still force it anyway, is this for cross compilation? I'll defer to @aleksuss who may have more context on this

zaidoon1 (approved):

zaidoon1:
@LucasXu0 can you squash your commits to just one commit and I can merge this,

LucasXu0:
@zaidoon1 Yes. Or can you use 'squash merge' in GitHub? Let me know if you prefer me to squash them into one commit.

Issue #806, opened 2023-07-25
Add Send/Sync to BackupEngine


Dr-Emann:
I would like to store an `Arc<Mutex<BackupEngine>>` to share a backup engine across threads. As far as I can tell from the docs and implementation, I'm pretty sure Send would be safe to implement, and I _think_ Sync is as well.

PR #873, opened 2024-03-23
Make BackupEngine Send

widagdos:
BackupEngine is a simple pointer wrapper, so it's safe to send to another thread since the underlying RocksDB backup engine is thread-safe.


widagdos:
#806 @Dr-Emann

stanislav-tkach (approved):

zaidoon1 (approved):

PR #882, opened 2024-04-30
!fix: Ensure default column family uses specified open options

0xdeafbeef:
Previously, the default column family was initialized with default settings if not explicitly included in the column family descriptors. This commit ensures that the default column family inherits the open options

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,8 @@
 # Changelog
 
 ## [Unreleased]
+* Document that `default` column family doesn't inherit open options of db (
+  0xdeafbeef)
 
 ## 0.22.0 (2024-02-13)
 
--- a/src/db.rs
+++ b/src/db.rs
@@ -408,6 +408,8 @@ impl<T: ThreadMode> DBWithThreadMode<T> {
 
     /// Opens a database with the given database with a Time to Live compaction filter and
     /// column family descriptors.
+    /// *NOTE*: `default` column family is opened with `Options::default()`.
+    /// If you want to open `default` cf with different options, set them explicitly in `cfs`.
     pub fn open_cf_descriptors_with_ttl<P, I>(
         opts: &Options,
         path: P,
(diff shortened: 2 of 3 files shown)


stanislav-tkach (approved):

zaidoon1:
I'm hesitant to make this change. The existing api clearly states that the cfs provided will get the provided options applied so if the user doesn't provide the default cf then the default cf will get the default settings which makes sense to me. The issue here is that this change is a breaking change that will change behaviour for users that have not set anything for the default cf which may break people... It might be better to document the functions to state that the default cf will get the default options if it's not included/not specifically configured.

stanislav-tkach:
I also thought that it is breaking change and with current `rust-rocksdb` versioning it is impossible to indicate that to the users. I tend to agree with @zaidoon1.

0xdeafbeef:
@stanislav-tkach @zaidoon1 
```rs
    #[test]
    fn test_default_cf_compaction() {
         use rocksdb::{CompactionDecision, Options, DB};
         
        let mut cf_opts = Options::default();
        cf_opts.set_compaction_filter("test", |_, _, _| {
            println!("compaction_filter in cf");
            CompactionDecision::Keep
        });

        let descriptor = rocksdb::ColumnFamilyDescriptor::new("test", cf_opts);

        opts.create_if_missing(true);
        opts.create_missing_column_families(true);
        opts.set_compaction_filter("test", |_, _, _| {
            println!("compaction_filter in DEFAULT");
            CompactionDecision::Keep
        });
        let db = DB::open_cf_descriptors(&opts, "test", vec![descriptor]).unwrap();
        let cf = db.cf_handle("test").unwrap();
        db.put_cf(cf, b"1", b"").unwrap();
        db.compact_range_cf(cf, None::<&[u8]>, None::<&[u8]>);

        db.put(b"key", "").unwrap();
        db.compact_range(None::<&[u8]>, None::<&[u8]>);
        drop(db);

        DB::destroy(&opts, "test").unwrap();
    }
 ```
 
 In this case, `compaction_filter in cf` will be printed and `compaction_filter in DEFAULT` won't.
    
    `open_cf_descriptors` docs says:
   > Opens a database with the given database options and column family descriptors.  
   
   Thus, when you call put on the database, you expect that the options set for the database will be applied.
   
   So if a change breaks backward compatibility, I think we should at least update the docs to make this behavior more clear.

zaidoon1:
> I think we should at least update the docs to make this behavior more clear.

I agree, if you can update the pr, that would be great

0xdeafbeef:
> > I think we should at least update the docs to make this behavior more clear.
> 
> I agree, if you can update the pr, that would be great

done

zaidoon1 (approved):

PR #354, opened 2019-11-21
Rework `build.rs` and support linking to system RocksDB

lovesegfault:
This is still very WIP, but I figured I'd open a PR to track progress.

TODO:

- [ ] Split vendored and system out of `main.rs`
- [ ] Figure out how to correctly link system libraries _without_ envvars
- [ ] Test that vendored deps work correctly
- [ ] Test that system deps work correctly
- [ ] Test NixOS, Windows, MacOS (the weird ones)
- [ ] Clippy pass
- [ ] Rustfmt pass

Fixes #310

PR #885, opened 2024-05-13
build: 💐 incremental builds are idempotent

cratelyn:
this branch makes a collection of small tweaks to the `librocksdb-sys` crate's
build script. these changes improve incremental builds of the library, by
adding conditional early returns that check if the rocksdb bindings, rocksdb
library, and snappy library have already been generated or compiled.

currently, repeated builds of the `librocksdb-sys` library are never considered
fresh. for projects that rely on rocksdb as backing storage, this can create
a bottleneck in builds, and lead to spuriously recompiling dependent libraries.

### DIFF
--- a/librocksdb-sys/build.rs
+++ b/librocksdb-sys/build.rs
@@ -53,6 +53,19 @@ fn bindgen_rocksdb() {
 }
 
 fn build_rocksdb() {
+    /// The name of the compiled library.
+    const ARCHIVE: &str = "librocksdb.a";
+
+    // Check if rocksdb has already been built.
+    if env::var("OUT_DIR")
+        .map(PathBuf::from)
+        .unwrap()
+        .join(ARCHIVE)
+        .exists()
+    {
+        return;
+    }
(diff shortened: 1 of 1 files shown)


cratelyn:
i pushed an (_identical_) ammended commit, hoping to try re-running CI. i don't see anything in particular about these changes that would lead to the linker errors [observed](https://github.com/rust-rocksdb/rust-rocksdb/actions/runs/9071307884) in CI.

cratelyn:
the issues appear to be related to the `trybuild` snapshot tests. there's an unfortunate interaction between the build script and cargo's test runner, i'm debugging further over in cratelyn/rust-rocksdb#2.

jhult:
@cratelyn, any news?

cratelyn:
i'm going to close this. #354 is a pre-existing rework of the build logic that this would conflict with, and for my personal use-case, linking against a pre-compiled version of rocksdb improved build times satisfactorily.

PR #890, opened 2024-06-01
add option set_avoid_unnecessary_blocking_io

w41ter:


zaidoon1:
would be good to add a basic test to make sure setting this option doesn't cause any panics similar to https://github.com/rust-rocksdb/rust-rocksdb/pull/845

w41ter:
Ok, I'll add the test later.

zaidoon1 (approved):

PR #891, opened 2024-06-01
add option to enable auto tuned ratelimiter

w41ter:


zaidoon1:
similar to your other pr, a basic test to set this option and making sure we don't panic would be great.

zaidoon1 (approved):

src/db_options.rs
```diff
@@ -3105,9 +3105,30 @@ impl Options {
...
+        fairness: i32,
+    ) {
+        unsafe {
+            let ratelimiter = ffi::rocksdb_ratelimiter_create_auto_tuned(
+                rate_bytes_per_sec,
+                refill_period_us,
+                fairness,
+            );
+            ffi::rocksdb_options_set_ratelimiter(self.inner, ratelimiter);
+            ffi::rocksdb_ratelimiter_destroy(ratelimiter);
```
aleksuss, line 3131: Do we really need to destroy the `ratedelimiter` explicitly here?

w41ter: If it is not called, there will be a pointer-sized memory leak. Of course, an app is unlikely to set options frequently, so it will not have much impact on user programs. However, if the user uses tools such as ASAN to check the program, it will have an effect.

zaidoon1: SGTM, @aleksuss is that good with you?

aleksuss: Yeah. It's fine. @w41ter, thanks for the explanation.

aleksuss: I missed C++ so much 😄 I decided to check it once again. As I understand, the main point of why we should destroy the `ratelimiter` there is that we increment the counter of the `shared_ptr` [here](https://github.com/facebook/rocksdb/blob/main/db/c.cc#L4051) and after that the `options` starts owning it [here](https://github.com/facebook/rocksdb/blob/main/include/rocksdb/options.h#L628). After that the shared_ptr's counter value becomes 2. By invoking the `rocksdb_ratelimiter_destroy` we decrease shared_ptr's counter to 1 and after destroying the `Options` we release memory of the `ratelimeter` as well.

PR #894, opened 2024-06-04
 [WIP] Add Encryption at Rest support using OpenSSl #893

petrpan26:
Implement BlockCipher using OpenSSL API and expose c bindings to the rust code to allow data encryption


aleksuss:
Hello @petrpan26. Thanks for your contribution. However, I do not prefer using custom headers and C++ implementation in the bindings. I guess the better way is to add the logic you want to add to the [`facebook/rocksdb`](https://github.com/facebook/rocksdb) first and then add it to this repository.

petrpan26:
@aleksuss Thank you for letting me know. I thought that was the case but wanted to try because i feel like rust is lacking a lot of feature from rocksdb which is not ideal. I might move forward with a fork of our own from rust-rocksdb for now. Closing this for now

PR #898, opened 2024-07-03
Add support for enabling blob cache

exabytes18:
This PR adds support for enabling blob caching.

### DIFF
--- a/src/db_options.rs
+++ b/src/db_options.rs
@@ -210,6 +210,7 @@ impl Cache {
 pub(crate) struct OptionsMustOutliveDB {
     env: Option<Env>,
     row_cache: Option<Cache>,
+    blob_cache: Option<Cache>,
     block_based: Option<BlockBasedOptionsMustOutliveDB>,
     write_buffer_manager: Option<WriteBufferManager>,
 }
@@ -219,6 +220,7 @@ impl OptionsMustOutliveDB {
         Self {
             env: self.env.clone(),
             row_cache: self.row_cache.clone(),
+            blob_cache: self.blob_cache.clone(),
             block_based: self
                 .block_based
                 .as_ref()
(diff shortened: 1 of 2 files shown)


src/db_options.rs
```diff
@@ -3329,6 +3331,14 @@ impl Options {
         }
     }
 
+    /// Sets the blob cache.
```
zaidoon1, line 3334: would be good to copy the information here https://github.com/facebook/rocksdb/blob/v9.3.1/include/rocksdb/advanced_options.h#L979-L987 to give more context on what this is.

zaidoon1:
Based on https://github.com/facebook/rocksdb/wiki/BlobDB#column-family-options it looks like there is multiple options that we should expose as well esepcially:
```
enable_blob_files: set it to true to enable key-value separation.
```

My guess is this pr as is won't actually do anything/enable blobdb which is not what you want?

exabytes18:
> Based on https://github.com/facebook/rocksdb/wiki/BlobDB#column-family-options it looks like there is multiple options that we should expose as well esepcially:
> 
> ```
> enable_blob_files: set it to true to enable key-value separation.
> ```
> 
> My guess is this pr as is won't actually do anything/enable blobdb which is not what you want?

Hi @zaidoon1. This PR is for enabling blobdb caching, not enabling blobdb itself (the `set_enable_blob_files()` option is already available in rust-rocksdb; please see the included test for how to enable blobdb + blob caching).

src/db_options.rs
exabytes18: Ok, I've copied the text from rocksdb.

zaidoon1:
> > Based on https://github.com/facebook/rocksdb/wiki/BlobDB#column-family-options it looks like there is multiple options that we should expose as well esepcially:
> > ```
> > enable_blob_files: set it to true to enable key-value separation.
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > My guess is this pr as is won't actually do anything/enable blobdb which is not what you want?
> 
> Hi @zaidoon1. This PR is for enabling blobdb caching, not enabling blobdb itself (the `set_enable_blob_files()` option is already available in rust-rocksdb; please see the included test for how to enable blobdb + blob caching).

oh i see, this is for exposing more options not adding support for blobdb.

zaidoon1 (approved):

PR #901, opened 2024-07-11
support user defined timestamp in rust bindings

siyuan0322:
This PR introduce support for operations with timestamp, including
- put / get / delete with timestamp in DB operations
- put / delete with timestamp in sst_file_writer
- a predefined comparator with timestamp implementation
- other relevant options with timestamp in ReadOptions and CompactOptions


siyuan0322:
Could someone give some review and comments? Thanks in advance :)

aleksuss (commented):
I'm not sure that we really need the publicity of the functions/methods everywhere. Does it make sense to change `pub` to `pub(crate)` or even private where it possible?

src/comparator.rs
```diff
@@ -13,13 +13,38 @@
...
+    &user_key[..user_key.len() - ts_sz]
+}
+
+pub fn extract_timestamp_from_user_key(user_key: &[u8], ts_sz: usize) -> &[u8] {
+    &user_key[user_key.len() - ts_sz..]
+}
+
+#[inline]
+pub fn decode_timestamp(ptr: &[u8]) -> u64 {
+    u64::from_be_bytes(ptr[..8].try_into().unwrap())
```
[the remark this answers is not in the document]
siyuan0322: Made it private and add comments for the reason choosing u64 as timestamp.

src/db.rs
```diff
@@ -1544,6 +1544,68 @@ impl<T: ThreadMode, D: DBInner> DBCommon<T, D> {
         }
     }
 
+    pub fn put_with_ts_opt<K, V, S>(
```
aleksuss, line 1547: Could you please add comment-docs to all public methods?

tests/test_comparator.rs
```diff
@@ -39,19 +39,15 @@ pub fn write_to_db_with_comparator(compare_fn: Box<CompareFn>) -> Vec<String> {
...
+/// This should verify backwards compatibility
 /// Then run a test with a clojure where an x-variable is passed
 /// Keep in mind that this variable must be moved to the clojure
 /// Then run a test with a reverse sorting clojure and make sure the order is reverted
 fn test_comparator() {
     let local_compare = move |one: &[u8], two: &[u8]| one.cmp(two);
 
-    let x = 0;
     let local_compare_reverse = move |one: &[u8], two: &[u8]| {
-        println!(
```
siyuan0322, line 51: These changes of `local_compare_reverse` is not intended. Will be reverted in next commit.

src/db.rs
siyuan0322: done

src/db.rs
```diff
@@ -1544,6 +1544,76 @@ impl<T: ThreadMode, D: DBInner> DBCommon<T, D> {
...
+                value.as_ptr() as *const c_char,
+                value.len() as size_t,
+            ));
+            Ok(())
+        }
+    }
+
+    /// Put with timestamp in a specific column family with WriteOptions.
+    /// If "key" already exists, it will be overwritten.
+    /// Takes an additional argument `ts` as the timestamp.
```
zaidoon1, line 1583: what happens if we have key_1, ts_1 and we try to write key_1, ts_2. Does ts_2 replace ts_1 or do we have two kvs now?

src/sst_file_writer.rs
```diff
@@ -122,6 +121,31 @@ impl<'a> SstFileWriter<'a> {
         }
     }
 
+    /// Adds a Put key with value to currently opened file
```
zaidoon1, line 124: the ` to currently opened file ` is confusing me and should be removed? This just stores the key/value/ts in the db right?

src/db.rs
```diff
@@ -1675,6 +1846,29 @@ impl<T: ThreadMode, D: DBInner> DBCommon<T, D> {
...
+        key: K,
+        ts: S,
+    ) -> Result<(), Error> {
+        self.delete_with_ts_opt(key.as_ref(), ts.as_ref(), &WriteOptions::default())
+    }
+
+    /// Delete with timestamp in a specific column family.
+    /// Takes an additional argument `ts` as the timestamp.
+    /// Note: the DB must be opened with user defined timestamp enabled.
+    pub fn delete_cf_with_ts<K: AsRef<[u8]>, S: AsRef<[u8]>>(
```
zaidoon1, line 1863: can you write a test that uses this function

src/db.rs
```diff
@@ -1647,6 +1775,49 @@ impl<T: ThreadMode, D: DBInner> DBCommon<T, D> {
...
+            value.as_ref(),
+            &WriteOptions::default(),
+        )
+    }
+
+    /// Put with timestamp in a specific column family.
+    /// If "key" already exists, it will be overwritten.
+    /// Takes an additional argument `ts` as the timestamp.
+    /// Note: the DB must be opened with user defined timestamp enabled.
+    pub fn put_cf_with_ts<K, V, S>(
```
zaidoon1, line 1800: can you write a test that uses this function

zaidoon1:
aside from adding a few more tests. this LGTM

src/db.rs
siyuan0322: It should be coexists. Thanks, I'll add more clarification.

zaidoon1 (approved):

aleksuss (approved):
@siyuan0322, thanks for the contribution. Good job! 👍🏻

PR #921, opened 2024-08-24
Expose WBWI through rust-rocksdb

### LINKED ISSUE #836: Is WriteBatchWithIndex supported?
I ever wrote a Java program with rocksdbjni -

```java
import io.korandoru.zeronos.server.record.BackendRangeResult;
import java.io.IOException;
import java.io.UncheckedIOException;
import java.util.ArrayList;
import java.util.Arrays;
import java.util.List;
import java.util.function.Predicate;
import lombok.extern.slf4j.Slf4j;
import org.rocksdb.RocksDB;
import org.rocksdb.RocksDBException;
import org.rocksdb.RocksIterator;
import org.rocksdb.WriteBatchWithIndex;
import org.rocksdb.WriteOptions;

@Slf4j
public class RocksDBRWBatch implements WriteTxn {

    private final WriteBatchWithIndex batch;
    private final RocksDB db;

    public RocksDBRWBatch(RocksDB db) {
        this.db = db;
        this.batch = new WriteBatchWithIndex();
    }

    @Override
    public BackendRangeResult unsafeRange(Namespace ns, byte[] key, byte[] end, long limit) {
        final long bound;
        final Predicate<byte[]> isMatch;
        if (end == null) {
            bound = 1;
            isMatch = bytes -> Arrays.equals(bytes, key);
        } else {
            bound = limit > 0 ? limit : Long.MAX_VALUE;
            isMatch = bytes -> Arrays.compare(bytes, end) < 0;
        }

        final List<byte[]> keys = new ArrayList<>();
        final List<byte[]> values = new ArrayList<>();
        final RocksIterator it = batch.newIteratorWithBase(db.newIterator());
        final byte[] fixedKey = ns.fixKey(key);
        for (it.seek(fixedKey); it.isValid(); it.next()) {
            final byte[] bytes = ns.unfixKey(it.key());
            if (!isMatch.test(bytes)) {
                break;
            }
            keys.add(bytes);
            values.add(it.value());
            if (keys.size() >= bound) {
                break;
            }
        }
        return new BackendRangeResult(keys, values);
    }

    @Override
    public void unsafePut(Namespace ns, byte[] key, byte[] value) {
        final byte[] bytes = ns.fixKey(key);
        try {
            batch.put(bytes, value);
        } catch (RocksDBException e) {
            log.atError().addKeyValue("namespace", ns).log("failed to write to the backend", e);
            throw new UncheckedIOException(new IOException(e));
        }
    }

    @Override
    public void close() throws Exception {
        try (final WriteOptions options = new WriteOptions()) {
            db.write(options, batch);
        } finally {
            batch.close();
        }
    }
}

```

New I found that to port it to Rust, the concept `WriteBatchWithIndex` mismatches. I wonder if I miss something in this crate or just this crate is yet to support this.

tillrohrmann:
Expose WBWI through rust-rocksdb. This PR depends on https://github.com/facebook/rocksdb/pull/12968 to work.

This fixes #836.

cc @zaidoon1

### DIFF
--- a/src/db.rs
+++ b/src/db.rs
@@ -23,7 +23,8 @@ use crate::{
     ColumnFamily, ColumnFamilyDescriptor, CompactOptions, DBIteratorWithThreadMode,
     DBPinnableSlice, DBRawIteratorWithThreadMode, DBWALIterator, Direction, Error, FlushOptions,
     IngestExternalFileOptions, IteratorMode, Options, ReadOptions, SnapshotWithThreadMode,
-    WaitForCompactOptions, WriteBatch, WriteOptions, DEFAULT_COLUMN_FAMILY_NAME,
+    WaitForCompactOptions, WriteBatch, WriteBatchWithIndex, WriteOptions,
+    DEFAULT_COLUMN_FAMILY_NAME,
 };
 
 use crate::column_family::ColumnFamilyTtl;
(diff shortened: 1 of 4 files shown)


alrevuelta:
Really interested in this feature, mainly to allow read-your-own-write in batches. The PR seems merged after a long time. So i guess should end up soon in master? thanks!

AhmedSoliman:
We have been using the code in this PR already in restatedev/restate in production for quite sometime now, it'd be nice if we can get some eyes on it and get have it merged upstream.

zaidoon1:
it would be good to add tests to make sure it does what it's supposed to do. Aside from that, we just wait for the next official rocksdb release that includes the commit needed and then it should be good

hackermondev:
any updates on this?

PR #924, opened 2024-09-04
Improve statistics by auto gen enum Ticker & enum Histogram

rockeet:
1. `librocksdb-sys/Makefile`: Add gen_statistics and `gen_statistics.bash`,
    `make -C librocksdb-sys gen_statistics` will generate 2 files:
     * src/statistics_enum_ticker.rs
     * src/statistics_enum_histogram.rs

2. In statistics.rs, we `include!` these 2 files and remove old hand writing
    `enum Tickers` & `enum Histograms`.

3. File `gen_statistics.bash` is used for generating these 2 files which is called in
   `librocksdb-sys/Makefile`.

4. On each update of submodule rocksdb, run  `make -C librocksdb-sys gen_statistics`


zaidoon1:
nice! Would be good to also update maybe the https://github.com/rust-rocksdb/rust-rocksdb/blob/master/MAINTAINERSHIP.md to document `On each update of submodule rocksdb, run make -C librocksdb-sys gen_statistics` so we don't forget

rockeet:
> nice! Would be good to also update maybe the https://github.com/rust-rocksdb/rust-rocksdb/blob/master/MAINTAINERSHIP.md to document `On each update of submodule rocksdb, run make -C librocksdb-sys gen_statistics` so we don't forget

Done!

zaidoon1 (approved):

PR #938, opened 2024-11-02
Fix two tests that want to write to the current working directory

mr-c:
I needed this patch for the Debian package of the `rocksdb` crate, as when we run the tests the current working directory is not writable.

I'm not a Rust developer at all, please let me know if there is a better way to do this.

Thank you!


stanislav-tkach (approved):
I don't think that the prefix is really needed here. Anyway, looking good to me.

zaidoon1 (approved):

PR #943, opened 2024-11-12
Use the pkg_config crate to find a prebuilt rocksdb library

mr-c:
Without needing to set any environment variables

I see this as complimentary to the work over in https://github.com/rust-rocksdb/rust-rocksdb/pull/354


zaidoon1:
hmm, I'm not sure about this change as it's a change in behaviour. it's possible for a system to have an older version of rocksdb installed while the user would want to use this library to pull in the latest version (for one reason or another, maybe a shared system with different applications requiring different versions of rocksdb or something like that). In addition, as we update the library/api to match the latest rocksdb features/api changes, we will need to keep the ` .atleast_version("9.0.0")` updated to make sure we don't break unexpectedly which is brittle. Also, if the system has rocksdb version older than 9.0.0 then we will end up bundling the rocksdb that comes with this library "silently" which again would be surprising to the user.

How about we introduce a new feature that allows the user to choose building rocksdb using this library vs using a prebuilt rocksdb that defaults to building using this library? And we can assume the user is a power user and skip the rocksdb version requirement and hope they use api that are supported by this library?

mr-c:
> hmm, I'm not sure about this change as it's a change in behaviour. it's possible for a system to have an older version of rocksdb installed while the user would want to use this library to pull in the latest version (for one reason or another, maybe a shared system with different applications requiring different versions of rocksdb or something like that). In addition, as we update the library/api to match the latest rocksdb features/api changes, we will need to keep the ` .atleast_version("9.0.0")` updated to make sure we don't break unexpectedly which is brittle. Also, if the system has rocksdb version older than 9.0.0 then we will end up bundling the rocksdb that comes with this library "silently" which again would be surprising to the user.

Yes, all of that is reasonable. I opened this PR to share a patch that we will be carrying in the Debian package of the `rocksdb` crate. I expected that it would need modification before merging.

> How about we introduce a new feature that allows the user to choose building rocksdb using this library vs using a prebuilt rocksdb that defaults to building using this library? And we can assume the user is a power user and skip the rocksdb version requirement and hope they use api that are supported by this library?

That should be fine for the Debian package. Whatever you all want to support.

PR #945, opened 2024-11-14
More temp directories for tests + CI assurance

mr-c:


mr-c:
Proof that the new CI changes catches when tests don't make a temporary directory to write into:

https://github.com/rust-rocksdb/rust-rocksdb/actions/runs/11837843124/job/32985673596?pr=945#step:7:182

```
 thread 'test_comparator_with_ts' panicked at 'called `Result::unwrap()` on an `Err` value: Error { message: "Failed to create RocksDB directory: `Os { code: 13, kind: PermissionDenied, message: \"Permission denied\" }`." }', tests/test_comparator.rs:98:43
```

.github/workflows/rust.yml
```diff
@@ -99,12 +99,35 @@ jobs:
...
+          chmod -R a+w target Cargo.lock
       - name: Run rocksdb tests
         run: |
           cargo test --all
           cargo test --all --features multi-threaded-cf
+      - name: Mark working directory as writable
+        if: runner.os == 'Linux'
+        run: chmod -R a+w .
       - name: Free disk space
         run: cargo clean
```
mr-c, line 119: With the tests using temp directories, perhaps `cargo clean` isn't needed anymore?

I can add a `df -h` before and after, and then try again without the `cargo clean`, if desired.

src/db_iterator.rs
```diff
@@ -33,7 +33,11 @@ pub type DBRawIterator<'a> = DBRawIteratorWithThreadMode<'a, DB>;
 /// ```
 /// use rocksdb::{DB, Options};
 ///
-/// let path = "_path_for_rocksdb_storage4";
+/// let tempdir = tempfile::Builder::new()
+///     .prefix("_path_for_rocksdb_storage4")
+///     .tempdir()
+///     .expect("Failed to create temporary path for the _path_for_rocksdb_storage4.");
+/// let path = tempdir.path();
```
stanislav-tkach, line 40: I'm not sure about that myself, but perhaps it would be better to reduce clutter in the examples by hiding the creation of the temporary directory. This can be done by using the `#` prefix. The disadvantage of such approach is that the temp variable will be implicit for a user of the documentation.

zaidoon1: I would say we should leave the examples/code comments as is (not using a temp path/directory) since that's what users will do in production. The temp directory is good for tests to avoid issues with building/testing the library.

mr-c: Yes. I didn't want to have to make these changes, but once I added the read-only CI test (to detect future regressions) then I noticed that the doc tests had the same problem.

zaidoon1: ok, in that case, I guess what you have here is good enough.

stanislav-tkach (approved):

PR #947, opened 2024-11-21
implement `with_capacity` for `WriteBatch`

0xdeafbeef:


clippy.toml
zaidoon1, line 1: why was this deleted?

zaidoon1: oh i see, we can keep Cargo.toml then

src/write_batch.rs
```diff
@@ -89,6 +89,22 @@ unsafe extern "C" fn writebatch_delete_callback(state: *mut c_void, k: *const c_
...
 impl<const TRANSACTION: bool> WriteBatchWithTransaction<TRANSACTION> {
+    /// Create a new `WriteBatch` without allocating memory.
+    pub fn new() -> Self {
+        Self {
+            inner: unsafe { ffi::rocksdb_writebatch_create() },
+        }
+    }
+
+    /// Creates `WriteBatch` with the specified `capacity` in bytes. Allocates immediately.
+    pub fn with_capacity(capacity: usize) -> Self {
```
zaidoon1, line 100: nit: might be safer to name this `with_capacity_bytes` or something like that. Typically with_capacity in vecs, etc.. is for number of elements and I can see somehow not reading the api docs and just using the number of kvs they want to put into the db as the value which would be bad.. But not sure..

0xdeafbeef: I've renamed it to `with_capacity_bytes`

zaidoon1 (approved):

stanislav-tkach (approved):

PR #949, opened 2024-11-24
add compaction priority options

0xdeafbeef:


zaidoon1:
nit: would be nice to add a test that sets this option without issues. Something like:
```
        opts.create_if_missing(true);
        opts.set_compaction_pri(CompactionPri::RoundRobin);
        let _db = DB::open(&opts, &path).unwrap();
 ```

zaidoon1 (approved):

aleksuss (approved):

Issue #958, opened 2024-12-11
Unable to Open DB with different TTL per-ColumnFamily


jszwec:
Hi,

I'm trying to open a database with different TTLs for each ColumnFamily. However, it seems like the current Rust API signature for [open_cf_descriptors_with_ttl](https://docs.rs/rocksdb/latest/rocksdb/struct.DBCommon.html#method.open_cf_descriptors_with_ttl) only accepts a single TTL.

Looking at the [C++ implementation](https://github.com/facebook/rocksdb/blob/b339d089ed4a7ac7a8ed4ab7d57cccdc0602f0ed/utilities/ttl/db_ttl_impl.cc#L354), it seems that it takes a vector, so per-ColumnFamily TTLs should be possible. Is this an oversight in the Rust API, or am I misunderstanding something?

Thanks in advance!
Jacek

0xdeafbeef:
@aleksuss @stanislav-tkach i think you can close this

aleksuss:
@jszwec if #963 resolves this issue, then close the issue, please.

jszwec:
perfect! thanks!

PR #963, opened 2024-12-23
Allow to set per cf ttl

0xdeafbeef:


src/column_family.rs
```diff
@@ -28,24 +29,64 @@ pub const DEFAULT_COLUMN_FAMILY_NAME: &str = "default";
...
+
+    pub fn ttl(&self) -> CfTtl {
+        self.ttl
+    }
+}
+
+#[derive(Debug, Clone, Copy)]
+/// Specifies the TTL behavior for a column family.
+/// https://github.com/facebook/rocksdb/blob/18cecb9c46b4c2a8b148659dac2fcab5a843d32b/include/rocksdb/utilities/db_ttl.h#L16-L46
+pub enum CfTtl {
```
aleksuss, line 83: I would suggest to rename to the `ColumnFamilyTtl` to be consistent. E.g. `ColumnFamilyDescriptor`

0xdeafbeef: Done

src/column_family.rs
```diff
@@ -28,24 +29,64 @@ pub const DEFAULT_COLUMN_FAMILY_NAME: &str = "default";
...
+    {
+        Self {
+            name: name.into(),
+            options,
+            ttl,
+        }
+    }
+
+    /// Sets ttl for the column family. It's applied only when DB is opened with
+    /// [`DB::open_with_ttl()`]. Changing ttl after DB is opened has no effect.
```
aleksuss, line 65:
```suggestion
    /// [`crate::db::DB::open_with_ttl()`]. Changing ttl after DB is opened has no effect.
```

0xdeafbeef:
fixes #958

src/column_family.rs
```diff
@@ -28,24 +29,64 @@ pub const DEFAULT_COLUMN_FAMILY_NAME: &str = "default";
...
+/// Specifies the TTL behavior for a column family.
+/// <https://github.com/facebook/rocksdb/blob/18cecb9c46b4c2a8b148659dac2fcab5a843d32b/include/rocksdb/utilities/db_ttl.h#L16-L46>
+pub enum ColumnFamilyTtl {
+    /// Will internally set TTL to -1 (disabled)
+    Disabled,
+    /// Will set ttl to the specified duration
+    Set(Duration),
+    /// Will use ttl specified at db open time
+    SameAsWholeDb,
 }
```
aleksuss, line 90: A couple of suggestions here. Looks like it makes sense to add the `Default` trait to the enum.  And I would rename the `Set` variant to `Duration`. It's a rust idiomatic, I guess. 
```rust
#[derive(Debug, Default, Clone, Copy)]
 /// Specifies the TTL behavior for a column family.
 /// <https://github.com/facebook/rocksdb/blob/18cecb9c46b4c2a8b148659dac2fcab5a843d32b/include/rocksdb/utilities/db_ttl.h#L16-L46>
 pub enum ColumnFamilyTtl {
     /// Will internally set TTL to -1 (disabled)
     #[default]
     Disabled,
     /// Will set ttl to the specified duration
     Duration(Duration),
     /// Will use ttl specified at db open time
     SameAsDb,
 }
```

aleksuss: By the way, I'm going to include this PR to #964 and I can do it by myself if you don't mind.

0xdeafbeef: Maybe InheritDbTtl?

aleksuss (approved):
Good work 👍🏻

stanislav-tkach (approved):

src/column_family.rs
aleksuss: > Naming is hard :)
Sometimes it is 😄

But as for me, whether `SameAsWholeDb` or  `SameAsDb` is fine.

PR #992, opened 2025-04-24
Optimize key_may_exist and multi_get methods for better efficiency

GarmashAlex:
Optimize key_may_exist and multi_get methods

This commit makes several small optimizations to improve code efficiency:

1. Simplified boolean comparisons in key_may_exist_* methods
   - Used `!= 0` directly on function calls instead of additional comparison
   - Removed unnecessary temporary variables

2. Optimized multi_get_opt method
   - Eliminated unnecessary collection creation by avoiding unzip()
   - Stored data more efficiently with a single collection

3. Removed unnecessary parentheses for cleaner code

These changes improve performance by reducing allocations and
simplifying expressions without changing functionality.

### DIFF
--- a/src/db.rs
+++ b/src/db.rs
@@ -1124,18 +1124,23 @@ impl<T: ThreadMode, D: DBInner> DBCommon<T, D> {
         K: AsRef<[u8]>,
         I: IntoIterator<Item = K>,
     {
-        let (keys, keys_sizes): (Vec<Box<[u8]>>, Vec<_>) = keys
+        let keys_data: Vec<_> = keys
             .into_iter()
             .map(|k| {
                 let k = k.as_ref();
                 (Box::from(k), k.len())
             })
-            .unzip();
-        let ptr_keys: Vec<_> = keys.iter().map(|k| k.as_ptr() as *const c_char).collect();
+            .collect();
+        
+        let ptr_keys: Vec<_> = keys_data
+            .iter()
+            .map(|(k, _)| k.as_ptr() as *const c_char)
+            .collect();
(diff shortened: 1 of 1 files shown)


src/db.rs
```diff
@@ -1124,18 +1124,23 @@ impl<T: ThreadMode, D: DBInner> DBCommon<T, D> {
...
             })
-            .unzip();
-        let ptr_keys: Vec<_> = keys.iter().map(|k| k.as_ptr() as *const c_char).collect();
+            .collect();
+        
+        let ptr_keys: Vec<_> = keys_data
+            .iter()
+            .map(|(k, _)| k.as_ptr() as *const c_char)
+            .collect();
+        let keys_sizes: Vec<_> = keys_data.iter().map(|(_, size)| *size).collect();
```
stanislav-tkach, line 1139: > Eliminated unnecessary collection creation by avoiding unzip()

I don't get this part: instead of using unzip you are just using another collect. What is the difference?

src/db.rs
```diff
@@ -1355,21 +1360,20 @@ impl<T: ThreadMode, D: DBInner> DBCommon<T, D> {
...
+                cf.inner(),
+                key.as_ptr() as *const c_char,
+                key.len() as size_t,
+                &mut val,         /*value*/
+                &mut val_len,     /*val_len*/
+                ptr::null(),      /*timestamp*/
+                0,                /*timestamp_len*/
+                &mut value_found, /*value_found*/
+            ) != 0
+        };
```
stanislav-tkach, line 1376: I don't understand how it improves anything.

Issue #693, opened 2022-10-10
Misuse of API `opt_bytes_to_ptr` easily leads to UAF


BurtonQin:
https://github.com/rust-rocksdb/rust-rocksdb/blob/6cd1c020b0f0f9517ebe962fa485008938a46aab/src/ffi_util.rs#L46-L51

Vec and array implement AsRef<[u8]> so the following code can compile:

```Rust
use core::ffi::c_char;
use std::ptr;

pub fn opt_bytes_to_ptr<T: AsRef<[u8]>>(opt: Option<T>) -> *const c_char {
    match opt {
        Some(v) => v.as_ref().as_ptr() as *const c_char,
        None => ptr::null(),
    }
}

fn main() {
    let v = vec![1, 2, 3];  // Vec
    // let v = [1, 2, 3];   // or array
    let c = opt_bytes_to_ptr(Some(v));  // Vec or array dropped inside
    unsafe {
        println!("{:?}", *c);  // UAF
    }
}
```

`opt_bytes_to_ptr` takes the ownership of Vec or array and dropped them inside.
The returned ptr comes dangling and dereference the ptr will cause UAF.

Currently, I did not find any prolem because only a reference is passed to `opt_bytes_to_ptr` in every poossible usage.
But I think this API needs to be improved to avoid future misuse.

One possible fix is to change the function signature to take `&[u8]` instead of `T: AsRef<[u8]>`.

PR #1003, opened 2025-05-18
ffi_util.rs: improve opt_bytes_to_str to fix (#693)

Chain-Fox:
# Unsound API design in opt_bytes_to_ptr

The original `opt_bytes_to_ptr` function accepts a parameter of type `Option<T> where T: AsRef<[u8]>`. While this is flexible, it also allows owned types like `String` or `Vec<u8>` to be passed in. Since the function takes ownership of T, any owned data (like a `String`) will be dropped at the end of the function scope.

This results in undefined behavior, as the returned `*const c_char` pointer may point to memory that is already freed, leading to dangling pointers.

See the example below:
https://play.rust-lang.org/?version=stable&mode=debug&edition=2024&gist=a59bc55798c258040f5aa04c9e60c1b5

## Fix: Change to borrowed data
To resolve this, the function should accept a borrowed reference instead, such as:
The function no longer takes ownership of the data.
The returned pointer is guaranteed to be valid for at least as long as the input reference.

## Safety Note
It is the caller's responsibility to ensure the borrowed data outlives any use of the returned raw pointer. This is a standard requirement when interfacing with raw C-style APIs.

## Additional Changes
Added documentation to clarify the ownership and lifetime requirements of the input data and returned pointer.

### DIFF
--- a/src/ffi_util.rs
+++ b/src/ffi_util.rs
@@ -43,7 +43,12 @@ pub fn error_message(ptr: *const c_char) -> String {
     }
 }
 
-pub fn opt_bytes_to_ptr<T: AsRef<[u8]>>(opt: Option<T>) -> *const c_char {
+/// Returns a raw pointer to borrowed bytes, or null if None.
+///
+/// # Safety
+/// - The input must outlive the returned pointer.
+/// - Common types: `&str`, `&[u8]`, `&String`, `&Vec<u8>`
+pub fn opt_bytes_to_ptr<T: AsRef<[u8]> + ?Sized>(opt: Option<&T>) -> *const c_char {
     match opt {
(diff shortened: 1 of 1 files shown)


zaidoon1 (approved):

aleksuss (approved):

stanislav-tkach:
Perhaps this function should be marked as unsafe if it has preconditions to not to be unsound.

BurtonQin:
In Rust, it is traditionally considered safe to convert a reference into a raw pointer. This conversion itself does not require unsafe and is harmless. However, when **using** the raw pointer later (e.g., dereferencing it), it becomes the programmer's responsibility to ensure that the original reference remains valid — that is, it outlives the raw pointer.

PR #1009, opened 2025-06-09
Update broken bindgen issue link to new repository location

VolodymyrBg:
The original comment referenced an outdated GitHub issue link for rust-bindgen regarding the max_align_t blocklist workaround.
This commit updates the URL to point to the current location of the issue in the rust-lang/rust-bindgen repository:
https://github.com/rust-lang/rust-bindgen/issues/550
This helps future maintainers understand the reason for blocking max_align_t and provides a working reference for more context.


stanislav-tkach:
The referenced issue is closed. Is this still relevant?.. I'm too lazy to check that right now... 🙃

aleksuss:
The original link goes to the relevant issue. So, the change is not necessary.

PR #1011, opened 2025-06-30
Untie PinnableSlices' lifetime from their transaction's lifetime

### LINKED ISSUE #1005: Transaction's get_pinned's DbPinnableSlice borrows Transaction instead of TransactionDB
Hey!

I'm trying to use rocksdb from an async context. In order to avoid memory copies, I'd like to use `get_pinned`.

However, there are `.await` points between my various `get_pinned` calls in a single transaction. This means that the `Transaction` ends up being borrowed from multiple threads, and thus requiring `Sync`, which it isn't. And so I cannot actually use `get_pinned`.

Reading the `PinnableSlice` rocksdb documentation, I'm not sure I totally understand the rational for it to borrow something, but to the best of my understanding it should at most need to borrow `TransactionDB`, and not the `Transaction` itself.

Do you think it would be possible to have the `get_pinned_*` actually return a `DbPinnableSlice` that borrows the `TransactionDB`? This'd require adding a borrow to the `get_pinned_*` methods so that the lifetime could be passed through, but I think it should be worth it — it'd unblock calling multiple `get_pinned` from async code.

(AFAICT I'd still need a mutex around the `Transaction`, because I wouldn't be able to explain to rustc that it's indeed only used in a single-threaded manner, but that's probably a lesser evil than copying around the data)

Ekleog:
Hey! Here is a small PR that should solve a problem of mine.

Fixes #1005

Disclaimer: I'm not 100% sure this is correct. And actually, considering the implementation of DBPinnableSlice, I'd suggest even removing the lifetime parameter altogether — lifetime seems to be reference-counted, so why are we even keeping the `&'db DB` as a phantomdata in the transaction itself?

Anyway, thank you for all your work on the rust rocksdb bindings!

### DIFF
--- a/src/transactions/transaction.rs
+++ b/src/transactions/transaction.rs
@@ -227,7 +227,10 @@ impl<'db, DB> Transaction<'db, DB> {
         self.get_opt(key, &ReadOptions::default())
     }
 
-    pub fn get_pinned<K: AsRef<[u8]>>(&self, key: K) -> Result<Option<DBPinnableSlice<'db>>, Error> {
+    pub fn get_pinned<K: AsRef<[u8]>>(
+        &self,
+        key: K,
+    ) -> Result<Option<DBPinnableSlice<'db>>, Error> {
         self.get_pinned_opt(key, &ReadOptions::default())
     }
 


Ekleog:
cc @aleksuss @stanislav-tkach @zaidoon1 based on the commenters to another recent PR; did one of you see this and/or the related issue? 😄

PR #1020, opened 2025-07-24
Remove set_skip_prepare method from TransactionOptions

MozirDmitriy:
The set_skip_prepare method has been removed from TransactionOptions because there is no corresponding function in the RocksDB C API or in the project's FFI layer to support this option. The previous implementation incorrectly called the set_snapshot FFI function, which is unrelated to skip_prepare. Keeping this method would mislead users into thinking this option is supported, while in reality it has no effect. This change ensures the Rust API accurately reflects the available functionality in the underlying RocksDB library.

### DIFF
--- a/src/transactions/options.rs
+++ b/src/transactions/options.rs
@@ -38,12 +38,6 @@ impl TransactionOptions {
         TransactionOptions::default()
     }
 
-    pub fn set_skip_prepare(&mut self, skip_prepare: bool) {
-        unsafe {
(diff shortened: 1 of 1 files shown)


aleksuss:
But [this](https://github.com/facebook/rocksdb/blob/812b12bc7827eb0589927befc1514cff86eb46c6/db/c.cc#L6667) is over there, right? It should be fixed.

PR #1021, opened 2025-07-25
Check sched_getcpu capability

bvscd:


librocksdb-sys/build.rs
```diff
@@ -247,6 +247,29 @@ fn build_rocksdb() {
...
+        .file(&out_file)
+        .try_compile("test_sched_getcpu")
+        .is_ok() 
+    {
+        config.define("ROCKSDB_SCHED_GETCPU_PRESENT", None);
+    }
+    fs::remove_file(&out_file).ok();
+
+    if cfg!(feature = "jemalloc") {
+        config.define("WITH_JEMALLOC", "ON");
```
zaidoon1, line 270: jemalloc config is already present below

PR #1032, opened 2025-08-25
fix: use correct FFI setter for skip_prepare in RocksDB transaction options

GarmashAlex:
- Replaced incorrect logic with the proper FFI call `rocksdb_transaction_options_set_skip_prepare`.
- Ensures `skip_prepare` option is correctly applied when configuring RocksDB transactions.


stanislav-tkach (approved):

aleksuss:
Nice catch 👍

aleksuss (approved):

Issue #1033, opened 2025-09-04
Expose raw pointers to objects


jszwec:
Hello,

I would like to make a proposal to expose raw C pointers via trait and gate it behind a feature flag.

This would make it possible to build custom shims or integrations when certain functionality is not yet available in the public API, without requiring internal patches or forks. The feature flag would ensure it remains opt-in for advanced users who understand the safety implications.

PR #1044, opened 2025-10-13
Proposal: Add feature-gated trait to expose raw C pointers

jszwec:
Hello, I would like to propose a way to expose raw C pointers.

This would make it possible to build custom shims or integrations when certain functionality is not yet available in the public API, without requiring internal patches or forks. The feature flag would ensure it remains opt-in for advanced users who understand the safety implications.

For now I propose only exposing these three impls, but please let me know if anything else would be useful

I’m not sure if this is the right approach to introduce such functionality, so I’d appreciate any feedback or guidance on whether this direction makes sense.

Closes: #1033

### DIFF
--- a/src/raw_ptr.rs
+++ b/src/raw_ptr.rs
@@ -15,7 +15,6 @@ use librocksdb_sys as ffi;
 /// ```rust,ignore
 /// use rocksdb::{DB, Options, AsRawPtr};
 ///
-/// # fn main() -> Result<(), Box<dyn std::error::Error>> {
 /// let db = DB::open_default("path")?;
 /// let raw_db_ptr = unsafe { db.as_raw_ptr() }; // *mut rocksdb_t
 ///
@@ -24,8 +23,6 @@ use librocksdb_sys as ffi;
 ///
 /// // You can now use these pointers with the C API directly
 /// // unsafe { rocksdb_some_c_function(raw_db_ptr, raw_opts_ptr); }
-/// # Ok(())
-/// # }
 /// ```
 ///
 /// # Safety


jszwec:
Hello, @aleksuss @zaidoon1 could you guys please take a look? 🙏

jszwec:
Hello @zaidoon1 

Thank you for getting back to me.

I have been particularly interested in using [GetLiveFilesStorageInfo](https://github.com/facebook/rocksdb/blob/9065ace05ada92e0a435fbb973ef957fc4401f43/db/db_impl/db_impl.h#L567) (different from LiveFiles).

I fully agree that reusable functionality should live in the public API of this crate. The blocker in this case is that on the RocksDB side `GetLiveFilesStorageInfo` is available in the C++ API, but not in the C API.  I’d like to eventually send a PR to RocksDB to expose it there, and then expose it in this crate. Until that happens, I’m forced to build a C++ shim that needs access to the underlying RocksDB objects. Right now that requires:
1. forking `rust-rocksdb`
2. updating the fork to give access to raw pointers
3. implementing shim

My motivation with this feature flag is to eliminate step 1 and 2 making it easier for advanced users to prototype missing functionality without maintaining forks.

src/raw_ptr.rs
```diff
@@ -0,0 +1,72 @@
...
+use librocksdb_sys as ffi;
+
+/// Trait for accessing raw pointers to underlying RocksDB objects.
+///
+/// This trait is only available when the `raw-ptr` feature is enabled.
+/// It provides access to the underlying C API pointers for advanced use cases.
+///
+/// # Examples
+///
+/// ```rust
```
zaidoon1: nit: this example won't compile, might want to use: `ignore`

src/raw_ptr.rs
```diff
@@ -0,0 +1,72 @@
...
+/// after the objects have been dropped.
+pub trait AsRawPtr<T> {
+    /// Returns a raw pointer to the underlying RocksDB object.
+    ///
+    /// # Safety
+    ///
+    /// The returned pointer is only valid as long as the object implementing
+    /// this trait is alive. The caller must ensure proper lifetime management
+    /// and avoid using the pointer after the object has been dropped.
+    fn as_raw_ptr(&self) -> *mut T;
```
zaidoon1: nit: maybe best to mark as unsafe to really really make sure the user knows they are in the danger zone

zaidoon1 (approved):

Issue #814, opened 2023-08-16
expose custom logger configuration


zaidoon1:
Per https://github.com/facebook/rocksdb/wiki/Logger#existing-logger-implementations , we can use a custom logger to get the logs to show up in the format we need and more importantly where we need. In my case, I want to use "StderrLogger | Pipes the messages to std::err equivalent" but this library doesn't appear to allow me to set this option.

zaidoon1:
facebook pr to expose the c api has been merged: https://github.com/facebook/rocksdb/pull/12262

will create the rust bindings once the c binding changes land in the rocksdb release

PR #1051, opened 2025-11-28
Added custom logging callback functionality

ischeinkman:
Implements #814 . 

I wasn't sure what was preferred for converting between the `u32` level the rocks C API exposes and the `LogLevel` enum they use internally, so I decided on `try_from_raw`. Please tell me if there's a more preferred way!


ischeinkman:
@aleksuss Is there anything else you need before this can be merged?

aleksuss (approved):

PR #1061, opened 2026-01-02
MemoryUsageBuilder: Prevent outliving the DBs/caches

evanj:
This type must not outlive the DBs and caches that are added to it, otherwise the caller will access invalid memory and likely crash. Use PhantomData to ensure it does not outlive its arguments. This is technically a breaking API change, but anything that used this code in this way was likely broken anyway.

### DIFF
--- a/src/perf.rs
+++ b/src/perf.rs
@@ -12,6 +12,8 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
+use std::marker::PhantomData;
+
 use libc::{c_int, c_uchar, c_void};
 
 use crate::{db::DBInner, ffi, ffi_util::from_cstr, Cache, Error};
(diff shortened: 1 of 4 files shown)


src/perf.rs
```diff
@@ -214,20 +216,24 @@ impl MemoryUsage {
...
 
-/// Builder for MemoryUsage
-pub struct MemoryUsageBuilder {
+/// Creates [`MemoryUsage`] from DBs and caches.
+///
+/// Most users should call [`get_memory_usage_stats`] instead.
+pub struct MemoryUsageBuilder<'a> {
     inner: *mut ffi::rocksdb_memory_consumers_t,
+    // must not outlive the DBs/caches that are added
+    _marker: PhantomData<&'a ()>,
```
evanj: This is going to conflict with my leak fix change #1060. I wanted to do this separately since it technically breaks the API, so I thought this might be more controversial. I'll fix the merge conflicts after one of the changes is merged.

evanj: Okay, I rebased it so it should no longer have conflicts. Thanks!

zaidoon1 (approved):

aleksuss (approved):

stanislav-tkach (approved):

PR #1065, opened 2026-01-18
Add timestamp() to DBRawIterator to retrieve user defined timestamps

dillonhicks:
This change exposes the user defined timestamp accessor from the DBRawIteratorWithThreadMode.

### DIFF
--- a/src/db_iterator.rs
+++ b/src/db_iterator.rs
@@ -357,6 +357,15 @@ impl<'a, D: DBAccess> DBRawIteratorWithThreadMode<'a, D> {
         }
     }
 
+    /// Returns a slice of the timestamp of the current entry.
+    pub fn timestamp(&self) -> Option<&[u8]> {
+        if self.valid() {
+            Some(self.timestamp_impl())
+        } else {
+            None
+        }
+    }
+
     /// Returns pair with slice of the current key and current value.
     pub fn item(&self) -> Option<(&[u8], &[u8])> {
         if self.valid() {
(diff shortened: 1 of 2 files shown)


src/db_iterator.rs
```diff
@@ -342,6 +342,15 @@ impl<'a, D: DBAccess> DBRawIteratorWithThreadMode<'a, D> {
         }
     }
 
+    /// Returns a slice of the current timestamp.
```
aleksuss: As far as I can see, it's not the current timestamp. It's the time at which the element was written that the iterator points to.

dillonhicks: You're right! It can also be a user supplied timestamp. I will update the language accordingly.

src/db_iterator.rs
```diff
@@ -357,6 +357,15 @@ impl<'a, D: DBAccess> DBRawIteratorWithThreadMode<'a, D> {
         }
     }
 
+    /// Returns a slice of the timestamp of the current entry.
+    pub fn timestamp(&self) -> Option<&[u8]> {
```
aleksuss, line 361: Mayby it makes sense to return `u64` or `U64Timestamp` instead of a slice of bytes? WDYT?

dillonhicks: Something more strongly typed would be great. However, I think [User Defined Timestamps (rocks wiki)](https://github.com/facebook/rocksdb/wiki/User-defined-Timestamp) can be arbitrarily sized. All of the examples use 8 byte timestamps, but I know of cases were the full 16 byte `struct timespec` is used.

dillonhicks: Any thoughts?

aleksuss: Ok. Let's keep `&[u8]`

aleksuss (approved):

PR #1075, opened 2026-02-17
Fifo compaction options

Suficio:
Introduces commonly used methods for `FifoCompactOptions`.

- Modifies `fifo_compaction_test` due to FIFO options being incorrectly set for the column family.
- Adds new unit test `fifo_compaction_allow_compaction_test` to validate automatic compaction.

### DIFF
--- a/src/db_options.rs
+++ b/src/db_options.rs
@@ -4596,6 +4596,24 @@ impl Drop for FifoCompactOptions {
 }
 
 impl FifoCompactOptions {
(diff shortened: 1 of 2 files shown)


Suficio:
Rebased this PR with a fix to the unit tests.

Did not realize that the `DB` type changed based on compile-time flags which caused a compile failure in CI while passing locally.

aleksuss (approved):

PR #1080, opened 2026-05-05
Add mempurge threshold option

0xdeafbeef:
I can drop flake.* or move to the other pr

### DIFF
--- a/src/db_options.rs
+++ b/src/db_options.rs
@@ -3577,6 +3577,29 @@ impl Options {
         }
     }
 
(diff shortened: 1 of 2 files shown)


flake.lock
aleksuss, line 1: Probably should be removed?

zaidoon1:
LGTM but please drop the flake stuff

flake.lock
0xdeafbeef: done

0xdeafbeef:
> LGTM but please drop the flake stuff

done

aleksuss (approved):

PR #1086, opened 2026-06-01
Add TransactionDB checkpoint support

calavera:
I've added a similar implementation to Checkpoint for `TransactionDB` to expose the internal `rocksdb_transactiondb_checkpoint_object_create` that's already implemented in `rocksdb-sys`.

### DIFF
--- a/src/checkpoint.rs
+++ b/src/checkpoint.rs
@@ -17,7 +17,7 @@
 //!
 //! [1]: https://github.com/facebook/rocksdb/wiki/Checkpoints
 
-use crate::{db::DBInner, ffi, ffi_util::to_cpath, DBCommon, Error, ThreadMode};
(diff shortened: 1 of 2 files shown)


aleksuss (approved):

calavera:
@zaidoon1 @aleksuss is there anything missing to merge this? I want to make sure I can close the loop here. Thanks!

gdorsi:
I've done a similar contribution on https://github.com/zaidoon1/rust-rocksdb/pull/217 and happy to see that we both come to the idea that a TransactionDBCheckpoint is the best direction for the API 😄 

In my work I haven't added `create_checkpoint_with_log_size` because with TransactionDB the WAL is always flushed when doing a checkpoint, making the method not really doing what it says.

Also, just realized that the test I added in the PR wasn't checking this correctly, so added a new test to better verify this claim: https://github.com/zaidoon1/rust-rocksdb/pull/235

PR #1092, opened 2026-07-03
Add set_cf_paths to allow per column family sst and blob paths

crypt0miester:
Hi @aleksuss, small parity addition, would appreciate a review when you have a moment.

This PR adds a safe wrapper for `rocksdb_options_set_cf_paths`, the per-column-family counterpart of the existing `set_db_paths`. The C symbol is already exposed in the bundled RocksDB, so this is a pure Rust addition mirroring `set_db_paths`, plus a placement test.

We need this for https://github.com/spool-labs/tape. Storage nodes there typically have a small fast NVMe drive and large HDDs, and the store keeps small hot metadata/index column families next to bulk BlobDB column families holding large payloads. `cf_paths` lets us pin the bulk column families to the HDD while metadata and the WAL stay on the SSD. It is also the only way to steer integrated BlobDB data: blob files are always written to `cf_paths.front()`, and the old `blob_dir` option is stacked-BlobDB-only and not in the C API.

Without this binding the option is unreachable from safe code, since `Options.inner` is `pub(crate)` and the crate does not re-export `librocksdb_sys`.

Regards, 
miester

### DIFF
--- a/src/db_options.rs
+++ b/src/db_options.rs
@@ -1373,6 +1373,27 @@ impl Options {
         }
     }
 
+    /// A list of paths where SST files for this column family can be put
+    /// into, with its target size. Similar to `set_db_paths`, newer data is
+    /// placed into paths specified earlier in the vector while older data
+    /// gradually moves to paths specified later in the vector.
+    ///
+    /// Note that, if a path is supplied to multiple column families, it would
+    /// have files and total size from all the column families combined. User
+    /// should provision for the total size (from all the column families) in
+    /// such cases.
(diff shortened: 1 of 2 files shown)


aleksuss (approved):

crypt0miester:
bump please. @stanislav-tkach @zaidoon1

crypt0miester:
much appreciated. thank you @aleksuss

dr333ws:
Opened #1096 asking about a release that includes this.

PR #1095, opened 2026-07-20
make ReadOptions::default() not allocate

alessandrod:
We recently made some code that consumes a few rocks column families faster, and noticed increased jemalloc decay churn in profiles. I started profiling the relevant code and discovered quite a few unexpected allocations. The most surprising was this one:

 
Screenshot 2026-07-20 at 4 55 01 pm

this is a 15s profile, where 15% of the allocations come from us calling `get_pinned_cf(...)` which internally does `get_pinned_cf_opt(..., &ReadOptions::default())` and turns out each `ReadOptions::default()` is a (very short lived) heap allocation.

This PR removes the allocation by making `ReadOptions::default()` return a static value, and by making `ReadOptions` copy-on-write.

The same result could be achieved by leaving `ReadOptions::default()` as is, and introducing an internal static `DEFAULT_READ_OPTIONS` or something that gets passed instead of `&ReadOptions::default()`. I have a slight preference for the CoW approach since it's self contained and works for user created `ReadOptions` instances too, but if you prefer I'm happy to switch to the other approach.

### DIFF
--- a/src/db.rs
+++ b/src/db.rs
@@ -220,7 +220,7 @@ impl<T: ThreadMode, D: DBInner> DBAccess for DBCommon<T, D> {
     }
 
     unsafe fn create_iterator(&self, readopts: &ReadOptions) -> *mut ffi::rocksdb_iterator_t {
-        unsafe { ffi::rocksdb_create_iterator(self.inner.inner(), readopts.inner) }
+        unsafe { ffi::rocksdb_create_iterator(self.inner.inner(), readopts.inner()) }
     }
 
     unsafe fn create_iterator_cf(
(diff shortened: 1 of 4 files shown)


alessandrod:
friendly bump!

aleksuss:
@alessandrod thank you for your contribution, but we need more time to assess the changes. At least, what I can see now is that it would probably be good to apply the same approach to other option structs as well.

alessandrod:
> @alessandrod thank you for your contribution, but we need more time to assess the changes. At least, what I can see now is that it would probably be good to apply the same approach to other option structs as well.

Yep I didn't want to send a huge PR, but I'm happy to generalize and cover the other ones too if you want (I was going to follow up with WriteOptions but can do all in this PR if you prefer)

PR #1097, opened 2026-08-03
fix(unix): preserve non-UTF-8 path bytes across FFI

grandpig:
## Summary

Preserve the original bytes of Unix paths when converting them to `CString` for RocksDB's C API.

Also make `BackupEngineOptions::new()` reuse `to_cpath()` instead of maintaining a separate path-conversion implementation.

## Problem

`to_cpath()` currently converts paths using:

```rust
CString::new(path.as_ref().to_string_lossy().as_bytes())
```

On Unix, paths are byte sequences and are not required to be valid UTF-8.

`to_string_lossy()` replaces invalid UTF-8 bytes with the UTF-8 encoding of U+FFFD.

As a result, the path passed to RocksDB may differ from the path supplied by the caller. During database opening, the Rust wrapper may create the original directory while RocksDB opens or creates the database at the lossy, rewritten path.

`BackupEngineOptions::new()`  independently performed the same lossy conversion.

### DIFF
--- a/src/ffi_util.rs
+++ b/src/ffi_util.rs
@@ -66,12 +66,35 @@ pub fn opt_bytes_to_ptr<T: AsRef<[u8]> + ?Sized>(opt: Option<&T>) -> *const c_ch
 }
 
 pub(crate) fn to_cpath<P: AsRef<Path>>(path: P) -> Result<CString, Error> {
-    match CString::new(path.as_ref().to_string_lossy().as_bytes()) {
+    let path = path.as_ref();
+
+    #[cfg(unix)]
+    let cpath = {
+        use std::os::unix::ffi::OsStrExt;
+        CString::new(path.as_os_str().as_bytes())
+    };
+
+    #[cfg(not(unix))]
+    let cpath = CString::new(path.to_string_lossy().as_bytes());
+
+    match cpath {
         Ok(c) => Ok(c),
         Err(e) => Err(Error::new(format!(
             "Failed to convert path to CString: {e}"
         ))),
     }
+}
+
+#[cfg(all(test, unix))]
+#[test]
+fn to_cpath_preserves_non_utf8_bytes() {
+    use std::ffi::OsString;
+    use std::os::unix::ffi::{OsStrExt, OsStringExt};
+
+    let path = Path::new(&OsString::from_vec(b"rocksdb-\xff".to_vec())).to_path_buf();
+    let cpath = to_cpath(&path).unwrap();
+
+    assert_eq!(path.as_os_str().as_bytes(), cpath.as_bytes());
 }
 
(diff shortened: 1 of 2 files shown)


grandpig:
Hi  @evanj  @spacejam Could you please take a look when you have a chance? Thanks!

src/ffi_util.rs
```diff
@@ -66,14 +66,37 @@ pub fn opt_bytes_to_ptr<T: AsRef<[u8]> + ?Sized>(opt: Option<&T>) -> *const c_ch
...
+
+    match cpath {
         Ok(c) => Ok(c),
         Err(e) => Err(Error::new(format!(
             "Failed to convert path to CString: {e}"
         ))),
     }
 }
 
+#[cfg(all(test, unix))]
```
aleksuss, line 88: Why is it bound by `unix` only?

grandpig: Thanks for your review.  

Because this regression test relies on Unix path semantics.

On Unix, paths are arbitrary byte sequences, so the test uses `OsStringExt::from_vec()` to construct a path containing an invalid UTF-8 byte and `OsStrExt::as_bytes()` to verify that the byte is preserved.
Those APIs are Unix-specific.

Windows paths use wide code units instead, and RocksDB is built with `ROCKSDB_WINDOWS_UTF8_FILENAMES`, so this particular raw-byte test does not apply there. The existing non-Unix conversion behavior is intentionally left unchanged.

grandpig:
The CI failures appear unrelated to this PR. The repository uses Rust 1.85 and declares `trybuild = "1"` without tracking Cargo.lock. `trybuild` 1.0.120 was recently released with MSRV 1.88, so fresh dependency resolution now selects a version incompatible with the repository's Rust version.

Pinning the dev-dependency to `trybuild = "=1.0.119"` should restore Rust 1.85 compatibility.

aleksuss:
> Pinning the dev-dependency to trybuild = "=1.0.119" should restore Rust 1.85 compatibility.

I believe it's better to bump the Rust version.

grandpig:
> > Pinning the dev-dependency to trybuild = "=1.0.119" should restore Rust 1.85 compatibility.
> 
> I believe it's better to bump the Rust version.

That makes sense. Would you like me to bump the Rust version in this PR, or would you prefer to handle it in a separate PR?

aleksuss:
> > > Pinning the dev-dependency to trybuild = "=1.0.119" should restore Rust 1.85 compatibility.
> > 
> > 
> > I believe it's better to bump the Rust version.
> 
> That makes sense. Would you like me to bump the Rust version in this PR, or would you prefer to handle it in a separate PR?

Yes. Let's try to do it within the scope of this PR. I hope it requires not too many changes.

